PR #6, opened 2022-05-09
Changes for Features V2

edgonmsft:
This includes the changes outlined in the following proposal:

https://github.com/devcontainers/spec/pull/27

Still has a couple of things in progress but can be commented on.

Fixes other issues: 
https://github.com/github/codespaces/issues/7101


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -192,17 +212,48 @@ USER $_DEV_CONTAINERS_IMAGE_USER
 
 export function getFeatureLayers(featuresConfig: FeaturesConfig) {
 	let result = '';
-	const folders = (featuresConfig.featureSets || []).map(x => getSourceInfoString(x.sourceInformation));
+
+	// Features version 1  FIX
```
edgonmsft, line 216: This can cause issues since it breaks the ordering of how features are executed, adding features v1 first and then features v2.
There is no easy way to fix it since some feature sets might need to be called together at the same time and that already breaks the convention.

One option would be to execute v2 first and ensure that we include a cache of v2 at the same time but its still an issue for user written features.

chrmarti: Features shouldn't depend on the order they are executed. Those that do we should take a look at.

edgonmsft: The only one so far is the Jupyter feature that requires some sort of Python installed.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch cli - up",
+            "program": "${workspaceFolder}/src/spec-node/devContainersSpecCLI.ts",
+            "cwd": "${workspaceFolder}",
+            "args": ["up", "--workspace-folder", "../features-playground", "--log-level", "debug", ],
```
joshspicer, line 13: Perhaps we can add in a very simple feature set/collection as one of the tests, and use that here?  That way we can always F5 into something useful.

package.json
```diff
@@ -65,6 +65,8 @@
 		"vinyl-fs": "^3.0.3"
 	},
 	"dependencies": {
+		"@types/ncp": "^2.0.5",
```
joshspicer, line 68: nit: `@types/ncp` can be added as a devDependency

src/spec-common/injectHeadless.ts
```diff
@@ -105,13 +105,18 @@ export type DevContainerConfigCommand = 'initializeCommand' | 'onCreateCommand'
 
 const defaultWaitFor: DevContainerConfigCommand = 'updateContentCommand';
 
+export interface DevContainerFeature{
```
joshspicer, line 108: nit: 
```suggestion
export interface DevContainerFeature {
```

src/spec-configuration/configuration.ts
```diff
@@ -26,6 +26,11 @@ export interface HostRequirements {
 	storage?: string;
 }
 
+export interface DevContainerFeature{
```
joshspicer, line 29: nit
```suggestion
export interface DevContainerFeature {
```

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -5,20 +5,32 @@
 
 import * as jsonc from 'jsonc-parser';
 import * as path from 'path';
-import * as semver from 'semver';
+//import * as semver from 'semver';
```
joshspicer, line 8: nit: remove commented import

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -5,20 +5,32 @@
...
 export interface Feature {
 	id: string;
 	name: string;
+	description?: string;
+	filename?: string;
+	runApp?: string;
+	runParams?: string;
+	infoString?: string;
+	internalVersion?: string;
+	tempLocalPath?: string;
```
joshspicer, line 29: It would be nice to know here which properties can be user-supplied in `devcontainer-feature.json`, and which are `// set programatically`

src/spec-node/containerFeatures.ts
```diff
@@ -212,28 +230,43 @@ ARG _DEV_CONTAINERS_BASE_IMAGE=mcr.microsoft.com/vscode/devcontainers/base:buste
...
-			...featureSet.features
+	for await (const fSet of featuresConfig.featureSets) {
+		let i = 0;
+		if(fSet.internalVersion === '2')
+		{
+			for await (const fe of fSet.features) {
+				if (fe.infoString)
+				{
+					fe.internalVersion = '2';
+					const envPath = cliHost.path.join(fe.infoString, 'devcontainer-features.env');
```
joshspicer, line 241: Should we move `'devcontainer-features.env'` out to be a global variable at the top of this file?

joshspicer (dismissed):

src/spec-common/injectHeadless.ts
```diff
@@ -105,13 +105,18 @@ export type DevContainerConfigCommand = 'initializeCommand' | 'onCreateCommand'
...
+}
+
 export interface CommonDevContainerConfig {
 	configFilePath?: URI;
 	remoteEnv?: Record<string, string | null>;
 	forwardPorts?: (number | string)[];
 	portsAttributes?: Record<string, PortAttributes>;
 	otherPortsAttributes?: PortAttributes;
-	features?: Record<string, string | boolean | Record<string, string | boolean>>;
+	features?: DevContainerFeature[] | Record<string, string | boolean | Record<string, string | boolean>>;
```
chrmarti, line 119: We don't want to support the array variant. Please see https://github.com/devcontainers/spec/pull/27/files#r869018039.

edgonmsft: Made a comment on that thread.

joshspicer: I have shared some thoughts as well

joshspicer (dismissed):

src/spec-common/injectHeadless.ts
jkeech: Dropping my thoughts here since the spec PR has already been merged for a couple weeks. I strongly believe we need to support the array syntax, and I would go a step further and say that we should only document the array syntax going forward, leaving the object syntax in place only for backwards compatibility (object can be easily mapped to the array form at runtime in a deterministic way).

There are several scenarios we've encountered from users who need to better control feature execution order as a consumer of the feature in a way that the feature author wouldn't know about in advance. Most of these involve custom logic that the consumer wants to run after a feature is installed, but are things the user wants baked into the image itself and not later on in a lifecycle hook.  A few of the examples:
* Running a tool from an earlier feature to install other tools in a later feature
    * A user wants to install a global npm package (`npm install -g`) after Node/npm are installed
    * A user wants to install some Python packages through pip after installing python/pip
    * A user wants to install an Azure CLI extension after installing the Azure CLI
    * A user wants to install a GH CLI extension after installing the GH CLI

I understand the argument that the tooling should be able to figure out the right order if dependencies are defined, but I don't think we have a solution for dependency declaration within each feature that is robust enough to handle all of the cases. Ultimately, giving users the ability to force a specific ordering puts them in control and allows them to make adjustments as needed instead of being at the mercy of the tooling to figure out the perfect install order.

For most users, they won't need to worry about the installation order of different features, and in that case, both syntaxes work just fine. But in the case where users do need the control for the ordering, the array syntax makes that possible, whereas the object syntax does not.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -0,0 +1,108 @@
...
+
+interface FeatureNode {
+    feature: FeatureSet;
+    before: Set<FeatureNode>;
+    after: Set<FeatureNode>;
+}
+
+export function computeFeatureInstallationOrder(config: DevContainerConfig, features: FeatureSet[]) {
+
+    if (!config.overrideFeatureInstallOrder) {
```
joshspicer, line 19: nit: Swap the order of this to make it more readable

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -0,0 +1,108 @@
...
+            throw new ContainerError({ description: `Feature ${featureId} not found` });
+        }
+        orderedFeatures.push(feature);
+        features.splice(features.indexOf(feature), 1);
+    }
+
+    return orderedFeatures.concat(features);
+}
+
+function computeInstallationOrder(features: FeatureSet[]) {
```
joshspicer, line 44: I think a comment on this function illustrating the set of rules its following would be helpful

eg:

- Looks at each feature's `installAfter` to resolve soft dependencies
- ???

joshspicer (dismissed):

joshspicer (approved):
As the automatic ordering (and accompanying overrides) get more complicated, I think it would be great to add in tests that run against the final `Dockerfile` and assert the order of statements, etc..

Can leave that for another PR 🚀

samruddhikhandale:
Yes! https://github.com/devcontainers/images/pull/31
Tests passed and even I locally tested by building few images. Looks good! 

(Didn't make changes to codespaces due to need of multiple excecution of features)

Ahmadkl054:
> This includes the changes outlined in the following proposal:
> 
> https://github.com/devcontainers/spec/pull/27
> 
> Still has a couple of things in progress but can be commented on.
> 
> Fixes other issues: 
> https://github.com/github/codespaces/issues/7101

PR #15, opened 2022-05-18
Try out package

chrmarti:
Updating the try out instructions. Mentioning the dependency on Python and C/C++.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,21 +20,19 @@ This CLI is in active development. Current status:
 
 ## Try it out
 
-You can try out the CLI in just a few steps. This repository has a [dev container configuration](https://github.com/devcontainers/cli/tree/main/.devcontainer), which you can use to ensure you have the right dependencies installed.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,6 +96,23 @@ Hello, VS Code Remote - Containers!
 {"outcome":"success"}
 ```
 
+Congrats, you've just run the dev container CLI and seen it in action!
+
+## Build from sources
```
bamurtaugh, line 101: Would we want this section before `Try out the CLI`, so that users know they can try it with the npm package or building from sources?

bamurtaugh (approved):
LGTM, thanks for adding this! One idea is if we want to move the info on how to build from sources a bit earlier so users are aware they can use the sample project in either way they get the CLI. But as I think about it, more advanced users would be interested in building, so 1) it may be nice to have it in its own section later, 2) advanced users would likely also understand how they can test with a project and use the steps above; so maybe we leave as-is.

README.md
chrmarti: The same README is used for the package on npmjs.org as with the repository on github.com. Not sure if there is a way to have distinct READMEs for the two. Giving the use on npmjs.org priority for now.

PR #19, opened 2022-05-18
New merge command to combine parent and child devcontainer.json configurations for inheritance 

jeremydelacruz:
# OCW Hack Project

## Description 

---

Work to help contribute to proposed devcontainer `extends` inheritance functionality from [this issue](https://github.com/devcontainers/spec/issues/22). Current state is a proof of concept that adds a new `merge` devcontainer CLI command, accepting "parent" and "child" devcontainer.json references. 

devcontainer build command will be enhanced with the following arguments:
**--parent-devcontainer** filepath to parent (or base) devcontainer JSON [string]
**--child-devcontainer** filepath to child (or extended) devcontainer JSON [string]

Also included are a set of default "rules" for all possible valid fields in the `DevContainerConfig` types.

## PR Checklist

---

- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My code follows the code style of this project.
- [ ] I ran the lint checks which produced no new errors nor warnings for my changes.
- [ ] I have checked to ensure there aren't other open Pull Requests for the same update/change.

## Does this introduce a breaking change?

---

- [ ] Yes
- [x] No

> If this introduces a breaking change, please describe the impact and migration path for existing applications below.

## Testing

>   - What OS was used for testing: Windows and MacOS
>   - Which test sets were used: mocka and manual testing
>   - Description of test scenarios that you have tried.
manual: Build images and upload them to container registry (docker hub)

Manual testing:

(add details here)

```
---


## Other information or known dependencies

---
n/a

### CODE REVIEW

File: src/spec-configuration/configuration.ts [hunk 1]
```diff
@@ -230,3 +239,48 @@ export async function getDockerComposeFilePaths(cliHost: FileHost, config: DevCo
 	}
 	return [];
 }
+
+export function buildExtendBehaviorTable(customBehaviors?: Partial<DevContainerExtendConfig>): DevContainerExtendConfig {
+	const defaultBehaviors: DevContainerExtendConfig = {
```
chrmarti, line 244: What makes this difficult to use is that the user can't tell by a simple, easy to remember rule how things are going to be merged because it takes a case-by-case approach.

joshspicer: I think their general idea (unless it has changed) is that arrays merged and objects/single values replaced by default.  However, this property can be overwritten (now sure how they landed on that) for specific use-cases.

chrmarti: Sounds good. I see a few arrays listed as "REPLACE" and am wondering if that is intended. (Also curious about the `customBehaviors` override, but didn't see that being used.)

greggroth: We've thus far talked about the logic behaving something like:

1. Hahes/Objects merge
2. Arrays merge
3. Strings/Ints/etc replace

Would it be possible to refactor this be based around the types rather than the specific keys?  I could imagine this being fragile if, for example, a single-value key is extended to support an array-value.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -48,6 +49,7 @@ const defaultDefaultUserEnvProbe: UserEnvProbe = 'loginInteractiveShell';
 	y.command('run-user-commands', 'Run user commands', runUserCommandsOptions, runUserCommandsHandler);
 	y.command('read-configuration', 'Read configuration', readConfigurationOptions, readConfigurationHandler);
 	y.command(restArgs ? ['exec', '*'] : ['exec <cmd> [args..]'], 'Execute a command on a running dev container', execOptions, execHandler);
+	y.command('merge [args..]', 'Merge a parent and a child dev container config', mergeOptions, mergeHandler);
```
greggroth, line 52: I think is this implicitly part of `read-configuration` and it should full-resolve any `extends` references in the configuration.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -803,6 +805,73 @@ async function doExec({
 	}
 }
 
+function mergeOptions(y: Argv) {
+	return y.options({
+		'parent-devcontainer': { type: 'string', description: 'Parent devcontainer path.' },
+		'child-devcontainer': { type: 'string', description: 'Child devcontainer path.' },
+	});
```
greggroth, line 812: I think this is taking a slightly different approach than what we've discussed with `"extends"` by having a parent and child declared as CLI arguments. The idea is that the devcontainer file itself would model this relationship with the `"extends"` keyword, e.g. 

```jsonc
// base.json
{
  "name": "Example",
  "ports": [80]
}

// devcontainer.json
{
  "extends": "./base.json",
  "ports": [22]
}
```

The CLI would resolve the "extends" keyword

```sh
$ devcontainer-cli read-configuration devcontainer.json
{
  "name": "Example",
  "ports": [22, 80]
}
```

PR #24, opened 2022-05-19
Add multi build support

juzuluag:
# [Work Item ID](https://github.com/microsoft/vscode-remote-release/issues/5652)

## Description

---

Enhance  `devcontainer` CLI **build** command by adding an option to build multi-platform container images and to push them to a container registry.

devcontainer build command will be enhanced with the following arguments:
**--platform** value1, value2.  Comma delimited string with platform types [string]. 
i.e. value: **linux/amd64,linux/arm64**
**--push** push to a container registry  [boolean] [default: false]

## PR Checklist

---

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] My code follows the code style of this project.
- [x] I ran the lint checks which produced no new errors nor warnings for my changes.
- [x] I have checked to ensure there aren't other open Pull Requests for the same update/change.

## Does this introduce a breaking change?

---

- [ ] Yes
- [x] No

> If this introduces a breaking change, please describe the impact and migration path for existing applications below.

## Testing

>   - What OS was used for testing: Window and MacOS
>   - Which test sets were used: mocka and manual testing
>   - Description of test scenarios that you have tried.
manual: Build images and upload them to container registry (docker hub)

Manual testing:

```bash
node devcontainer.js build --workspace-folder ../demo_container/  --platform linux/arm64,linux/amd64 --push --image-name dockerhubuname/test1:v1

[+] Building 1.9s (9/9) FINISHED                                                
 => [internal] load build definition from Dockerfile                       0.0s
 => => transferring dockerfile: 245B                                       0.0s
...
 => => pushing manifest for docker.io/dockerhubuname/test1:v1@sha256:00b62d3859  0.4s
 => [auth] dockerhubuname/test1:pull,push token for registry-1.docker.io         0.0s
 => [auth] dockerhubuname/test1:pull,push token for registry-1.docker.io         0.0s
````
List manifest from container registry

```bash
docker buildx imagetools inspect dockerhubuname/test1:v1
Name:      docker.io/dockerhubuname/test1:v1
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:00b62d385953096b5b09ba1458d0e2467b53b3c6d2409958a67eabd68f4fa0a0
           
Manifests: 
  Name:      docker.io/dockerhubuname/test1:v1@sha256:a18591ab9c4d416b9ee5b86c1b66c5ff9b89f8b8b522bc3d26db1395bd736f34
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64
             
  Name:      docker.io/dockerhubuname/test1:v1@sha256:4ae8b07ea9a8c61da4a06548d758fc27709408b79d52d293bdc0edb940041576
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64

```
---

## Other information or known dependencies

---
n/a


stuartleeks:
(disclaimer: I'm not on the VS Code team, so will defer to @chrmarti - my comments are those of an interested observer and someone who has made some contributions in similar areas of the code :smile:)

Hi @juzuluag, this looks pretty cool - being able to create multi-platform dev container images would be really useful I think. Especially building them in CI and publishing so that they can be referenced in the `cacheFrom` in `devcontainer.json` (or the equivalent `cache_from` in `docker-compose.yml`) and then work across multiple platforms

There has been some recent work here that shifts to using BuildKit/buildx by default (if it is present) for `Dockerfile`-based dev contaienrs (the work on `docker-compose` dev containers is still in-flight). Hopefully that will simplify the work to support multi-platform? 

With the buildx work, I think that `enableBuildx` could be dropped in favour of the `buildKitVersion` checks?
Similarly, the `buildxLoad` looks like it could be removed as the current `buildx` path includes the `--load`  argument when calling `docker`.

For the `--push` parameter, I'm curious whether there a particular benefit to adding the `--push` rather than the caller simply executing a `docker push` command after running the `devcontainer build` step?

src/spec-node/singleContainer.ts
```diff
@@ -159,15 +178,34 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 	}
 
 	const args: string[] = [];
-	if (buildParams.useBuildKit) {
+	if (buildParams.useBuildKit && !buildParams.enableBuildx) {
 		args.push('buildx', 'build',
 			'--load', // (short for --output=docker, i.e. load into normal 'docker images' collection)
 			'--build-arg', 'BUILDKIT_INLINE_CACHE=1', // ensure cache manifest is included in the image
 		);
+	} else if (buildParams.enableBuildx && argImageName) {
```
stuartleeks, line 186: My feeling here is that this would be better merged with the above branch, i.e. have a single branch that handles the BuildKit/buildx path and optionally adds the `--platform` args as needed

chrmarti: I agree, we no longer need a `--buildx` flag. We also `--load` by default, so maybe we only need `--platform` and `--push` then?

stuartleeks: Currently, when building, the CLI generates an image name based on the devcontainer folder etc and uses that to build. If the `--image-name` option is passed then the image is tagged with that name after the build.

So if the caller adds a `docker push` call after calling the `devcontainer` CLI then that will work. To add the `--push` option to the `devcontainer` CLI, I think we would need to change the CLI behaviour to use the name specified via `--image-name` so that the name of the image being built (and pushed) is correct?

chrmarti: My understanding is that `--push` (when passed to `buildx build`) would push at the end of the build to avoid `--load` because `--load` only works for the local platform. So we would have to use the correct tag on `buildx build`. (I just noticed that we can use multiple tags on the same build if we have to.)

stuartleeks: Yeah, ignore my previous comment. You're right, the `--load` will drop the multi-platform aspect of the image!

I'm not sure what the behaviour would be if we have multiple tags when building and also include the `--push` option, does it attempt to push all tags?

We could modify the `buildNamedImageAndExtend` function to take an `imageName` and then use that if it's set in preference of the [other options currently used](https://github.com/devcontainers/cli/blob/b6ee1258c203672be53b669183f2a3c35ed3da5f/src/spec-node/singleContainer.ts#L105)? This would save the later [`docker tag` step](https://github.com/devcontainers/cli/blob/main/src/spec-node/devContainersSpecCLI.ts#L331)

juzuluag: Hi @stuartleeks, @chrmarti . These are great comments. Thanks so much. With regards the comment about `--load`, it is not possible to build more than one image platform at the time for instance, so kind of useful for knowing that the image build works for a specific platform.
Also `--push` is quite useful to just push to a container registry with all the platforms specified (>= 1) without the need to create one at the time and keep track of the manifest that encapsulates all the images. (one manifest uploaded with all the images references too), which is what this enhancement aims for.

chrmarti: Sounds good, so we only need `--platform` and `--push`. I think without `--push` we would still default to `--load` and potentially having multiple tags is fine. With `--push` we can omit the automatically generated tag and only use the user-specified one (unless I'm missing something).

stuartleeks: Is there any reason not to pass the value of `--image-name` into the `buildImageAndExtend` function, defaulting to the current generated name if no value is specified? Then we could just apply a single tag, and skip the `docker tag` step in the CLI file

juzuluag: > Is there any reason not to pass the value of `--image-name` into the `buildImageAndExtend` function, defaulting to the current generated name if no value is specified? Then we could just apply a single tag, and skip the `docker tag` step in the CLI file

@chrmarti, your comments sounds reasonable. For the image name, it is needed for `--push` since it is the container registry name and not the autogenerated name. I'll make the changes. Thanks to both of you for all the comments.

stuartleeks: @juzuluag I agree that the image name specified in the command args needs to be passed through.

My question above was around whether there is any reason that in this case we would also need the generated image name? Or whether, when the user has specified the image name, it would be sufficient to just apply that one as a tag.

If so, I think it simplifies the code around specifying the tags as it no longer needs to account for whether `--push` has been specified.

juzuluag: Hi @chrmarti  and @stuartleeks . I made some changes, so let me know if these are more aligned with your thoughts.

juzuluag: hmm, I would defer to @chrmarti . I just did some changes, so hopefully that gives us more insights. Thanks

src/spec-node/singleContainer.ts
```diff
@@ -160,14 +161,24 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
...
 	if (buildParams.buildKitVersion) {
-		args.push('buildx', 'build',
-			'--load', // (short for --output=docker, i.e. load into normal 'docker images' collection)
-			'--build-arg', 'BUILDKIT_INLINE_CACHE=1', // ensure cache manifest is included in the image
-		);
+		if (buildParams.buildxPlatform && buildParams.buildxPush && argImageName) {
+			args.push('buildx', 'build');
+			args.push('--platform', buildParams.buildxPlatform);
+			args.push('--push');
+			args.push('-f', finalDockerfilePath, '-t', argImageName);
```
stuartleeks, line 168: I think we probably want the `--build-arg BUILDKIT_INLINE_CACHE=1` as per the section below so that the cache metadata is written included in the image to enable rebuilds to re-use cached layers even when not present on the build machine

src/spec-node/singleContainer.ts
```diff
@@ -160,14 +161,24 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
...
+			args.push('buildx', 'build',
+				'--load', // (short for --output=docker, i.e. load into normal 'docker images' collection)
+				'--build-arg', 'BUILDKIT_INLINE_CACHE=1', // ensure cache manifest is included in the image
+			);
+		}
 	} else {
 		args.push('build');
 	}
-	args.push('-f', finalDockerfilePath, '-t', baseImageName);
+	if (!argImageName) {
```
stuartleeks, line 178: If this method just took a single image name (i.e. the `buildNamedImageAndExtend` used the image name it is passed and falls back to the previous `'image' in config ? config.image : getFolderImageName(params.common)` value if not set) then this if block wouldn't be needed. I think it would also allow the added code above to be more similar between the single and multi-platform options?

juzuluag: @stuartleeks , thanks for the suggestion. Let me know if this is what you had in mind. PR updated.

src/spec-node/singleContainer.ts
```diff
@@ -101,23 +101,25 @@ async function setupContainer(container: ContainerDetails, params: DockerResolve
...
 	if (isDockerFileConfig(config)) {
-		return await buildAndExtendImage(params, config, imageName, params.buildNoCache ?? false);
+		return await buildAndExtendImage(params, config, imageName, params.buildNoCache ?? false, argImageName);
 	}
 	// image-based dev container - extend
 	return await extendImage(params, config, imageName, 'image' in config);
 }
-async function buildAndExtendImage(buildParams: DockerResolverParameters, config: DevContainerFromDockerfileConfig, baseImageName: string, noCache: boolean) {
+
+async function buildAndExtendImage(buildParams: DockerResolverParameters, config: DevContainerFromDockerfileConfig, baseImageName: string, noCache: boolean, argImageName?: string) {
```
stuartleeks, line 114: We still have two image names being passed in here - we could simplify back to a single name by handling the default value in `buildNamedImageAndExtend`? (see [here](https://github.com/juzuluag/cli/pull/1))

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -184,6 +184,8 @@ async function provision({
 		remoteEnv: keyValuesToRecord(addRemoteEnvs),
 		additionalCacheFroms: addCacheFroms,
 		useBuildKit: buildkit,
+		buildxPlatform: undefined,
```
stuartleeks, line 187: Could we add tests that exercise the `--platform` option?

Tests for `--push` might need a bit more thought as we need a registry to push to?

src/spec-node/singleContainer.ts
chrmarti: Maybe add both tags (when not pushing)? Is there an advantage of having the generated tag when later doing a `devcontainer up` in CI?

juzuluag: I looked at the simplified version and made the changes. But either `--load` or `--push` has to be passed not both. 
This is the error if you use both: `error: push and load may not be set together at the moment`

src/spec-node/devContainersSpecCLI.ts
juzuluag: I was looking into that, but couldn't find any utests that exercises the args (i.e `cache-from, buildkit`... ). Any pointers or suggestions are welcome.

src/spec-node/singleContainer.ts
```diff
@@ -160,14 +162,19 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 
 	const args: string[] = [];
 	if (buildParams.buildKitVersion) {
-		args.push('buildx', 'build',
-			'--load', // (short for --output=docker, i.e. load into normal 'docker images' collection)
-			'--build-arg', 'BUILDKIT_INLINE_CACHE=1', // ensure cache manifest is included in the image
-		);
+		args.push('buildx', 'build');
+		if (buildParams.buildxPlatform && buildParams.buildxPush && argImageName) {
+			args.push('--platform', buildParams.buildxPlatform);
```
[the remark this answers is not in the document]
juzuluag: For the unsupported platform, docker fails eventually with something like this (this error is specific of my test dockerfile)
```sh
error: failed to solve: mcr.microsoft.com/vscode/devcontainers/python:3.10-bullseye: no match for platform in manifest sha256:21a12816fcadaa16dabb4ba0e8c358361d02ea062b1b89db8786eb67173489d0: not found
Error: Command failed:
```

For adding test to platform, is there any utest that can be enhanced for other args?

src/spec-node/devContainersSpecCLI.ts
juzuluag: nvm, I found some at `cli.test.ts`

stuartleeks: Unresolving as I can't see any added tests 😄

juzuluag: @stuartleeks , @chrmarti , latest [commit](https://github.com/devcontainers/cli/pull/24/commits/cc6b381972409434e2d9052fab4942a5c8cfaba2) has utest. But seems like buildkit is not set, so I was wondering if any of you can give me some hints here, since the test that needs to fail, it is not. Thanks for your help.

juzuluag:
Hi @chrmarti and @stuartleeks , any chance to look at the most recent changes? Thanks.

src/spec-node/singleContainer.ts
```diff
@@ -159,14 +164,21 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
...
+		}
+		if (buildParams.buildxPush) {
+			args.push('--push');
+		} else {
+			args.push('--load'); // (short for --output=docker, i.e. load into normal 'docker images' collection)
+		}
+		args.push('--build-arg', 'BUILDKIT_INLINE_CACHE=1');
 	} else {
 		args.push('build');
```
chrmarti, line 178: Can we fail here (when BuildKit is not available) when any of the new options are set? That will make it easier for the user to understand what is happening.

chrmarti (commented):
Thanks @juzuluag. Added a few more comments.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -325,10 +333,12 @@ async function doBuild({
 		if (isDockerFileConfig(config)) {
 
 			// Build the base image and extend with features etc.
-			const { updatedImageName } = await buildNamedImageAndExtend(params, config);
+			const { updatedImageName } = await buildNamedImageAndExtend(params, config, argImageName);
```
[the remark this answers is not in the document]
juzuluag: Unfortunately, this command requires access to the dockerfile.

samruddhikhandale:
Could we be able to prioritize this PR? cc @chrmarti 

I am working on [devcontainers/images](https://github.com/devcontainers/images) to create images based on features by using the devcontainers/cli. We are looking forward to publish these images, but can't yet as they need to be built with `--progress` flag which this PR targets

PR #27, opened 2022-05-20
Sanitize feature input options to be set as `_BUILD_ARG..`

joshspicer:
Through experimentation and via the web with out auto-generated `_BUILD_ARG`s, it appears :

`
Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.
`

This transforms the outputted `_BUILD_ARG_..` in an opinionated way.

cc/ @greggroth

### DIFF
--- a/src/test/container-features/helpers.offline.test.ts
+++ b/src/test/container-features/helpers.offline.test.ts
@@ -6,14 +6,29 @@ export const output = makeLog(createPlainLog(text => process.stdout.write(text),
 
 describe('getIdSafe should return safe environment variable name', function () {
 
-    it('should replace all "-" with "_"', function () {
+    it('should replace a "-" with "_"', function () {
         const ex = 'option-name';
         assert.strictEqual(getSafeId(ex), 'OPTION_NAME');
     });
 
-    it('should replace all leading numbers', function () {
-        const ex = '12_option-name';
-        assert.strictEqual(getSafeId(ex), '___OPTION_NAME');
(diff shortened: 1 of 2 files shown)


chrmarti:
Does this break existing features? (Should we have a version that uses old and new names, so features can update?)

joshspicer:
I noticed this due to a crash when the option name contained a `-`.  Previously, we were just changing the case of the feature name.  I don't think this will break anyone. In fact, I think this is more consistent with the intellisense we provide

src/test/container-features/helpers.offline.test.ts
```diff
@@ -1,8 +1,37 @@
...
+    it('should delete a leading numbers and add a _', function () {
+        const ex = '1name';
+        assert.strictEqual(getSafeId(ex), '_NAME');
+    });
+
+    it('should delete all leading numbers and add a _', function () {
+        const ex = '12345_option-name';
+        assert.strictEqual(getSafeId(ex), '_OPTION_NAME');
+    });
+});
```
greggroth, line 33: <3 the unit tests. Makes it very clear what the behavior is

greggroth (approved):

chrmarti (approved):

PR #37, opened 2022-05-24
Speed up tests

stuartleeks:
Convert `beforeEach` to `before`
For example, exec tests need the container to be started before running but each test for a configuration doesn't require the container to be started/stopped.

In timings tests on my local machine, this drops the test time from 3m20 to 2m20. Hopefully this will help with the test timings as more tests are added.

### DIFF
--- a/src/test/cli.test.ts
+++ b/src/test/cli.test.ts
@@ -71,7 +71,7 @@ describe('Dev Containers CLI', function () {
 	});
 
 	describe('Command up', () => {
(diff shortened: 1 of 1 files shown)


joshspicer:
@stuartleeks is this still something you think we want to get in?

stuartleeks:
@joshspicer - whilst it would be nice to speed up the tests, I've not managed to get any time to look at this for a while. I'll close it and we can refer back to it if we revisit this in future

PR #41, opened 2022-05-31
Handle string format for build in getBuildInfoForService

stuartleeks:
Rename `getInfoForService` -> `getBuildInfoForService`
Handle `build` being a string or an object
Add tests for `getBuildInfoForService`

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -108,8 +108,10 @@ export function getRemoteWorkspaceFolder(config: DevContainerFromDockerComposeCo
 	return config.workspaceFolder || '/';
 }
 
-function getInfoForService(composeService: any) {
-	// composeService should taken from readDockerComposeConfig which normalises the yml format
+// exported for testing
+export function getBuildInfoForService(composeService: any) {
+	// composeService should taken from readDockerComposeConfig
+	// the 'build' property can be a string or an object (https://docs.docker.com/compose/compose-file/build/#build-definition)
 
 	const image = composeService.image as string | undefined;
 	const composeBuild = composeService.build;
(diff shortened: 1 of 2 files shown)


src/spec-node/dockerCompose.ts
```diff
@@ -118,10 +120,19 @@ function getInfoForService(composeService: any) {
...
+				dockerfilePath: 'Dockerfile'
+			}
+		};
+	}
 	return {
 		image,
 		build: {
-			dockerfilePath: composeBuild.dockerfile as string,
+			dockerfilePath: (composeBuild.dockerfile as string | undefined) ?? 'Dockerfile',
 			context: composeBuild.context as string,
```
chrmarti, line 136: Is context required?

src/spec-node/dockerCompose.ts
```diff
@@ -140,7 +151,7 @@ export async function buildAndExtendDockerCompose(config: DevContainerFromDocker
 	// determine base imageName for generated features build stage(s)
 	let baseName = 'dev_container_auto_added_stage_label';
 	let dockerfile = null;
-	const serviceInfo = getInfoForService(composeService);
+	const serviceInfo = getBuildInfoForService(composeService);
 	if (serviceInfo.build) {
 		const { context, dockerfilePath, target } = serviceInfo.build;
 		const resolvedDockerfilePath = cliHost.path.isAbsolute(dockerfilePath) ? dockerfilePath : path.resolve(context, dockerfilePath);
```
chrmarti, line 157: Can context be relative?

src/spec-node/dockerCompose.ts
stuartleeks: Yes, according to the docs: https://docs.docker.com/compose/compose-file/build/#context-required

src/spec-node/dockerCompose.ts
stuartleeks: Just checked the docs, and it can be. Will update to handle that

chrmarti (approved):

PR #44, opened 2022-06-01
Regenerate all overrides as needed if not exist

stuartleeks:
Refactor the handling when a container exists but the persisted files do not to handle scenarios such as the build override not existing and needing recreating.

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -6,7 +6,7 @@
 import * as yaml from 'js-yaml';
 import * as shellQuote from 'shell-quote';
 
-import { createContainerProperties, startEventSeen, ResolverResult, getTunnelInformation, DockerResolverParameters, createFeaturesTempFolder, inspectDockerImage, ensureDockerfileHasFinalStageName } from './utils';
+import { createContainerProperties, startEventSeen, ResolverResult, getTunnelInformation, DockerResolverParameters, inspectDockerImage, ensureDockerfileHasFinalStageName } from './utils';
 import { ContainerProperties, setupInContainer, ResolverProgress } from '../spec-common/injectHeadless';
 import { ContainerError } from '../spec-common/errors';
 import { Workspace } from '../spec-utils/workspaces';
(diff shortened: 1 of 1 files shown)


src/spec-node/dockerCompose.ts
```diff
@@ -327,32 +330,18 @@ async function startContainer(params: DockerResolverParameters, buildParams: Doc
...
-				if (overrideFilePath) {
-					// Add file path to override file as parameter
-					composeGlobalArgs.push('-f', overrideFilePath);
-				}
-			}
-		}
-	} else {
+	}
+
+	if (!container || !didRestoreFromPersistedShare) {
```
chrmarti, line 335: There is a case where we have a "build" but no "start" file. In that case `if (persistedBuildFile && persistedStartFile) {` would not hold and `didRestoreFromPersistedShare` would be false.

src/spec-node/dockerCompose.ts
```diff
@@ -322,12 +333,16 @@ async function startContainer(params: DockerResolverParameters, buildParams: Doc
 			const files = configFiles?.split(',') ?? [];
 			const persistedBuildFile = await checkForPersistedFile(buildCLIHost, output, files, featuresBuildOverrideFilePrefix);
 			const persistedStartFile = await checkForPersistedFile(buildCLIHost, output, files, featuresStartOverrideFilePrefix);
-			if (persistedBuildFile && persistedStartFile) {
+			if ((persistedBuildFile.foundLabel && persistedBuildFile.fileExists) // require build file if in label
```
chrmarti, line 336: Should this be `(!persistedBuildFile.foundLabel || persistedBuildFile.fileExists)`?

stuartleeks: Yes - thanks!

src/spec-node/dockerCompose.ts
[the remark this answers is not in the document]
stuartleeks: Is this a duplicate of the [comment above](https://github.com/devcontainers/cli/pull/44#pullrequestreview-997611901)?

chrmarti (approved):

PR #45, opened 2022-06-01
Enable explicit CacheFrom when no cacheFrom in devcontainer.json

stuartleeks:
Fix bug that only allowed `--cache-from` args to be applied when the `devcontainer.json` contained a `build.cacheFrom` entry


chrmarti (approved):

chrmarti:
Thanks!

PR #51, opened 2022-06-09
Implement optional default values in localEnv/containerEnv expansions

leopoldsedev:
This is related to #50 and might even close it.


chrmarti (approved):

joshspicer (dismissed):

joshspicer:
Thanks!

chrmarti (approved):

chrmarti:
Thanks @leopoldsedev !

mraarone (commented):
These changes sufficiently seek out and implement default values. There is an assumption that the user will only input a single extra split for colons, and there is no limitation on how many can be inputted. Perhaps this is taken care of elsewhere, but I do not see a maximum evaluateSingleVariable, "variable" length or limit on "parts".

PR #61, opened 2022-06-24
Build command: Support multiple --image-name parameters 

samruddhikhandale:
It allows to tag the same image more than once.

eg - `devcontainer build --workspace-folder ./ --image-name alpine3 --image-name alpine3.0`

This is mainly needed for multiple tagging of multi-arch images created with `--platform`.
If the cli creates a multi arch image, and if you retag with `docker tag` then the new image ends up using architecture of the current machine for tagging.

Hence, such multi-arch images needs to be tagged during image build else they won't support multi-architecture platform.

### DIFF
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -328,7 +328,7 @@ async function doBuild({
 			throw new ContainerError({ description: `Dev container config (${uriToFsPath(configFile || getDefaultDevContainerConfigPath(cliHost, workspace!.configFolderPath), cliHost.platform)}) not found.` });
 		}
 		const { config } = configs;
-		let imageNameResult = [''];
+		let imageNameResult: string[] = [''];
 		
 		// Support multiple use of `--image-name` 
 		const imageNames = (argImageName && (Array.isArray(argImageName) ? argImageName : [argImageName]) as string[]) || undefined;
@@ -337,13 +337,10 @@ async function doBuild({
 
 			// Build the base image and extend with features etc.
 			let { updatedImageName } = await buildNamedImageAndExtend(params, config, imageNames);
-			updatedImageName = (updatedImageName && (Array.isArray(updatedImageName) ? updatedImageName : [updatedImageName]) as string[]) || [''];
 
 			if (imageNames) {
-				if (!buildxPush && typeof updatedImageName === 'string') {
-					imageNames.forEach(async function (image) {
-						await dockerPtyCLI(params, 'tag', JSON.stringify(updatedImageName), image);
-					});
+				if (!buildxPush) {
+					await Promise.all(imageNames.map(imageName => dockerPtyCLI(params, 'tag', updatedImageName[0], imageName)));
 				}
 				imageNameResult = imageNames;
 			} else {
@@ -381,9 +378,7 @@ async function doBuild({
 			const originalImageName = service.image || `${projectName}_${config.service}`;
 
 			if (imageNames) {
-				imageNames.forEach(async function (image) {
-					await dockerPtyCLI(params, 'tag', originalImageName, image);
-				});
+				await Promise.all(imageNames.map(imageName => dockerPtyCLI(params, 'tag', originalImageName, imageName)));
 				imageNameResult = imageNames;
 			} else {
 				imageNameResult = originalImageName;
@@ -397,9 +392,7 @@ async function doBuild({
 				throw new ContainerError({ description: '--platform or --push require dockerfilePath.' });
 			}
 			if (imageNames) {
-				imageNames.forEach(async function (image) {
-					await dockerPtyCLI(params, 'tag', updatedImageName[0], image);
-				});
+				await Promise.all(imageNames.map(imageName => dockerPtyCLI(params, 'tag', updatedImageName[0], imageName)));
 				imageNameResult = imageNames;
 			} else {
 				imageNameResult = updatedImageName;
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -398,11 +398,11 @@ async function doBuild({
 			}
 			if (imageNames) {
 				imageNames.forEach(async function (image) {
-					await dockerPtyCLI(params, 'tag', updatedImageName, image);
+					await dockerPtyCLI(params, 'tag', updatedImageName[0], image);
 				});
 				imageNameResult = imageNames;
 			} else {
-				imageNameResult = [updatedImageName];
+				imageNameResult = updatedImageName;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


Chuxel:
Love this idea!  Super useful given how you have to use tags for versioning. 👍

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -328,18 +328,24 @@ async function doBuild({
 			throw new ContainerError({ description: `Dev container config (${uriToFsPath(configFile || getDefaultDevContainerConfigPath(cliHost, workspace!.configFolderPath), cliHost.platform)}) not found.` });
 		}
 		const { config } = configs;
-		let imageNameResult = '';
+		let imageNameResult = [''];
```
chrmarti, line 331: Use `let imageNameResult: string[];` that will let the compiler check all code paths assign to it before it is accessed.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -328,18 +328,24 @@ async function doBuild({
...
 		if (isDockerFileConfig(config)) {
 
 			// Build the base image and extend with features etc.
-			const { updatedImageName } = await buildNamedImageAndExtend(params, config, argImageName);
-
-			if (argImageName) {
-				if (!buildxPush) {
-					await dockerPtyCLI(params, 'tag', updatedImageName, argImageName);
+			let { updatedImageName } = await buildNamedImageAndExtend(params, config, imageNames);
+			updatedImageName = (updatedImageName && (Array.isArray(updatedImageName) ? updatedImageName : [updatedImageName]) as string[]) || [''];
```
chrmarti, line 340: This line shouldn't be necessary, shouldn't it always be an array? (The handling of argImageName above is an exception due to the typing being incomplete.)

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -328,18 +328,24 @@ async function doBuild({
...
+
+			if (imageNames) {
+				if (!buildxPush && typeof updatedImageName === 'string') {
+					imageNames.forEach(async function (image) {
```
chrmarti, line 344: This wouldn't wait. Try something like:
await Promise.all(imageNames.map(imageName => dockerPtyCLI(params, 'tag', updatedImageName[0], imageName)));

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -374,9 +380,11 @@ async function doBuild({
 			const service = composeConfig.services[config.service];
 			const originalImageName = service.image || `${projectName}_${config.service}`;
 
-			if (argImageName) {
-				await dockerPtyCLI(params, 'tag', originalImageName, argImageName);
-				imageNameResult = argImageName;
+			if (imageNames) {
+				imageNames.forEach(async function (image) {
```
chrmarti, line 384: See above.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -388,9 +396,11 @@ async function doBuild({
 			if (buildxPlatform || buildxPush) {
 				throw new ContainerError({ description: '--platform or --push require dockerfilePath.' });
 			}
-			if (argImageName) {
-				await dockerPtyCLI(params, 'tag', updatedImageName, argImageName);
-				imageNameResult = argImageName;
+			if (imageNames) {
+				imageNames.forEach(async function (image) {
```
chrmarti, line 400: See above.

src/spec-node/singleContainer.ts
```diff
@@ -180,7 +180,11 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 	} else {
 		args.push('build');
 	}
-	args.push('-f', finalDockerfilePath, '-t', baseImageName);
+	args.push('-f', finalDockerfilePath);
+	
+	baseImageNames.forEach(function(image) {
```
chrmarti, line 185: Use arrow function for consistency.

src/spec-node/devContainersSpecCLI.ts
samruddhikhandale: That's neat. Modified, thanks

src/spec-node/devContainersSpecCLI.ts
samruddhikhandale: Makes sense. Modified.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -328,18 +328,24 @@ async function doBuild({
...
-			const { updatedImageName } = await buildNamedImageAndExtend(params, config, argImageName);
-
-			if (argImageName) {
-				if (!buildxPush) {
-					await dockerPtyCLI(params, 'tag', updatedImageName, argImageName);
+			let { updatedImageName } = await buildNamedImageAndExtend(params, config, imageNames);
+			updatedImageName = (updatedImageName && (Array.isArray(updatedImageName) ? updatedImageName : [updatedImageName]) as string[]) || [''];
+
+			if (imageNames) {
+				if (!buildxPush && typeof updatedImageName === 'string') {
```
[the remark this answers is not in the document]
samruddhikhandale: Right, Modified.

src/spec-node/devContainersSpecCLI.ts
samruddhikhandale: Ah! Modified., thanks

src/spec-node/singleContainer.ts
samruddhikhandale: Modified.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -388,9 +391,9 @@ async function doBuild({
 			if (buildxPlatform || buildxPush) {
 				throw new ContainerError({ description: '--platform or --push require dockerfilePath.' });
 			}
-			if (argImageName) {
-				await dockerPtyCLI(params, 'tag', updatedImageName, argImageName);
-				imageNameResult = argImageName;
+			if (imageNames) {
+				await Promise.all(imageNames.map(imageName => dockerPtyCLI(params, 'tag', updatedImageName[0], imageName)));
```
joshspicer, line 395: It is not clear to me why we always want the first element of `updatedImageName`.  Does `extendImage` return an array with a single value in it? If that's the case, can it just return the element itself?

samruddhikhandale: >  Does extendImage return an array with a single value in it?

Yes, that's the case

> If that's the case, can it just return the element itself?

If we make `extendImage` return a single element, then things get complex because [buildNamedImageAndExtend](https://github.com/devcontainers/cli/blob/65044246c2d73176004cebdd6c00dd2cd9483896/src/spec-node/singleContainer.ts#L106) (which calls `extendImage` based on a condition) then returns a `string | string[]`. We have to fix all compiler issues and also we aren't sure if we want to to use `updatedImageName` or `updatedImageName[]`.

Ref - https://github.com/devcontainers/cli/pull/61/commits/1b598c36c8282628ae99a16277cba2af603a4894 which made extendImage return [] because tests were failing and that change simplied the compiler checks (commit ref)[https://github.com/devcontainers/cli/pull/61/commits/65044246c2d73176004cebdd6c00dd2cd9483896]

Issue #70, opened 2022-06-29
Regression: Feature entrypoints run as remoteUser


Chuxel:
From https://github.com/microsoft/vscode-dev-containers/issues/1517. It appears that feature entrypoints are run as `remoteUser` instead of the containerUser as expected. In the repro below, commenting out `remoteUser` causes things to function - the only reason sudo would get tripped is if the script is not being run as root.

//cc @joshspicer @edgonmsft @jkeech @chrmarti @bamurtaugh @harrywithers

---

- VSCode Version: 1.68.1
- Local OS Version: macOS Monterey 12.3.1
- Local chip architecture: Intel Core i7
- Reproduces in: Remote - Containers
- Name of Dev Container Definition with Issue: docker-in-docker

Steps to Reproduce:

1. Create a devcontainer file and docker file with the following configuration
```json
{
    "name": "name",
    "build": {
        "dockerfile": "./Dockerfile"
    },
    "remoteUser": "node",
    "features": {
        "docker-in-docker": "latest",
    }
}
```
```Dockerfile
FROM node:16.14.0

```
2. Reopen in dev container
3. Run `docker ps` in the dev container

Result:
```txt
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

But expected docker to be up and running in the container.

Docker container log:
```txt
Container started
/usr/local/share/docker-init.sh: 65: /usr/local/share/docker-init.sh: sudo: not found
```

Docker version output in dev container:
```txt
Client:
 Version:           20.10.17+azure-1
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c70180fde3601def79a59cc3e996aa553c9b9
 Built:             Mon Jun  6 21:36:39 UTC 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Docker version output locally
```txt
Client:
 Cloud integration: v1.0.25
 Version:           20.10.16
 API version:       1.41
 Go version:        go1.17.10
 Git commit:        aa7e414
 Built:             Thu May 12 09:20:34 2022
 OS/Arch:           darwin/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.9.1 (81317)
 Engine:
  Version:          20.10.16
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.10
  Git commit:       f756502
[...]

PR #74, opened 2022-07-01
Use previous USER (devcontainers/cli#70)

chrmarti:


src/spec-node/utils.ts
```diff
@@ -332,17 +332,44 @@ export function getLocalCacheFolder(): string {
...
+const findFromLines = new RegExp(/^(?<line>\s*FROM.*)/, 'gm');
+const parseFromLine = /FROM\s+(?<platform>--platform=\S+\s+)?(?<image>\S+)(\s+[Aa][Ss]\s+(?<label>[^\s]+))?/;
+const findUserLines = new RegExp(/^\s*USER\s+(?<user>\S+)/, 'gm');
+
+export async function getImageUser(params: DockerResolverParameters, dockerfile: string) {
+	return internalGetImageUser(imageName => inspectDockerImage(params, imageName, true), dockerfile);
+}
+
+export async function internalGetImageUser(inspectDockerImage: (imageName: string) => Promise<ImageDetails>, dockerfile: string) {
+	// TODO: Other targets.
```
alexr00, line 344: Worth linking an issue here so that the todo doesn't get lost?

alexr00 (approved):

PR #93, opened 2022-07-29
`features package <...>` command

joshspicer:
Implements the packaging step of the [proposed feature specification](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features-distribution.md#packaging).

This PR also refactors the `features <test> ..` command into it's own file for tidyness.


.vscode/launch.json
```diff
@@ -10,10 +10,11 @@
             "name": "Launch CLI - up",
             "program": "${workspaceFolder}/src/spec-node/devContainersSpecCLI.ts",
             "cwd": "${workspaceFolder}",
+            // "envFile": "${workspaceFolder}/.env",  // Create .env file in the workspace folder to pass environment variables to the CLI.
             "args": [
                 "up",
                 "--workspace-folder",
-                "../devcontainers-features",
+                "/home/node/example-project",
```
jkeech, line 17: Is this path guaranteed to exist? What's the reason for moving away from launching against the `../devcontainers-features`

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,90 @@
...
+import { UnpackArgv } from '../devContainersSpecCLI';
+import { getPackageConfig } from '../utils';
+import { doFeaturesPackageCommand } from './packageCommandImpl';
+
+export function featuresPackageOptions(y: Argv) {
+    return y
+        .options({
+            // 'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to auto-detect all features in collection directory.  Cannot be combined with \'-s\'.', },
+            'features-folder': { type: 'string', alias: 'c', default: '.', description: 'Path to folder containing features source code' },
+            'output-dir': { type: 'string', alias: 'o', default: '/tmp/build', description: 'Path to output' },
```
jkeech, line 17: Should the default also be the current working directory?

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,90 @@
...
+            'log-level': { choices: ['info' as 'info', 'debug' as 'debug', 'trace' as 'trace'], default: 'info' as 'info', description: 'Log level.' },
+            'quiet': { type: 'boolean', alias: 'q', default: false, description: 'Quiets output' },
+        })
+        .check(_argv => {
+            // if (argv['scenarios'] && argv['features']) {
```
jkeech, line 22: Nit: Remove commented out code

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,90 @@
...
+    'features-folder': featuresFolder,
+    'log-level': inputLogLevel,
+    'output-dir': outputDir,
+}: FeaturesPackageArgs) {
+    const disposables: (() => Promise<unknown> | undefined)[] = [];
+    const dispose = async () => {
+        await Promise.all(disposables.map(d => d()));
+    };
+
+    const extensionPath = path.join(__dirname, '..', '..', '..');
```
jkeech, line 53: Does this work at runtime once the CLI has been packaged up for distribution?

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,90 @@
...
+    const cwd = process.cwd();
+    const cliHost = await getCLIHost(cwd, loadNativeModule);
+    const output = createLog({
+        logLevel: mapLogLevel(inputLogLevel),
+        logFormat: 'text',
+        log: (str) => process.stdout.write(str),
+        terminalDimensions: undefined,
+    }, pkg, new Date(), disposables);
+
+    output.write('Packaging features...\n', LogLevel.Info);
```
jkeech, line 65: Do we need the extra new line character here? I notice that the other `output.write` calls don't have it.

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const featuresDirResolved = cliHost.path.resolve(featuresFolder);
+    if (!(await isLocalFolder(featuresDirResolved))) {
+        throw new Error(`Features folder '${featuresDirResolved}' does not exist`);
+    }
+
+    const outputDirResolved = cliHost.path.resolve(outputDir);
+    if (!(await isLocalFolder(outputDirResolved))) {
+        // TODO: Delete folder first?
```
jkeech, line 74: I don't think we should delete the output folder. That seems dangerous and likely unexpected for the user running this command. Shouldn't we be able to drop the new file into the specified folder without deleting it? If so, let's remove this comment.

src/spec-node/featuresCLI/packageCommandImpl.ts
```diff
@@ -0,0 +1,84 @@
+import path from 'path';
+import tar from 'tar';
+import { Feature } from '../../spec-configuration/containerFeaturesConfiguration';
+import { LogLevel } from '../../spec-utils/log';
+import { isLocalFile, readLocalDir, readLocalFile, writeLocalFile } from '../../spec-utils/pfs';
+import { FeaturesPackageCommandInput } from './package';
+
```
jkeech, line 7: Remove extra blank lines here and below

src/spec-node/featuresCLI/packageCommandImpl.ts
```diff
@@ -0,0 +1,84 @@
...
+    return 0;
+}
+
+async function tarDirectory(featureFolder: string, archiveName: string, outputDir: string) {
+    return new Promise<void>((resolve) => resolve(tar.create({ file: path.join(outputDir, archiveName), cwd: featureFolder }, ['.'])));
+}
+
+export async function getFeaturesAndPackage(args: FeaturesPackageCommandInput): Promise<Feature[] | undefined> {
+    const { output, featuresFolder, outputDir } = args;
+    const featureDirs = await readLocalDir(featuresFolder);
```
jkeech, line 56: Should we also support a mode where you directly package a single feature directory without pointing to the parent directory and collecting all subdirectories as individual features? I imagine that a lot of people will start with publishing a single feature, so optimizing to make that use case simple would be great.

src/spec-node/featuresCLI/packageCommandImpl.ts
```diff
@@ -0,0 +1,84 @@
...
+    let metadatas: Feature[] = [];
+
+    for await (const f of featureDirs) {
+        output.write(`Processing feature: ${f}...`, LogLevel.Info);
+        if (!f.startsWith('.')) {
+            const featureFolder = path.join(featuresFolder, f);
+            const archiveName = `devcontainer-feature-${f}.tgz`;
+
+            await tarDirectory(featureFolder, archiveName, outputDir);
```
jkeech, line 65: Can you move this down to below the validations? We should check for the existence of a `devcontainer-feature.json` and an `install.sh` before tarring up the directory. If one of those does not exist, print a warning (probably not an error unless you are trying the single feature case mentioned above?) and skip that folder.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -49,14 +49,16 @@ const defaultDefaultUserEnvProbe: UserEnvProbe = 'loginInteractiveShell';
 	y.command('build [path]', 'Build a dev container image', buildOptions, buildHandler);
 	y.command('run-user-commands', 'Run user commands', runUserCommandsOptions, runUserCommandsHandler);
 	y.command('read-configuration', 'Read configuration', readConfigurationOptions, readConfigurationHandler);
-	y.command('features', 'Features commands', (y: Argv) =>
-		y.command('test', 'Test features', featuresTestOptions, featuresTestHandler));
+	y.command('features', 'Features commands', (y: Argv) => {
+		y.command('test', 'Test features', featuresTestOptions, featuresTestHandler);
+		y.command('package', 'Package features', featuresPackageOptions, featuresPackageHandler);
```
jkeech, line 54: Could you add some tests for this new command?

jkeech (commented):
Thanks for adding this command! It seems like this will be really helpful for feature authors to package and distribute their features.

I left a few comments/suggestions.

.vscode/launch.json
joshspicer: The previous path also doesn't exist by default.  I think it's worth checking in a scaffold project (and maybe .gitignore it so people can edit/debug configs from the barebones template).

src/spec-node/featuresCLI/package.ts
joshspicer: maybe it generates an `output` directory in the output dir? I didn't want it to just spit out 20 tgz files in the root.

src/spec-node/featuresCLI/package.ts
joshspicer: it does, this is done all over the codebase (And I just confirmed by packaging it locally)

src/spec-node/featuresCLI/packageCommandImpl.ts
joshspicer: I am not so sure that it is simpler to have two different flags. The interface would then look like (for a single feature in some folder named `src`)

`devcontainer feature package -f ./src/myFeature` 

or

`devcontainer feature package -c ./src`

...both of which would output exactly the same given only one feature in `src`.

If I then add another feature to `./src`, it no longer makes sense to use the former command, since in the spec say features colocated in a repo will be a part of a "collection" and share the same "metadata 


We have the spec structured in a way that we always will generate a metadata artifact, even if there is only a single feature in the 'collection'.  We **wouldn't** want people to run:

`devcontainer feature package -f ./src/myFeature && devcontainer feature package -f ./src/myFeature2`

,since that would generate two `devcontainer-collection.json` files with 1 feature in each.

jkeech: I wasn't thinking that we would need different command line flags. Rather, if users run the command with a directory that points directly to a feature (e.g. it has a `devcontainer-feature.json` and `install.sh` in the directory), then we would package up that feature into a single-feature collection.

Basically I want to keep it simple for repos that are just trying to publish a single feature, which will likely be the majority of user-contributed features. I expect many people will try to run the command from the feature directory, and the command today is expecting a very specific folder structure which features might not all be in.

joshspicer: hmm gotcha, that makes sense. I will add that in that auto-detection bit (and perhaps remove the flag, instead opting for a positional argument to supply the path if you aren't invoking from the current directory)

src/spec-node/featuresCLI/test.ts
```diff
@@ -0,0 +1,87 @@
...
+    'log-level': inputLogLevel,
+}: FeaturesTestArgs) {
+    const disposables: (() => Promise<unknown> | undefined)[] = [];
+    const dispose = async () => {
+        await Promise.all(disposables.map(d => d()));
+    };
+
+    const cwd = process.cwd();
+    const cliHost = await getCLIHost(cwd, loadNativeModule);
+    const extensionPath = path.join(__dirname, '..', '..');
```
alexdima, line 64: Does this need another '..' now that it's been moved into the `featuresCLI` folder?

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,98 @@
+import path from 'path';
```
alexdima, line 1: Consider converting newly added `.ts` files to use tabs instead of spaces to align with the current sources (`package.ts`, `packageCommandImpl.ts`, `test.ts`, `testCommandImpl.ts`, maybe some the file containing the actual test under `src/test` ?).

This doesn't seem to be enforced in the CI, but it might be good to try to keep things consistent until we add something to enforce it. (we could add `"editor.insertSpaces": false` to the `.vscode/settings.json` file, which will make newly created files default to using tabs in vscode).

src/spec-node/featuresCLI/packageCommandImpl.ts
```diff
@@ -0,0 +1,87 @@
...
+            const featureMetadata: Feature = JSON.parse(await readLocalFile(featureJsonPath, 'utf-8'));
+            metadatas.push(featureMetadata);
+        }
+    }
+
+    if (metadatas.length === 0) {
+        return;
+    }
+
+    output.write(`Packaged ${metadatas.length.toString()} features!`, LogLevel.Info);
```
alexdima, line 85: small nit: `.toString()` is not necessary here.

alexdima (changes requested):
Looks really good! I left a few comments.

src/spec-node/featuresCLI/test.ts
joshspicer: Good catch, i'll add a test as well

src/spec-node/featuresCLI/packageCommandImpl.ts
joshspicer: addressed this 🚀

src/spec-node/featuresCLI/package.ts
```diff
@@ -0,0 +1,119 @@
...
+import { Log, LogLevel, mapLogLevel } from '../../spec-utils/log';
+import { isLocalFile, isLocalFolder, mkdirpLocal, rmLocal } from '../../spec-utils/pfs';
+import { createLog } from '../devContainers';
+import { UnpackArgv } from '../devContainersSpecCLI';
+import { getPackageConfig } from '../utils';
+import { doFeaturesPackageCommand } from './packageCommandImpl';
+
+const targetPositionalDescription = `
+Package features at provided [target] (default is cwd), where [target] is either:
+   1. A collection folder containing a './src' folder with [1..n] features.
```
chrmarti, line 14: Why not supply the src-folder's path directly? This would avoid the requirement that the features are located in a `src` folder (e.g., they could be in a subfolder of `src` or the folder could be named differently like `features-src` or the repo root could contain the feature folders directly).

joshspicer: I was attempting to keep consistent with `features test ...`, which expects that folder layout.  I thought it may be less confusion to always operate from the "root" of the collection (this would be the root of the `devcontainers/features` repo, for example.

joshspicer: i've made the change to directly expect the src folder, whereever that may be.

chrmarti: Maybe align `features test` with this then?

chrmarti (approved):

src/spec-node/featuresCLI/package.ts
joshspicer: The way I have it wired up, `features test` looks for two mirrored folders `src` and `test`.  An improvement there is certainly to let a user indicate both the src and test folders independently, in which case it would line up better with this command.  

Either way, not a huge issue and simple enough 👍

PR #94, opened 2022-08-01
Initial version of OCI packages for features.

edgonmsft:
The code here implements the latest changes to the features proposal as written here:

https://github.com/devcontainers/spec/pull/66

The code parses the feature id and decides its origin between the new OCI features and old github release based features by verifiying a manifest is available and reachable for said feature.

The main TODO is implementing authentication support for the various operating system keychains and obtain the docker credentials.


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -412,6 +425,17 @@ export async function generateFeaturesConfig(params: { extensionPath: string; cw
 	return featuresConfig;
 }
 
+async function prepareGHCRauthAndOCICache(dstFolder: string) {
```
joshspicer, line 428: should this function be renamed and my POC comments removed?

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -650,6 +681,17 @@ async function fetchFeatures(params: { extensionPath: string; cwd: string; outpu
 			const featureDebugId = `${feature.consecutiveId}_${sourceInfoType}`;
 			params.output.write(`* Fetching feature: ${featureDebugId}`);
 
+			if (sourceInfoType === 'oci') {
+				params.output.write(`Fetching from OCI`, LogLevel.Trace);
+				await mkdirpLocal(featCachePath);
+				const success = await fetchOCIFeature(params.output, params.env, featureSet, ociCacheDir, featCachePath);
+				if (!success) {
+					// TODO: FAIL
```
joshspicer, line 689: If we don't know how exactly to fail yet, at the very least should add a `LogLevel.Error` so we can catch it later?

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -869,3 +911,35 @@ function getFeatureValueDefaults(feature: Feature) {
...
+	if (id.includes('@')) {
+		return { type: 'github-repo', manifest: undefined };
+	}
+
+	const manifest = await validateOCIFeature(output, env, id);
+	if (manifest) {
+		return { type: 'oci', manifest: manifest };
+	}
+
+	return { type: 'github-repo', manifest: undefined };
```
joshspicer, line 944: This is also deprecated, no?

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    return featureSet;
+}
+
+export function getFeatureRef(output: Log, identifier: string): OCIFeatureRef {
+
+    // ex: ghcr.io/codspace/features/ruby:1
+    const splitOnColon = identifier.split(':');
+    const id = splitOnColon[0];
+    const version = splitOnColon[1] ? splitOnColon[1] : '1';
```
joshspicer, line 67: In the spec (and in our publishing tooling) we can emphasize the convention we always publish a `latest` tag (for tags that indicate a new latest breakfix version according to semver), and hardcode version-less features to `latest` instead of 1?

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    const splitOnSlash = id.split('/');
+    const featureName = splitOnSlash[splitOnSlash.length - 1];
+    const owner = splitOnSlash[1];
+    const registry = splitOnSlash[0];
+    const namespace = splitOnSlash.slice(1, -1).join('/');
+
+    output.write(`identifier: ${identifier}`);
```
joshspicer, line 75: Maybe set these to `LogLevel.Trace`?

joshspicer: I've been wiring in a `log-level` argument to all the new feature-related commands and the existing commands already allow us to specify log level.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+        owner,
+        namespace,
+        registry
+    };
+}
+
+// Validate if a manifest exists and is reachable about the declared feature.
+export async function validateOCIFeature(output: Log, env: NodeJS.ProcessEnv, identifier: string): Promise<OCIManifest | undefined> {
+    const featureRef = getFeatureRef(output, identifier);
+
```
joshspicer, line 96: Is this the spot where we should check media type?  I can imagine people trying to use _some_ other oci artifact as a feature and our tooling getting confused.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+// Download a feature from which a manifest was previously downloaded.
+export async function fetchOCIFeature(output: Log, env: NodeJS.ProcessEnv, featureSet: FeatureSet, ociCacheDir: string, featCachePath: string): Promise<boolean> {
+
+    if (featureSet.sourceInformation.type !== 'oci') {
+        output.write(`FeatureSet is not an OCI featureSet.`);
+        throw new Error('FeatureSet is not an OCI featureSet.');
+    }
+
+    const blobUrl = `https://${featureSet.sourceInformation.featureRef.registry}/v2/${featureSet.sourceInformation.featureRef.namespace}/${featureSet.sourceInformation.featureRef.featureName}/blobs/${featureSet.sourceInformation.manifest?.layers[0].digest}`;
```
joshspicer, line 112: Ensuring the feature declares media type will also allow us to be confident that `featureSet.sourceInformation.manifest?.layers[0].digest` and any other fields of fields exist.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+    output.write(`blob url: ${blobUrl}`, LogLevel.Trace);
+
+    const success = await getFeatureBlob(output, env, blobUrl, ociCacheDir, featCachePath);
+
+    if (!success) {
+        output.write(`Failed to download package for ${featureSet.sourceInformation.featureRef.featureName}`, LogLevel.Error);
+        throw new Error(`Failed to download package for ${featureSet.sourceInformation.featureRef.featureName}`);
```
joshspicer, line 119: I know the pattern @chrmarti has followed is generally to never throw and handle errors in the caller.  I do agree though that this is probably fatal enough that we should stop execution

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,197 @@
...
+        await tar.x(
+            {
+                file: tempTarballPath,
+                cwd: featCachePath,
+            }
+        );
+
+        return true;
+    } catch (e) {
+        output.write(`error: ${e}`, LogLevel.Trace);
```
joshspicer, line 181: I think we should always bubble up fatal errors with `LogLevel.Error`. Easier for us in the future with bug reports.

src/test/container-features/containerFeaturesOCI.offline.test.ts
```diff
@@ -0,0 +1,46 @@
+import { assert } from 'chai';
+import { getFeatureManifest, getFeatureBlob, getFeatureRef } from '../../spec-configuration/containerFeaturesOCI';
+import { createPlainLog, LogLevel, makeLog } from '../../spec-utils/log';
+
+export const output = makeLog(createPlainLog(text => process.stdout.write(text), () => LogLevel.Trace));
+
+describe('Test OCI', () => {
+
```
joshspicer, line 8: I wonder if the github action token could somehow be used for an authenticated pull 🤔.  
It would be very nice to test a non-public oci resource as well. (Perhaps for a future as this is just the initial scaffolding).

joshspicer (approved):
Looks great, super excited for this!  We just gotta make sure the test is properly fetching (and once we open source `devcontainers/features`, can point the tests to those features.)

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: Yep, just to confirm, which one is the correct media type?

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: Yeah my thought is that it would make it clearer for the user exactly where and why things failed, its probably missing from other places as well.

src/spec-configuration/containerFeaturesConfiguration.ts
edgonmsft: Fixed.

src/spec-configuration/containerFeaturesConfiguration.ts
edgonmsft: Added.

src/spec-configuration/containerFeaturesConfiguration.ts
edgonmsft: Added a note there as well.

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: Yep missed it.

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: I think I fixed/added all of them.

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: It depends which layer you're looking at, but as an example the full manifest is:

So we'd probably want to check `config.mediaType` == `application/vnd.devcontainers`.

```
{
  "digest": "sha256:61b4349617cd5eb5b18d5e42498a842b409b8367acf4ade1815fd176587479fd",
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "size": 449,
  "config": {
    "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "mediaType": "application/vnd.devcontainers",
    "size": 0
  },
  "layers": [
    {
      "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
      "mediaType": "application/vnd.oci.image.layer.v1.tar",
      "size": 0
    },
    {
      "digest": "sha256:c33008d0dc12d0e631734082401bec692da809eae2ac51e24f58c1cac68fc0c9",
      "mediaType": "application/octet-stream",
      "size": 16896
    }
  ]
}
```

joshspicer (approved):

src/test/container-features/containerFeaturesOCI.offline.test.ts
edgonmsft: I left authentication as a TODO, I think the idea would be (like the oras client) to use a token if we have and it mantches, if we don't fall back to anonymous and if that fails try to obtain a token.

We should also cache the tokens so that they can be reused.

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: Added the validation.

src/spec-configuration/containerFeaturesOCI.ts
edgonmsft: Added.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+        featureName,
+        owner,
+        namespace,
+        registry
+    };
+}
+
+// Validate if a manifest exists and is reachable about the declared feature.
+export async function validateOCIFeature(output: Log, env: NodeJS.ProcessEnv, identifier: string): Promise<OCIManifest | undefined> {
+    const featureRef = getFeatureRef(output, identifier);
```
alexdima, line 95: Should we early return here if `featureRef.registry` doesn't look like a domain name? Otherwise some of the `github-repo` features might lead to a HTTP request, which might take a long time to resolve, depending how DNS is set up. (i.e. maybe the code will do a request to `https://devcontainers/v2/...`, which will first attempt to resolve `devcontainers`)

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -433,19 +453,17 @@ function featuresToArray(config: DevContainerConfig): DevContainerFeature[] | un
...
 				featuresConfig.featureSets.push(newFeatureSet);
 			}
-		}
-	);
+	}
 	return featuresConfig;
 }
 
-export function parseFeatureIdentifier(output: Log, userFeature: DevContainerFeature) : FeatureSet | undefined {
+export async function parseFeatureIdentifier(output: Log, env: NodeJS.ProcessEnv, userFeature: DevContainerFeature): Promise<FeatureSet | undefined> {
```
alexdima, line 466: Should we update the large comment below to reflect the new OCI format? I would also suggest to perhaps move `getFeatureIdType` up from the bottom of the file, to have it closer to this comment since its logic follows this comment.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+async function getAuthenticationToken(env: NodeJS.ProcessEnv, output: Log, registry: string, id: string): Promise<string> {
+    // TODO: Use operating system keychain to get credentials.
+    // TODO: Fallback to read docker config to get credentials.
+
+    // TEMP for ghcr.io
+    const githubToken = env['GITHUB_TOKEN'];
+    if (registry !== 'ghcr.io' || !githubToken) {
```
alexdima, line 196: Adding knowledge about `GITHUB_TOKEN` which makes this work for `ghcr.io` seamlessly in places like github actions is nice, but then we should consider adding support for env variables for other registries.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+        namespace,
+        registry
+    };
+}
+
+// Validate if a manifest exists and is reachable about the declared feature.
+export async function validateOCIFeature(output: Log, env: NodeJS.ProcessEnv, identifier: string): Promise<OCIManifest | undefined> {
+    const featureRef = getFeatureRef(output, identifier);
+
+    const manifestUrl = `https://${featureRef.registry}/v2/${featureRef.namespace}/${featureRef.featureName}/manifests/${featureRef.version}`;
```
alexdima, line 97: I'd love to have some links here or in other places where we construct URLs and headers to the relevant OCI spec (so we can follow along that we implement the spec correctly).

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+    output.write(`manifest url: ${manifestUrl}`, LogLevel.Trace);
+    const manifest = await getFeatureManifest(output, env, manifestUrl, featureRef);
+
+    if (manifest?.config.mediaType !== 'application/vnd.devcontainers') {
+        return undefined;
```
alexdima, line 102: I suggest to add a warning log statement here. Otherwise, if someone publishes a feature and forgets to use this media type they will struggle to figure out what's wrong.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+
+// Download a feature from which a manifest was previously downloaded.
+export async function fetchOCIFeature(output: Log, env: NodeJS.ProcessEnv, featureSet: FeatureSet, ociCacheDir: string, featCachePath: string, featureRef: OCIFeatureRef): Promise<boolean> {
+
+    if (featureSet.sourceInformation.type !== 'oci') {
+        output.write(`FeatureSet is not an OCI featureSet.`, LogLevel.Error);
+        throw new Error('FeatureSet is not an OCI featureSet.');
+    }
+
+    const blobUrl = `https://${featureSet.sourceInformation.featureRef.registry}/v2/${featureSet.sourceInformation.featureRef.namespace}/${featureSet.sourceInformation.featureRef.featureName}/blobs/${featureSet.sourceInformation.manifest?.layers[0].digest}`;
```
alexdima, line 116: Another great place to add a comment with a link to the relevant OCI spec.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -869,3 +907,36 @@ function getFeatureValueDefaults(feature: Feature) {
...
+
+export async function getFeatureIdType(output: Log, env: NodeJS.ProcessEnv, id: string) {
+
+	// DEPRECATED: This is a legacy feature-set ID
+	if (!id.includes('/') && !id.includes('\\')) {
+		return { type: 'local-cache', manifest: undefined };
+	}
+
+	// Direct tarball reference
+	if (id.startsWith('http://') || id.startsWith('https://')) {
```
alexdima, line 919: We should really disallow `http`. Not sure when we added it and if it is safe to remove without adding backwards compatibility, but `http` is so prone to MITM attacks that I think we should just completely disallow it.

alexdima (changes requested):
Looks really good! I left a few comments, and one more general one would be that we might consider adding some simple retry logic to these `https` requests.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    return '';
+}
+
+export async function getAuthToken(output: Log, registry: string, id: string) {
+    const headers = {
+        'user-agent': 'devcontainer',
+    };
+
+    const url = `https://${registry}/token?scope=repo:${id}:pull&service=ghcr.io`;
```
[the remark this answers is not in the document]
edgonmsft: Thats a mistake actually, it should be the registry itself.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -187,34 +208,232 @@ export async function getFeatureBlob(output: Log, env: NodeJS.ProcessEnv, url: s
...
+    // Obtain session ID with `/v2/<namespace>/blobs/uploads/` 
+    const blobPutLocationUriPath = await postUploadSessionId(output, featureRef, registryAuthToken);
+    if (!blobPutLocationUriPath) {
+        output.write(`Failed to get upload session ID`, LogLevel.Error);
+        return false;
+    }
+
+    for await (const blob of blobsToPush) {
+        const { name, digest } = blob;
+        const blobExistsConfigLayer = await checkIfBlobExists(output, featureRef, digest, registryAuthToken);
```
alexdima, line 299: `blobExistsConfigLayer` is IMHO a bit misleading here as it looks like the loop will also handle the `tgzLayer` blob, not just the `configLayer` blob.

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -187,34 +208,232 @@ export async function getFeatureBlob(output: Log, env: NodeJS.ProcessEnv, url: s
...
+        output.write(`blob: ${name} with digest ${digest}  ${blobExistsConfigLayer ? 'already exists' : 'does not exist'} in registry.`, LogLevel.Trace);
+
+        // PUT blobs
+        if (!blobExistsConfigLayer) {
+            putBlob(output, blobPutLocationUriPath, featureRef, digest, registryAuthToken);
+        }
+
+        // Send a final PUT to combine blobs and tag manifest properly.
```
alexdima, line 307: The comment suggests that this final `PUT` should run after both blobs have been uploaded, outside the `for` loop, but it currently sits inside the `for` loop 🤔 .

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -187,34 +208,232 @@ export async function getFeatureBlob(output: Log, env: NodeJS.ProcessEnv, url: s
...
+
+    // Success!
+    return true;
+}
+
+// Spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests (PUT /manifests/<ref>)
+export async function putManifestWithTags(output: Log, manifest: string, tags: string[]) {
+    output.write(`Tagging manifest with tags: ${tags.join(', ')}`, LogLevel.Trace);
+    output.write(manifest);
+    throw new Error('Not implemented');
```
alexdima, line 320: Should we implement this before merging?

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -187,34 +208,232 @@ export async function getFeatureBlob(output: Log, env: NodeJS.ProcessEnv, url: s
...
+    const url = `https://${featureRef.registry}/v2/${featureRef.namespace}/${featureRef.featureName}/blobs/${digest}`;
+    const statusCode = await headRequest({ url, headers }, output);
+
+    output.write(`${url}: ${statusCode}`, LogLevel.Trace);
+    return statusCode === 200;
+}
+
+    // Spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#post-then-put
+    //       Requires registry auth token.
+    export async function postUploadSessionId(output: Log, featureRef: OCIFeatureRef, authToken: string): Promise<string | undefined> {
```
alexdima, line 395: A couple of functions in this file are indentend in a weird way, like `checkIfBlobExists`, `postUploadSessionId`.

src/test/container-features/containerFeaturesOCI.offline.test.ts
```diff
@@ -1,10 +1,13 @@
+// import { assert } from 'chai';
```
alexdima, line 1: A couple of unnecessary commented out imports

alexdima (changes requested):
Left more comments

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: Moving forward we'll want to read the OS keychain, etc to read the more standard .docker/config.json file.  

I do like this idea as temporary flexibility, or in cases where it would just be easier to set a secret (like in actions?)

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: agreed! I have already done this in a lot of places for the push portion (in a separate PR).  I'll sprinkle some more in here.

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer: done and moved! Also linking to our spec now

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer: definitely worth removing.  I don't believe it will cause any impact and improves the integrity of downloads for a user.

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: moved to another PR

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: moved to another PR

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: yes :) moved to another PR

edgonmsft:
LGTM

alexdima (approved):
Nice!

joshspicer (approved):

PR #97, opened 2022-08-02
Introducing the interface for `features publish` command

samruddhikhandale:
Builds on https://github.com/devcontainers/cli/pull/94 and https://github.com/devcontainers/cli/pull/99

Introducing the interface for `features publish` command. 
- Packs the features
- Computes semantic versions for the features
-  Consumes changes in https://github.com/devcontainers/cli/pull/99 
- Added tests

image

PR #99, opened 2022-08-02
OCI push

joshspicer:
Builds on #94 

Relevant to : https://github.com/devcontainers/spec/pull/66

Implements the relevant parts of the OCI distribution-spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md

Example of manually triggering `pushOCIFeature` with `--log-level trace`.  Note that here all the blobs had already been uploaded to the registry previously, so the CLI did not try to republish them.  The result can be seen publicly here: https://github.com/users/joshspicer/packages/container/package/mypackages%2Fcolor

image

image

This PR provides the distribution spec implementation. The user-facing portion is being completed separately in PR https://github.com/devcontainers/cli/pull/97


src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -7,262 +7,289 @@ import { FeatureSet } from './containerFeaturesConfiguration';
```
joshspicer, line 7: Also converted this file to tabs.  I suggest disabling 'show whitespace changes' in the GitHub PR UI.

Issue #88, opened 2022-07-18
Remove builtin features


alexdima:
Stop shipping features defined in [vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers/releases) and switch to rely entirely on feature resolving as defined via https://github.com/devcontainers/spec/issues/7

samruddhikhandale:
Working on CLI changes to automatically map the old feature names to their new locations for back compatibility. 
Also, mapping deprecated features from [here](https://github.com/devcontainers/features/issues/64) to `microsoft/vscode-dev-containers`

samruddhikhandale:
Draft PR - https://github.com/devcontainers/cli/pull/100

Waiting for

- [ ] https://github.com/devcontainers/cli/pull/94

PR #100, opened 2022-08-02
Auto mapping feature ids from old shorthand syntax to devcontainers/features

samruddhikhandale:
Addresses - https://github.com/devcontainers/cli/issues/88 and https://github.com/devcontainers/cli/issues/88

Mapping of old shorthand syntax of features
- Migrated feature ids are mapped to `ghcr.io/devcontainers/features/${id}:1`
- Renamed feature ids are mapped to `ghcr.io/devcontainers/features/${id}:1`
- Deprecated feature ids are kept as it is

Added tests? 
Yes! 

**Dependency -** 
Should not be merged until we have started pointing people to new repository and pulled off the plug from https://github.com/microsoft/vscode-dev-containers


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -505,12 +505,38 @@ export async function getFeatureIdType(output: Log, env: NodeJS.ProcessEnv, id:
 	return { type: 'github-repo', manifest: undefined };
 }
 
+export function getBackwardCompatibleFeatureId(id: string) {
```
samruddhikhandale, line 508: 💀 Should not be merged until we are ready for open sourcing /images and /features repositories

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -514,14 +514,42 @@ export async function getFeatureIdType(output: Log, env: NodeJS.ProcessEnv, user
...
+	renamedFeatures.set('golang', 'go');
+	renamedFeatures.set('common', 'common-utils');
+
+	const newFeaturePath = 'ghcr.io/devcontainers/features';
+	// Note: Pin the versionBackwardComp to '1' to avoid breaking changes.
+	const versionBackwardComp = '1';
+
+	// Mapping feature references (old shorthand syntax) from "microsoft/vscode-dev-containers" to "ghcr.io/devcontainers/features"
+	if (migratedfeatures.includes(id)) {
+		return `${newFeaturePath}/${id}:${versionBackwardComp}`;
```
chrmarti, line 529: We are using `@` as the separator (unless this changed, but the parsing below still uses that).

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -514,14 +514,42 @@ export async function getFeatureIdType(output: Log, env: NodeJS.ProcessEnv, user
...
+	}
+
+	// Mapping feature references (renamed old shorthand syntax) from "microsoft/vscode-dev-containers" to "ghcr.io/devcontainers/features"
+	if (renamedFeatures.get(id) !== undefined) {
+		return `${newFeaturePath}/${renamedFeatures.get(id)}:${versionBackwardComp}`;
```
chrmarti, line 534: Use `@` as separator before version.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -514,14 +514,42 @@ export async function getFeatureIdType(output: Log, env: NodeJS.ProcessEnv, user
 	}
 }
 
+export function getBackwardCompatibleFeatureId(id: string) {
+	const migratedfeatures = ['aws-cli', 'azure-cli', 'desktop-lite', 'docker-in-docker', 'docker-from-docker', 'dotnet', 'git', 'git-lfs', 'github-cli', 'java', 'kubectl-helm-minikube', 'node', 'powershell', 'python', 'ruby', 'rust', 'sshd', 'terraform'];
```
chrmarti, line 518: I see fish, gradle, homebrew, jupyterlab, maven are missing, assuming these are deprecated, should we add a log message when one of these is used?

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: According the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature) changes for OCI, we now use `:` 😄 

> `<oci-registry>/<namespace>/<feature>[:<semantic-version>]`
> eg. ghcr.io/user/repo/go:1

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Same comment as above!

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer: We're trying to conform to the OCI spec for identifiers (you should be to just do an `oras pull <identifier>`).  

Tags are preceded by a `:`.  One thing we _don't_ support right now is `...@<digest>`, which we should do and update the spec accordingly.

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Sounds good. Added a message. 
Once features repo is open source, I'm thinking of adding more detailed log messages for these as well as migrated/renamed features

src/spec-configuration/containerFeaturesConfiguration.ts
chrmarti: Sounds good, thanks.

chrmarti (approved):

chrmarti (changes requested):
Removing approval to discuss the option to let the new functionality get more testing before we redirect existing configurations to use it.

chrmarti (approved):
Reapproving since Remote-Containers stable version 0.245.0 is released. (As discussed.)

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: I removed the warning log messages, as per today's discussion. 

I'll add them back only when auto-mapping is ready to be put in Remote-Containers STABLE version

@chrmarti  does that sound correct to you?

chrmarti: Makes sense, thanks!

chrmarti (approved):

PR #103, opened 2022-08-03
Bug fix: v1 features - Incorrect V1_DEVCONTAINER_FEATURES_FILE_NAME

samruddhikhandale:
With v0.9.0, image builds with features following the 'v1' pattern are failing. Fixing that! 

image


samruddhikhandale:
Was confused for a bit with above error and `v1` features reference. 
The features (in the above error) uses `v2` pattern but does not contain OCI path, hence the failure.

PR #105, opened 2022-08-04
update `direct-tarball` to follow spec + refactor `processFeatureIdentifier` tests

joshspicer:
Implements https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features-distribution.md#directly-reference-tarball

Attempts to clean up relevant block of tests to make the new spec changes more clear


src/test/container-features/helpers.test.ts
```diff
@@ -1,274 +0,0 @@
-import { assert } from 'chai';
```
joshspicer, line 1: This was originally the contents of `featureHelpers.test.ts` that I renamed and refactored

chrmarti (approved):

PR #107, opened 2022-08-05
Add more examples of using the CLI

Chuxel:
This PR adds in scripts and a README that illustrates how to use the CLI to:

- Inject tools for use inside a development container
- Use a dev container as your CI build environment to build an application (even if it is not deployed as a container)
- Build a container image from a devcontainer.json file that includes dev container features

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ The [example-usage](./example-usage) folder contains some simple shell scripts t
 
 - Inject tools for use inside a development container
 - Use a dev container as your CI build environment to build an application (even if it is not deployed as a container)
-- Build a container image from a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md)
+- Build a container image from a devcontainer.json file that includes [dev container features](https://containers.dev/implementors/features/#devcontainer-json-properties)
 
 ## Build from sources
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ The [example-usage](./example-usage) folder contains some simple shell scripts t
 
 - Inject tools for use inside a development container
 - Use a dev container as your CI build environment to build an application (even if it is not deployed as a container)
-- Build a container image rom a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md).
+- Build a container image from a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md)
 
 ## Build from sources
 
--- a/README.md
+++ b/README.md
@@ -100,6 +100,14 @@ Hello, VS Code Remote - Containers!
 
 Congrats, you've just run the dev container CLI and seen it in action!
 
+## More CLI examples
+
+The [example-usage](./example-usage) folder contains some simple shell scripts to illustrate how the CLI can be used to:
+
+- Inject tools for use inside a development container
+- Use a dev container as your CI build environment to build an application (even if it is not deployed as a container)
+- Build a container image rom a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md).
+
 ## Build from sources
 
 This repository has a [dev container configuration](https://github.com/devcontainers/cli/tree/main/.devcontainer), which you can use to ensure you have the right dependencies installed.
--- a/example-usage/README.md
+++ b/example-usage/README.md
@@ -4,7 +4,7 @@ This folder contains a set of basic examples that use the devcontainer CLI for d
 
 1. [Use three different tools from a development container](#tool-examples)
 2. [Use a dev container as your CI build environment](#ci-build-environment-example) (even if your app is not deployed as a container)
-3. [Build a container image](#building-an-image-from-devcontainerjson) from a devcontainer.json file that includes [dev container features](https://containers.dev/implementors/features/)
+3. [Build a container image](#building-an-image-from-devcontainerjson) from a devcontainer.json file that includes [dev container features](https://containers.dev/implementors/features/#devcontainer-json-properties)
 
 Each should run on macOS or Linux. For Windows, you can use these scripts from WSL2.
 
@@ -39,7 +39,7 @@ This also adds a desktop to the container that can be accessed from a web browse
 
 These examples demonstrate the use of the dev container CLI to:
 
-1. Simplify setup using the "dev container features" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
+1. Simplify setup using the "[dev container features](https://containers.dev/implementors/features/#devcontainer-json-properties)" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
 
 2. How the dev container CLI can be used to inject tools without building them into the base image:
 
@@ -65,7 +65,7 @@ The initial build can take a bit since it is building the dev container image, w
 
 This example demonstrates the use of the dev container CLI to:
 
-1. Simplify setup using the "dev container features" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
+1. Simplify setup using the "[dev container features](https://containers.dev/implementors/features/#devcontainer-json-properties)" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
 
 2. Execute an application build script inside a dev container as follows:
 
--- a/example-usage/README.md
+++ b/example-usage/README.md
@@ -19,7 +19,7 @@ Each should run on macOS or Linux. For Windows, you can use these scripts from W
 
 ## Using the examples
 
-All examples use the contents of the `workspace` folder for their configuration, which is where you can modifications if you'd like. The example scripts are then in different sub-folders. 
+All examples use the contents of the `workspace` folder for their configuration, which is where you can make modifications if you'd like. The example scripts are then in different sub-folders. 
 
 ### Tool examples
 
@@ -50,6 +50,8 @@ These examples demonstrate the use of the dev container CLI to:
 Currently the `appPort` property is used in `devcontainer.json` instead of `forwardPorts` due to a gap in the current dev container CLI ([see here](https://github.com/devcontainers/cli/issues/22)).
 
 ### CI build environment example
+
+This example illustrates how you can use the dev container CLI to build your application in any CI system. (Note there is also a [GitHub Action](https://github.com/marketplace/actions/devcontainers-ci) and [Azure DevOps task](https://marketplace.visualstudio.com/items?itemName=devcontainers.ci) if you are using those automation systems, but this example will focus on direct use of the CLI.)
 
 You can use the example by opening a terminal and typing the following:
 
--- a/example-usage/README.md
+++ b/example-usage/README.md
@@ -41,7 +41,7 @@ This example demonstrates the use of the dev container CLI to:
 
 1. Simplify setup using the "dev container features" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
 
-2. How the dev container CLI can be used to inject tools without building them into the base image as follows:
+2. How the dev container CLI can be used to inject tools without building them into the base image:
 
     1. Use `devcontainer up` to spin up the container and mount a `server` and `workspace` folder into the container.
     2. Use `devcontainer exec` to run a script from this mounted folder to set up the appropriate server.
--- a/example-usage/README.md
+++ b/example-usage/README.md
@@ -1,12 +1,12 @@
 # Dev Container CLI Examples
 
-This repository provides a basic examples of using the devcontainer CLI for different use cases. It includes example scripts to:
+This folder contains a set of basic examples that use the devcontainer CLI for different use cases. It includes example scripts to:
 
 1. [Use three different tools from a development container](#tool-examples)
 2. [Use a dev container as your CI build environment](#ci-build-environment-example) (even if your app is not deployed as a container)
 3. [Build a container image](#building-an-image-from-devcontainerjson) from a devcontainer.json file that includes [dev container features](https://containers.dev/implementors/features/)
 
-Each should run on macOS or Linux (including WSL2 on Windows).
+Each should run on macOS or Linux. For Windows, you can use these scripts from WSL2.
 
 ## Pre-requisites
 
--- a/example-usage/README.md
+++ b/example-usage/README.md
@@ -1,0 +1,92 @@
+# Dev Container CLI Examples
+
+This repository provides a basic examples of using the devcontainer CLI for different use cases. It includes example scripts to:
+
+1. [Use three different tools from a development container](#tool-examples)
+2. [Use a dev container as your CI build environment to build an application](#ci-build-environment-example) (even if it is not deployed as a container)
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -100,6 +100,14 @@ Hello, VS Code Remote - Containers!
 
 Congrats, you've just run the dev container CLI and seen it in action!
 
+## More CLI examples
+
+The [example-usage](./example-usage) folder contains some simple shell scripts to illustrate how the CLI can be used to:
+
+- Inject tools for use inside a development container
+- Use a dev container as your CI build environment to build an application (even if it is not deployed as a container)
+- Build a container image rom a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md).
```
bamurtaugh, line 109: nit: typo + punctuation consistency across bullets
```suggestion
- Build a container image from a devcontainer.json file that includes [dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md)
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
+# Dev Container CLI Examples
+
+This repository provides a basic examples of using the devcontainer CLI for different use cases. It includes example scripts to:
```
bamurtaugh:
```suggestion
This folder provides basic examples of using the dev container CLI for different use cases. It includes example scripts to:
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. [Use three different tools from a development container](#tool-examples)
+2. [Use a dev container as your CI build environment to build an application](#ci-build-environment-example) (even if it is not deployed as a container)
+3. [Build a container image rom a devcontainer.json file](#building-an-image-from-devcontainerjson) that includes [dev container features](https://containers.dev/implementors/features/)
```
bamurtaugh:
```suggestion
3. [Build a container image from a devcontainer.json file](#building-an-image-from-devcontainerjson) that includes [dev container features](https://containers.dev/implementors/features/)
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Each should run on macOS or Linux (including WSL2 on Windows).
+
+## Pre-requisites
+
+1. Install Node.js 16 (e.g., using [nvm](https://github.com/nvm-sh/nvm))
+2. Install [node-gyp pre-requisites](https://github.com/nodejs/node-gyp):
+   - **Linux:** Use your distro's package manager. E.g. on Ubuntu/Debian: `sudo apt-get update && sudo apt-get install python3-minimal gcc g++ make`
+   - **macOS:** Install the XCode Command Line Tools ([more info](https://github.com/nodejs/node-gyp/blob/main/README.md#on-macos))
+3. Make sure you have an OpenSSH compliant `ssh` command available and in your path if you plan to use the `Vim via SSH` example. (It should already be there on macOS, and in Linux/WSL, you can install `openssh-client` using your distro's package manager if its missing.)
```
bamurtaugh, line 17:
```suggestion
3. Make sure you have an OpenSSH compliant `ssh` command available and in your path if you plan to use the `Vim via SSH` example (it should already be there on macOS, and in Linux/WSL, you can install `openssh-client` using your distro's package manager if its missing)
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+3. Install the latest dev container CLI: `npm install -g @devcontainers/cli`
+
+## Using the examples
+
+All examples use the contents of the `workspace` folder for their configuration if you want to make any modifications. The example scripts are then in different sub-folders.
```
bamurtaugh, line 22:
```suggestion
All examples use the contents of the `workspace` folder for their configuration, which is where you can modifications if you'd like. The example scripts are then in different sub-folders. 
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+3. Make sure you have an OpenSSH compliant `ssh` command available and in your path if you plan to use the `Vim via SSH` example. (It should already be there on macOS, and in Linux/WSL, you can install `openssh-client` using your distro's package manager if its missing.)
+3. Install the latest dev container CLI: `npm install -g @devcontainers/cli`
+
+## Using the examples
+
+All examples use the contents of the `workspace` folder for their configuration if you want to make any modifications. The example scripts are then in different sub-folders. 
+
+### Tool examples
+
+You can use the example by opening a terminal and typing one of the following:
```
bamurtaugh, line 26:
```suggestion
You can use these examples by opening a terminal and typing one of the following:
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+-  `tool-vscode-server/start.sh` - [VS Code Server](https://code.visualstudio.com/docs/remote/vscode-server) (official)
```
bamurtaugh, line 28:
```suggestion
- `tool-vscode-server/start.sh` - [VS Code Server](https://code.visualstudio.com/docs/remote/vscode-server) (official)
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+- `tool-openvscode-server/start.sh` - [openvscode-server](https://github.com/gitpod-io/openvscode-server)
+- `tool-vim-via-ssh/start.sh` - Vim via a SSH connection. SSH is used primarily to demonstrate how this could be achieved from other SSH supporting client tools.
```
bamurtaugh, line 30:
```suggestion
- `tool-vim-via-ssh/start.sh` - Vim via an SSH connection. SSH is used primarily to demonstrate how this could be achieved from other SSH supporting client tools.
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+When switching between examples, pass `true` in as an argument to get the container recreated to avoid port conflicts. e.g., `./start.sh true`
+
+In the first two examples, you'll be instructed to go to `http://localhost:8000` in a browser.
+
+This also adds a desktop to the container that can be accessed from a web browser at `http://localhost:6080` and you can connect using the password `vscode`.
+
+#### How the tool examples work
+
+This example demonstrates the use of the dev container CLI to:
```
bamurtaugh, line 40:
```suggestion
These examples demonstrate the use of the dev container CLI to:
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. Simplify setup using the "dev container features" concept. For example, SSH support is added just using a feature reference. See ``workspace/.devcontainer/devcontainer.jsonn` for more information.
```
bamurtaugh: Add a link to features info?
```suggestion
1. Simplify setup using the "[dev container features](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md)" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+2. How the dev container CLI can be used to inject tools without building them into the base image as follows:
```
bamurtaugh, line 44:
```suggestion
2. How the dev container CLI can be used to inject tools without building them into the base image:
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+    1. Use `devcontainer up` to spin up the container and mount a `server` and `workspace` folder into the container.
+    2. Use `devcontainer exec` to run a script from this mounted folder to set up the appropriate server.
+    3. In the `vim` example, a temporary SSH key is setup and configured, and then SSH is used from the command line to connect to the container once it is up and running. See `tool-vim-via-ssh/start.sh` for details.
```
bamurtaugh, line 48:
```suggestion
    3. In the `vim` example, a temporary SSH key is set up and configured, and then SSH is used from the command line to connect to the container once it is up and running. See `tool-vim-via-ssh/start.sh` for details.
```

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. Simplify setup using the "dev container features" concept. For example, SSH support is added just using a feature reference. See `workspace/.devcontainer/devcontainer.json` for more information.
+
+2. Execute an application build script inside a dev container as follows:
+
+    1. Use `devcontainer up` to spin up the container and mount the the `workspace` folder into the container.
+    2. Use `devcontainer exec` to run a build script from the mounted folder inside the development container.
+    3. Delete the container when the build is finished.
+
+All environment variables are automatically available from "exec", including those that are are set in the non-root user's ".bashrc" file. The devcontainer CLI also automatically adjusts to UID/GID differences for the user inside the container on Linux to ensure the workspace folder is writable.
```
bamurtaugh, line 76: Should this be formatted as code?
```suggestion
All environment variables are automatically available from `exec`, including those that are are set in the non-root user's `.bashrc` file. The dev container CLI also automatically adjusts to UID/GID differences for the user inside the container on Linux to ensure the workspace folder is writable.
```

Chuxel:
BTW - I raised several issues while putting this together, which seems to indicate it could be adapted to regression tests. I did not do that with this PR.

bamurtaugh (commented):
This is great to see, and I think it'll really help users! Left a few comments and suggestions.

example-usage/workspace/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+{
+	"name": "devcontainer CLI Demo",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+
+	// 👇 Dev Container Features - https://containers.dev/implementors/features/
+	"features": {
+		"ghcr.io/devcontainers/features/docker-in-docker": "latest",
```
jkeech, line 9: One thing @alexdima, @chrmarti, @edgonmsft, and I talked about a couple weeks ago was updating all of our examples and templates to pin features to the latest major version so that people who configure their repos based on these will pick up small updates automatically without picking up major breaking changes. Can we update these samples to use the `:1` version which is the current major version of all of the features?

A separate nit is whether we use the shorthand string value or the more explicit object syntax. I'd prefer us to phase out the shorthand form in our examples/templates and not treat the `version` option specially. I think that will help users understand that `version` is just one of the options supported by that feature, and there could be other options or no options depending on which feature they are using.

```suggestion
		"ghcr.io/devcontainers/features/docker-in-docker:1": {
		    "version": "latest"
		},
```

example-usage/ci-app-build-script/build-application.sh
```diff
@@ -0,0 +1,20 @@
...
+build_date="$(date +%s)"
+
+# Create a label for use during cleanup since the devcontainer CLI does 
+# not have a "remove" or "down" command yet (though this is planned).
+id_label="ci-container=${build_date}"
+
+# Run build
+devcontainer up --id-label ${id_label} --workspace-folder ../workspace
+set +e
+devcontainer exec --id-label ${id_label} --workspace-folder ../workspace scripts/execute-app-build.sh
```
jkeech, line 15: Do we need the `--id-label` argument here? I thought the CLI should be able to find the container based on the workspace folder. If so, it might be nice to remove the argument from this example to keep the execution simpler so folks don't think they have to pass the label when exec'ing.

jkeech (dismissed):
Thanks for making these examples! I think it will be really helpful for people to get started using the CLI. ✨ 

I left a couple small suggestions, but otherwise looks good to me. I agree that we could turn some scenarios like these into regression tests. That would be a great improvement for a future PR.

example-usage/workspace/.devcontainer/devcontainer.json
Chuxel: We should talk about the phasing out of the shorthand.  When you are hand editing, it's really handy.  Can do the other one though.

example-usage/ci-app-build-script/build-application.sh
Chuxel: @jkeech Yes.  As it explains above, there's no "down" yet. So there's no other way to identify the container and kill it yet.

jkeech: I'm not talking about the `docker rm` below -- I think it's fine to do the `up` with a specific label in order to delete as the third step. This comment here is for the middle `exec` step. Can't we still `exec` using the workspace folder to find the previous container?

example-usage/workspace/.devcontainer/devcontainer.json
jkeech: Yeah we aren't removing the shorthand. The main thing is that too much magic in the examples/docs doesn't help users understand what's going on, so being a bit more explicit/verbose in the examples can be helpful.

example-usage/ci-app-build-script/build-application.sh
Chuxel: @jkeech No because it identifies the container using the `id_label`.  If this was an "additional label" that would work, but that doesn't appear to be the case.  I initially didn't have it in exec, and kept getting errors that the container couldn't be found until I realized what was going on.

Might indicate we want a separate `--label` property independent of this one though since I agree that would be better.

jkeech: Got it. I thought we always added the workspace folder label. No problem if it's currently required. I think we should improve this in the CLI though which would make the sample cleaner (in addition to supporting the `down` command).

example-usage/workspace/.devcontainer/devcontainer.json
Chuxel: I'll switch the "latest" ones to `{}` and then add an option to the node one to help clarify. Otherwise just having "version" for each of them ends up looking pretty verbose. I can leave that for the "go" one for comparison though.

Chuxel: Done!

jkeech (dismissed):

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+    1. Use `devcontainer up` to spin up the container and mount a `server` and `workspace` folder into the container.
+    2. Use `devcontainer exec` to run a script from this mounted folder to set up the appropriate server.
+    3. In the `vim` example, a temporary SSH key is set up and configured, and then SSH is used from the command line to connect to the container once it is up and running. See `tool-vim-via-ssh/start.sh` for details.
+
+Currently the `appPort` property is used in `devcontainer.json` instead of `forwardPorts` due to a gap in the current dev container CLI ([see here](https://github.com/devcontainers/cli/issues/22)).
+
+### CI build environment example
+
+You can use the example by opening a terminal and typing the following:
```
chrmarti, line 54: Add a link to the CI action, so GitHub users are guided towards that.

chrmarti (approved):

example-usage/README.md
Chuxel: Updated!

example-usage/README.md
```diff
@@ -0,0 +1,92 @@
...
+1. Install Node.js 16 (e.g., using [nvm](https://github.com/nvm-sh/nvm))
+2. Install [node-gyp pre-requisites](https://github.com/nodejs/node-gyp):
+   - **Linux:** Use your distro's package manager. E.g. on Ubuntu/Debian: `sudo apt-get update && sudo apt-get install python3-minimal gcc g++ make`
+   - **macOS:** Install the XCode Command Line Tools ([more info](https://github.com/nodejs/node-gyp/blob/main/README.md#on-macos))
+3. Make sure you have an OpenSSH compliant `ssh` command available and in your path if you plan to use the `Vim via SSH` example (it should already be there on macOS, and in Linux/WSL, you can install `openssh-client` using your distro's package manager if its missing)
+3. Install the latest dev container CLI: `npm install -g @devcontainers/cli`
+
+## Using the examples
+
+All examples use the contents of the `workspace` folder for their configuration, which is where you can modifications if you'd like. The example scripts are then in different sub-folders.
```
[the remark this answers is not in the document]
Chuxel: Good catch. Yes. Updated.

jkeech (approved):

PR #110, opened 2022-08-07
Create home folder if it doesn't exist during updateUID

### LINKED ISSUE #109: updateUID fails if user home directory was not created
As [best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user), non-root users are created as system users without a home folder. `updateUID.Dockerfile` will fail when used with such images, because it tries to chown the non-existing home folder.

We could either skip chowning or create the home folder. The latter would be useful since vscode seems to use the remote user home folder for its own setup.

barraponto:
Fixes #109.

We could either skip chowning or create the home folder.
I find the latter to be more useful because otherwise vscode will complain it can't create stuff in the non-existing home folder.

### DIFF
--- a/scripts/updateUID.Dockerfile
+++ b/scripts/updateUID.Dockerfile
@@ -26,6 +26,9 @@ RUN eval $(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/
 		if [ "$OLD_GID" != "$NEW_GID" ]; then \
 			sed -i -e "s/\([^:]*:[^:]*:\)${OLD_GID}:/\1${NEW_GID}:/" /etc/group; \
 		fi; \
+		if [ ! -d "$HOME_FOLDER" ]; then \
+			mkhomedir_helper ${REMOTE_USER}; \
+		fi; \
 		chown -R $NEW_UID:$NEW_GID $HOME_FOLDER; \
 	fi;
 


jkeech (approved):
Thanks for making this change! It looks like a nice improvement.

I don't think it will work for alpine-based devcontainer images as-is because `mkhomedir_helper` is not installed there. Perhaps we could do something like `mkhomedir_helper ${REMOTE_USER} || mkdir -p $HOME_FOLDER` as a fallback.

@chrmarti any concerns? Do we have tests for the UID updating logic? I couldn't find any with a quick scan. If we have tests for this, it would be great to update to include this scenario.

scripts/updateUID.Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN eval $(sed -n "s/${REMOTE_USER}:[^:]*:\([^:]*\):\([^:]*\):[^:]*:\([^:]*\).*/
 		if [ "$OLD_GID" != "$NEW_GID" ]; then \
 			sed -i -e "s/\([^:]*:[^:]*:\)${OLD_GID}:/\1${NEW_GID}:/" /etc/group; \
 		fi; \
+		if [ ! -d "$HOME_FOLDER" ]; then \
+			mkhomedir_helper ${REMOTE_USER}; \
```
chrmarti, line 30: This copies `/etc/skel` over only on Linux (the only platform where the UID/GID update runs), but not on Windows or Mac. It is better to keep the platforms aligned. Let's skip the `chown` below when the folder does not exist.

chrmarti (changes requested):
Thanks for the PR! I have left a comment for discussion.

scripts/updateUID.Dockerfile
barraponto: oh, ok. I'll send an updated version.

Unrelated but... I love how this dockerfile gets picked up and builds another image (with another name) on top of the user's one. How can I add my own dockerfiles?

I'm changing both the dockerfile and the image name in my own `.devcontainer/docker-compose.yaml` so i can base my `.devcontainer/Dockerfile` on top of my project's own image. It works, but I have to build the regular project image beforehand...

chrmarti:
@jkeech (not having a reply field on your comment): There is no test for this yet. Would be good to have.

Note that we don't run the UID/GID update in Codespaces currently.

scripts/updateUID.Dockerfile
chrmarti: Great, thanks!

We currently don't support adding more Dockerfiles beyond this one which we use internally. I suggest you open a feature request. Thanks.

Toppole6 (approved):

chrmarti:
Closing due to inactivity. Thanks.

PR #112, opened 2022-08-08
Bug fix: OCI - Reference features in overrideFeatureInstallOrder with id

samruddhikhandale:
The CLI should differentiate features for `overrideFeatureInstallOrder` with feature `id`. However, it fails to resolve for OCI features. Fixing that! 

Error thrown by current CLI:
image

```
"features": {
    "ghcr.io/devcontainers/features:common-utils:1",
    "ghcr.io/devcontainers/features:node:1",
},
"overrideFeatureInstallOrder": [
    "ghcr.io/devcontainers/features:common-utils"
]
```


example-usage/workspace/.devcontainer/devcontainer.json
```diff
@@ -18,6 +18,10 @@
 		// Optional - For tools that require SSH
 		"ghcr.io/devcontainers/features/sshd:1": { }
 	},
+	"overrideFeatureInstallOrder": [
```
jkeech, line 21: Does order actually matter for these examples? I don't think these examples are meant to be a comprehensive view of all possible devcontainer options. If order doesn't matter, I would recommend removing this from the example to keep it simpler. `overrideFeatureInstallOrder` is not a property that most people should ever need to use.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -31,7 +31,7 @@ export function computeOverrideInstallationOrder(config: DevContainerConfig, fea
     // Moves to the beginning the features that are explicitly configured.
     const orderedFeatures = [];
     for (const featureId of config.overrideFeatureInstallOrder!) {
-        const feature = automaticOrder.find(feature => feature.features[0].name === featureId);
+        const feature = automaticOrder.find(feature => feature.features[0].name === featureId || (feature.sourceInformation.type === 'oci' && feature.sourceInformation.featureRef.resource === featureId));
```
jkeech, line 34: Would this have to possibility to find the wrong feature? Suppose I have `ghcr.io/devcontainers/features/node` and `ghcr.io/myorg/myrepo/node`. If it's searching by `name` alone (which is `node` in this case?), then it's ambiguous which one it will find.

I think we always want to go by the fully-qualified id and not rely on shorter name matching.

jkeech: You could also add a test for this case with two features uploaded to different registries with the same name/suffix and asserting that you can pin the order correctly for either one.

example-usage/workspace/.devcontainer/devcontainer.json
samruddhikhandale: I just wanted to add a test coverage, so thought of adding to current devcontainer.json. I'll create a separate one for `overrideFeatureInstallOrder` test. 👍

jkeech: These `example-usage` ones are not currently covered by tests, but are setup as an example for users to see how to use the CLI in different scenarios. They were recently added in https://github.com/devcontainers/cli/pull/107.

src/spec-configuration/containerFeaturesOrder.ts
samruddhikhandale: For OCI, the feature would not be resolved with `name`, but instead it will use the newly added condition which checks for `resource`.  (feature.features[0].name="node" and featureId="ghcr.io/devcontainers/features/node or ghcr.io/myorg/myrepo/node"  )

`name` is useful only when referencing a shorthand syntax or local-feature.

image

jkeech: @samruddhikhandale when using OCI-published features, is the `name` undefined at runtime or would it be `node` in the example above? If it's `node`, then having the `name` check as the first condition will cause the shorter one to short-circuit the fully-qualified name check. That's why I think we'd always want to prefer the fully qualified name with higher precedence, or better yet, only ever match on the fully-qualified name.

samruddhikhandale: Ah, sorry! I misunderstood your question. I have updated the condition and added tests, for OCI features it will never resolve with `name` but use `resource` instead. 

image

example-usage/workspace/.devcontainer/devcontainer.json
joshspicer: I have added a lot of tests recently under [`./src/test/container-features`](https://github.com/devcontainers/cli/tree/main/src/test/container-features), including example projects that the tests will execute.  I suggest adding a test there instead

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: I think we want to standardize on using `feature[0].id` here, which should always match the `id` of the individual feature's `devcontainer-feature.json`.  The issues I discovered recently was due to that not being consistent in the new code.  For information on _where_ the feature came from, we have the `sourceInformation` object attached to each feature.

I believe we have no standardized pattern for what 'name' should look like.  I have always thought of it as the human-readable name and am surprised we're using it here

joshspicer: I am working on additional changes that will always assert that the `feature[0].id` has no special characters https://github.com/devcontainers/cli/blob/jospicer/feature-test-invariants/src/test/container-features/faetureHelpers.test.ts#L39-L44

chrmarti: > I think we want to standardize on using `feature[0].id` here, which should always match the `id` of the individual feature's `devcontainer-feature.json`.

@joshspicer Isn't "id" in the devcontainer-feature.json a short name that is ambiguous?

joshspicer: Perhaps this need clarification in the spec, but the way `devcontainers/features` is going to be structured, each folder name, `devcontainer-feature.json` id, and the identifier that would go in a user's devcontainer should all be the same and unique within the collection.  We will parse out `sourceInformation` based on how the user references the feature in their devcontainer.json

So for example,

`ghcr.io/devcontainer/features/ruby:1`

the `id` of this feature MUST be `ruby`, and its source code MUST be enclosed in a folder named `ruby`.

(https://github.com/devcontainers/features/tree/main/src/ruby)

samruddhikhandale: > I think we want to standardize on using feature[0].id here, which should always match the id of the individual feature's devcontainer-feature.json

@joshspicer According to the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature), the `overrideFeatureInstallOrder` expects list of feature IDs. In case of a local feature (aka file-path), it expects `<path-to-feature-dir>` which can only be resolved by matching the `feature[0].name`
  
image

Similarly, for tgz, the automatic order can be resolved only on `feature[0].name` - [Reference](https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L571)

samruddhikhandale: Looking at [processFeatureIdentifier()](https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L520), the `feature.name` is not necessarily the `name` param from `devcontainer.json` 🤔

joshspicer: My view is that:

- `name` should never be used for anything other than UX/human-readable purposes.  Any current uses of it in the CLI for processing is invalid and should be fixed.
- `id` is the unique name of a feature in a given collection. _It contains NO information about _where_ this feature came from_.
- Each feature object also has a `sourceInformation` object attached to it, which is the result of parsing the user's devcontainer to separate _where_ the feature came from, and the feature itself.

joshspicer: A `file-path` feature has a `resolveFilePath` attribute that points to the feature's location on disk.  https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L93-L96

Be sure not to get this with the antiquate 'local-cache' `SourceInformation` type.  Those are the packaged features that we ship directly with the CLI (https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L78-L80)

joshspicer: > Isn't "id" in the devcontainer-feature.json a short name that is ambiguous?

@chrmarti It _would_ be ambiguous, but every Feature has a `SourceInformation` object attached to it that indicates where it was downloaded/sourced from.  

@edgonmsft 's solution to ensuring the node layer can handle >1 feature unambiguously is to name each temporary build folder uniquely
https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L715-L722

samruddhikhandale: > A file-path feature has a resolveFilePath attribute that points to the feature's location on disk.

For local feature scenario, the `resolveFilePath`  is not necessarily same as the id provided in `overrideFeatureInstallOrder`

image

We can definitely not use `id` because how can we differentiate `./local-feature/node` with `/local/node`

samruddhikhandale: Discussed offline with @joshspicer , I'll add another property `userFeatureId` in `SourceInformation` which will provide us with the `id` expected by https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature. This prop will be used in the above condition (no need of `feature[0].name`).

I'll create a followup PR to modify the `processFeatureIdentifier` function which should take in a DevContainerFeature (the raw JSON from the user's devcontainer), and return a feature set with all the relevant information parsed out.

chrmarti (dismissed):
LGTM!

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -73,26 +73,31 @@ export type SourceInformation = LocalCacheSourceInformation | GithubSourceInform
 
 interface BaseSourceInformation {
 	type: string;
+	referenceId: string; // Dictates how a supporting tool will locate and download a given feature. See https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature
```
joshspicer, line 76: In other places in the codebase we've used `userFeature` (which has an `id` and `options`) to represent the entire object in the `devcontainer.json`'s `features` map.

I'm thinking here we want to try and align the names as much as possible, either by changing what we already have, or choosing a name here closer to other, existing uses

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -52,7 +52,7 @@ export function getOCIFeatureSet(output: Log, identifier: string, options: boole
             type: 'oci',
             manifest: manifest,
             featureRef: featureRef,
-
+            referenceId: `${featureRef.resource}:${featureRef.version}`
```
joshspicer, line 55: I believe `identifier` here is exactly the value you're looking for.

This is the only current use of that function 
```
let newFeaturesSet: FeatureSet = getOCIFeatureSet(output, userFeature.id, userFeature.options, manifest);

```

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -31,7 +31,7 @@ export function computeOverrideInstallationOrder(config: DevContainerConfig, fea
     // Moves to the beginning the features that are explicitly configured.
     const orderedFeatures = [];
     for (const featureId of config.overrideFeatureInstallOrder!) {
-        const feature = automaticOrder.find(feature => feature.features[0].name === featureId);
+        const feature = automaticOrder.find(feature => (feature.sourceInformation.type === 'oci' && feature.sourceInformation.referenceId.split(':')[0] === featureId) || feature.sourceInformation.referenceId === featureId);
```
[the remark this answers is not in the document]
joshspicer: This might mean keeping track of two new variables in the `sourceInformation`.  That is nice since we can write static tests against the generated `featureConfig` before we actually use it anywhere (like here).

src/test/container-features/containerFeaturesOrder.test.ts
```diff
@@ -109,4 +173,40 @@ describe('Container features install order', () => {
...
+        const spiltOnSlash = splitOnCollon[0].split('/');
+        return {
+            sourceInformation: {
+                type: 'oci',
+                featureRef: {
+                    id: spiltOnSlash[3],
+                    namespace: `${spiltOnSlash[1]}/${spiltOnSlash[2]}`,
+                    owner: spiltOnSlash[1],
+                    registry: spiltOnSlash[0],
+                    resource: splitOnCollon[0],
```
joshspicer, line 189:
```suggestion
                    resource: splitOnColon[0],
```

src/test/container-features/containerFeaturesOrder.test.ts
```diff
@@ -109,4 +173,40 @@ describe('Container features install order', () => {
...
+                    version: splitOnCollon[1]
```
joshspicer, line 190:
```suggestion
                    version: splitOnColon[1]
```

src/test/container-features/containerFeaturesOrder.test.ts
```diff
@@ -109,4 +173,40 @@ describe('Container features install order', () => {
             }],
         };
     }
-});
\ No newline at end of file
+
+    function getOCIFeatureSet(id: string): FeatureSet {
+        // example - ghcr.io/devcontainers/features/node:1
+        const splitOnCollon = id.split(':');
```
joshspicer, line 179:
```suggestion
        const splitOnColon = id.split(':');
```

joshspicer (changes requested):
Some minor requests regarding naming conventions and assumptions we make after parsing. Otherwise, this is a great change for separating responsibilities between different parts of the codebase 🚀

src/spec-configuration/containerFeaturesOCI.ts
samruddhikhandale: Ah! right 👍

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Renamed to `userFeatureId`

src/spec-configuration/containerFeaturesOrder.ts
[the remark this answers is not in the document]
samruddhikhandale: Thanks for catching that. Modified condition and added test.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -31,7 +31,7 @@ export function computeOverrideInstallationOrder(config: DevContainerConfig, fea
     // Moves to the beginning the features that are explicitly configured.
     const orderedFeatures = [];
     for (const featureId of config.overrideFeatureInstallOrder!) {
-        const feature = automaticOrder.find(feature => feature.features[0].name === featureId);
+        const feature = automaticOrder.find(feature => (feature.sourceInformation.type === 'oci' && feature.sourceInformation.userFeatureId.split(':')[0] === featureId) || (feature.sourceInformation.type === 'github-repo' && feature.sourceInformation.userFeatureId.split('@')[0] === featureId) || feature.sourceInformation.userFeatureId === featureId);
```
joshspicer, line 34: Remind me why this needs to change, and why we're filtering for only OCI features?

src/spec-configuration/containerFeaturesOCI.ts
```diff
@@ -13,103 +13,102 @@ export const DEVCONTAINER_COLLECTION_LAYER_MEDIATYPE = 'application/vnd.devconta
...
+    const feat = {
+        id: featureRef.id,
+        name: featureRef.id,
+        included: true,
+        value: options
+    };
+
+    let featureSet: FeatureSet = {
+        sourceInformation: {
+            type: 'oci',
```
joshspicer, line 66: Can all the whitespace changes be reverted?

src/spec-configuration/containerFeaturesOrder.ts
samruddhikhandale: We are changing this to match with the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#1-overridefeatureinstallorder). (ref issue - https://github.com/devcontainers/images/pull/53#discussion_r939429374 )

> why we're filtering for only OCI features
 We are filtering for all the types, `userFeatureId` follows this [part of spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature), however for `OCI` the `overridefeatureinstallorder` needs unversioned feature IDs, hence the special condition.

joshspicer: I would prefer breaking up this long condition into a couple local variables for readability.  

Better yet, break up the individual components (location bit and version bit) and store them on the SourceInformation as well. That will make it easier to unit test.

src/spec-configuration/containerFeaturesOCI.ts
joshspicer: Note that i'm trying to make all new files I had use tabs instead of spaces for whitespace, to conform to how vscode/the rest of the codebase is set up.

joshspicer (dismissed):
lgtm, left a couple comments

src/spec-configuration/containerFeaturesOCI.ts
samruddhikhandale: Reverted

src/spec-configuration/containerFeaturesOrder.ts
samruddhikhandale: > Better yet, break up the individual components (location bit and version bit) and store them on the SourceInformation as well.

Done!

joshspicer (approved):
thanks!

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -87,19 +88,19 @@ export function computeInstallationOrder(features: FeatureSet[]) {
         }
         orderedFeatures.push(
             ...current.map(node => node.feature)
-                .sort((a, b) => a.features[0].name.localeCompare(b.features[0].name)) // stable order
+                .sort((a, b) => a.sourceInformation.userFeatureId.localeCompare(b.sourceInformation.userFeatureId)) // stable order
```
alexdima, line 91: This was there before, but I think using `localeCompare` is not a very good idea, as it might be influenced by the locale of the current machine running the code. So `u` and `ü` might be sorted differently on different machines with different locales.

alexdima (dismissed):
Looks good! I left a couple of small comments.

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: Something to note is that this will also likely change the execution order for users, which may have unintended effects.  This auto ordering is by-design though, and unavoidable right now

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -73,26 +73,33 @@ export type SourceInformation = LocalCacheSourceInformation | GithubSourceInform
 
 interface BaseSourceInformation {
 	type: string;
+	userFeatureId: string; // Dictates how a supporting tool will locate and download a given feature. See https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#referencing-a-feature
+	userFeatureIdWithoutVersion?: string;
 }
 
 export interface LocalCacheSourceInformation extends BaseSourceInformation {
 	type: 'local-cache';
+	userFeatureId: string;
```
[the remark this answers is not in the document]
samruddhikhandale: Good catch. Modified .

src/spec-configuration/containerFeaturesOrder.ts
samruddhikhandale: @alexdima Switched to 👇 , should that help address the issue? 

> .sort((a, b) => a.sourceInformation.userFeatureId < b.sourceInformation.userFeatureId ? -1 : 1)

alexdima: :+1: Yes, switching to use `<` guarantees a stable order across different machines, as it is not locale dependent. See https://github.com/microsoft/vscode/issues/70920#issuecomment-546405072 for more context.

alexdima (approved):

Issue #120, opened 2022-08-12
Target not overwritten when using features


chrmarti:
Reported by @sajaysurya in https://github.com/microsoft/vscode-dev-containers/issues/1519:

Currently the docker-from-docker feature doesn't support multistage Dockerfiles

Relates to: Remote - Containers

It works well with the following Dockerfile
```
FROM python:3.10.5 
RUN apt-get update -q \
    && apt-get install -q -y --no-install-recommends \
        sudo
# add non-root user with sudo access (sudo required by docker-from-docker feature)
RUN adduser vscode
RUN echo vscode ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/vscode \
    && chmod 0440 /etc/sudoers.d/$USERNAME
USER vscode
```
and the following .devcontainer.json file
```
        "dockerFile": "Dockerfile",
        "containerUser": "vscode",
	"features": {
		"docker-from-docker": {
			"version": "latest"
		}
	}
```

It actually works with a named stage in the Dockerfile like
```
FROM python:3.10.5 AS vscode_dev
RUN apt-get update -q \
    && apt-get install -q -y --no-install-recommends \
        sudo
# add non-root user with sudo access (sudo required by docker-from-docker feature)
RUN adduser vscode
RUN echo vscode ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/vscode \
    && chmod 0440 /etc/sudoers.d/$USERNAME
USER vscode
```
as long as the build target in not explicitly mentioned in the .devcontainer.json file
```
        "dockerFile": "Dockerfile",
	"build": { "target": "vscode_dev" }, <--- NOTE: works without this line!
        "containerUser": "vscode",
	"features": {
		"docker-from-docker": {
			"version": "latest"
		}
	}
```

chrmarti:
PR #121 up for review.

chrmarti:
Fix is available in Dev Containers CLI 0.10.1, 0.12.0 and Remote-Containers 0.245.1 and 0.248.0-pre-release.

rzhao271:
I'm unable to repro the original issue on Windows with older Remote-Containers versions. What are some verification steps for this issue?

chrmarti:
@rzhao271 Thanks for the ping, the following show repro and verify it:

With the following with 0.247.0-pre-release there is no `gh` command in the dev container, with the fix in later versions `gh` is available:

devcontainer.json:
```json
{
	"build": {
[...]

PR #121, opened 2022-08-12
Override target when using features (#120)

chrmarti:


joshspicer (dismissed):
Could the example from https://github.com/devcontainers/cli/issues/120#issue-1336964922 be captured as a test?

chrmarti:
@joshspicer Added a check to an existing test. Could you reapprove? (Also: There already was a check for the Docker Compose case which does not pass with my Docker Compose version 2.7.0 locally. I'm adding a log entry with the Compose version to get more insight.)

joshspicer (approved):

PR #124, opened 2022-08-12
run scenarios.json nested within a ./test/<feature> folder

joshspicer:
This change makes it easier on a per-feature level to assert a set of "scenarios", which are full `devcontainer.json` configs with an associated test script.

The main advantage of this is that in CI, we can't _just_ run scenarios where the code has been touched.  The [current, single scenario test takes 40 minutes to run right now](https://github.com/devcontainers/features/runs/7809178393?check_suite_focus=true).  This change will let us scope scenarios to individual features when it makes sense.

I've also created an optional, special case folder under `tests/_global`, that will have its own set of scenarios to execute.  The idea here is that there may be tests that don't neatly fit into a given feature (a test spanning several features).  this will replace the `test-scenarios` folder.

## Example 

Each feature (eg: `dotnet`) can provide a list of possible scenarios scoped to the feature itself.  This is useful for testing different options for a single feature.

```
@joshspicer ➜ ~/features/test/dotnet (main ✗) $ tree
.
├── install_additional_dotnet.sh
├── scenarios.json
└── test.sh
```

The `scenarios.json` file in this case would be an object of an arbitrary number of labelled dev container configs, representing test cases.

```json
{
"install_additional_dotnet": {
      "image": "ubuntu:focal",
      "features": {
      "dotnet": {
                  "version": "6.0.301",
                  "additionalVersions": "5.0,3.1.420"
      }
   },
   "..." : {}
}
```

The `features test` command will create a container for **each** scenario based on the declared dev container config, and then assert that an accompanying shell script (named the same as the label (in this case, `install_additional_dotnet.sh`), exits with a zero exit code.

Additionally, a test case will be always be autogenerated with no options (an empty object, `{}`) passed.

At the end, a summary of the results of both the auto-generated test, and each scenario ran, will be printed at the bottom of the test.

image

If any test fails, the CLI will exit with a non-zero exit code, otherwise it will exit with a 0 exit code.


jungaretti:
This change would make it much easier and faster to test a feature's options (version, add-ons, etc.)

src/spec-node/featuresCLI/test.ts
```diff
@@ -11,17 +11,22 @@ import { doFeaturesTestCommand } from './testCommandImpl';
...
-			'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to auto-detect all features in collection directory.  Cannot be combined with \'-s\'.', },
-			'scenarios': { type: 'string', alias: 's', description: 'Path to scenario test directory (containing scenarios.json).  Cannot be combined with \'-f\'.' },
-			'remote-user': { type: 'string', alias: 'u', default: 'root', describe: 'Remote user', },
-			'collection-folder': { type: 'string', alias: 'c', default: '.', description: 'Path to folder containing \'src\' and \'test\' sub-folders.' },
+			'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to run all tests.  Cannot be combined with \'--global-scenarios-only\'.', },
+			'global-scenarios-only': { type: 'boolean', default: false, description: 'Run only scenario tests under \'tests/_global\' .  Cannot be combined with \'-f\'.' },
+			'skip-scenarios': { type: 'boolean', default: false, description: 'Skip all \'scenario\' style tests under  Cannot be combined with \'-global--scenarios-only\'.' },
```
chrmarti, line 16: Description seems to be missing a word or two after 'under'? '-global--scenarios-only' has a misplaced dash.

src/spec-node/featuresCLI/test.ts
```diff
@@ -11,17 +11,22 @@ import { doFeaturesTestCommand } from './testCommandImpl';
 export function featuresTestOptions(y: Argv) {
 	return y
 		.options({
-			'base-image': { type: 'string', alias: 'i', default: 'ubuntu:focal', description: 'Base Image' },
-			'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to auto-detect all features in collection directory.  Cannot be combined with \'-s\'.', },
-			'scenarios': { type: 'string', alias: 's', description: 'Path to scenario test directory (containing scenarios.json).  Cannot be combined with \'-f\'.' },
-			'remote-user': { type: 'string', alias: 'u', default: 'root', describe: 'Remote user', },
-			'collection-folder': { type: 'string', alias: 'c', default: '.', description: 'Path to folder containing \'src\' and \'test\' sub-folders.' },
+			'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to run all tests.  Cannot be combined with \'--global-scenarios-only\'.', },
```
chrmarti, line 14: Can you add `conflicts: ['global-scenarios-only']`? (Haven't tried that yet.)

chrmarti (changes requested):

joshspicer:
i'm seeing the `conflicts` statement blocking when it shouldn't, probably since `--global-scenarios-only` has a default value of false.
image

I think we're covered by the `.check(...)` statements on lines 24-31, so I'm going to leave out conflicts (unless there's another reason to add that over the checks I previously had)

chrmarti (approved):
LGTM!

joshspicer:
Validated by manually running build against features repo (using updated commands as seen in https://github.com/devcontainers/features/pull/114)

Issue #125, opened 2022-08-12
"customizations" (and legacy extension, settings properties) not processed in features


Chuxel:
Running `devcontainer read-configuration --include-features-configuration` on a workspace with a devcontainer.json with features does not return `customizations` or the legacy `extensions` or `settings` properties (whether original or converted).

In the latest Remote - Containers release extensions are not installed and settings are not applied either. 

They seem to be missing in several spots along with other properties from the spec (https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#devcontainer-featurejson-properties).

Possible places where the issue could be coming from (if not all of them):

- https://github.com/devcontainers/cli/blob/0147a3685f281f0ec743cf1e5e245a1023f09fff/src/spec-configuration/containerFeaturesConfiguration.ts#L931-L940
- https://github.com/devcontainers/cli/blob/0147a3685f281f0ec743cf1e5e245a1023f09fff/src/spec-configuration/containerFeaturesConfiguration.ts#L964-L971
- https://github.com/devcontainers/cli/blob/c97190e95f44e344fbd02789ab43a48f62aef7b0/src/spec-configuration/containerFeaturesConfiguration.ts#L24-L48

Chuxel:
//cc @chrmarti @alexdima @bamurtaugh given the Remote - Containers impact.

PR #126, opened 2022-08-13
(IMP) bug fix: missing devcontainer-feature properties are now processed in features

samruddhikhandale:
Fixes: https://github.com/devcontainers/cli/issues/125

All props from https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#devcontainer-featurejson-properties are now included in processed features. 

image

Added tests

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -7,7 +7,7 @@ import * as jsonc from 'jsonc-parser';
 import * as path from 'path';
 import * as URL from 'url';
 import * as tar from 'tar';
-import { DevContainerConfig, DevContainerFeature } from './configuration';
+import { DevContainerConfig, DevContainerFeature, VSCodeCustomizations } from './configuration';
 import { mkdirpLocal, readLocalFile, rmLocal, writeLocalFile, cpDirectoryLocal, isLocalFile, isLocalFolder } from '../spec-utils/pfs';
 import { Log, LogLevel } from '../spec-utils/log';
 import { request } from '../spec-utils/httpRequest';
@@ -45,6 +45,8 @@ export interface Feature {
 	exclude?: string[];
 	value: boolean | string | Record<string, boolean | string | undefined>; // set programmatically
 	included: boolean; // set programmatically
+	customizations?: VSCodeCustomizations;
+	installsAfter?: string[];
 }
 
 export type FeatureOption = {
@@ -230,19 +232,19 @@ export function getFeatureLayers(featuresConfig: FeaturesConfig) {
 && ./install.sh
 
 `;
-  });
+	});
 	// Features version 2
 	featuresConfig.featureSets.filter(y => y.internalVersion === '2').forEach(featureSet => {
 		featureSet.features.forEach(feature => {
-				result += generateContainerEnvs(feature);
-				result += `
+			result += generateContainerEnvs(feature);
+			result += `
 				
 RUN cd /tmp/build-features/${feature.consecutiveId} \\
 && export $(cat devcontainer-features.env | xargs) \\
 && chmod +x ./install.sh \\
 && ./install.sh
 
-`;	
+`;
 		});
 	});
 	return result;
@@ -251,8 +253,7 @@ RUN cd /tmp/build-features/${feature.consecutiveId} \\
 // Features version two export their environment variables as part of the Dockerfile to make them available to subsequent features.
 export function generateContainerEnvs(feature: Feature) {
 	let result = '';
-	if(!feature.containerEnv)
-	{
+	if (!feature.containerEnv) {
 		return result;
 	}
 	let keys = Object.keys(feature.containerEnv);
@@ -351,7 +352,7 @@ export async function loadV1FeaturesJsonFromDisk(pathToDirectory: string, output
 	return loadFeaturesJson(jsonBuffer, filePath, output);
 }
 
-function updateFromOldProperties<T extends { features: (Feature & { extensions?: string[]; settings?: object; customizations?: { vscode?: { extensions?: string[]; settings?: object } } })[] }>(original: T): T {
+function updateFromOldProperties<T extends { features: (Feature & { extensions?: string[]; settings?: object; customizations?: VSCodeCustomizations })[] }>(original: T): T {
 	// https://github.com/microsoft/dev-container-spec/issues/1
 	if (!original.features.find(f => f.extensions || f.settings)) {
 		return original;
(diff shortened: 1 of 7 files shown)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -839,10 +847,13 @@ async function fetchFeatures(params: { extensionPath: string; cwd: string; outpu
 
 				if (didSucceed) {
 					output.write(`Succeeded fetching ${tarballUri}`, LogLevel.Trace);
-					if (!(await applyFeatureConfigToFeature(output, featureSet, feature, featCachePath))) {
+					let updatedFeature = await applyFeatureConfigToFeature(output, featureSet, feature, featCachePath);
```
joshspicer, line 850: No need to break this out, we still return `Feature | undefined`

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -769,11 +769,13 @@ async function fetchFeatures(params: { extensionPath: string; cwd: string; outpu
...
 				}
 
-				if (!(await applyFeatureConfigToFeature(output, featureSet, feature, featCachePath))) {
+				let updatedFeature = await applyFeatureConfigToFeature(output, featureSet, feature, featCachePath);
+				if (!updatedFeature) {
 					const err = `Failed to parse feature '${featureDebugId}'. Please check your devcontainer.json 'features' attribute.`;
 					throw new Error(err);
 				}
 
+				featuresConfig.featureSets[i].features[0] = updatedFeature;
```
joshspicer, line 778: Unless you noticed otherwise, it should be fine to just set this value in `appleFeatureConfigToFeature`. The featureConfig is passed as a ["copy of a reference"](https://stackoverflow.com/questions/13104494/does-javascript-pass-by-reference)

joshspicer: that would prevent the need to switch to a counted for loop as well

Chuxel:
Just to double check - I assume this will also cover v1 features with `extensions` and `settings` at the top level for a feature?

samruddhikhandale:
> Just to double check - I assume this will also cover v1 features with `extensions` and `settings` at the top level for a feature?

@Chuxel yes, tested it locally, they are covered. 👍

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Ah! I did not realize that featureConfig is passed as a ["copy of a reference"](https://stackoverflow.com/questions/13104494/does-javascript-pass-by-reference). 

Modified my changes

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -926,19 +925,19 @@ async function applyFeatureConfigToFeature(output: Log, featureSet: FeatureSet,
...
+	feature = {
+		...featureJson,
+		...feature
+	};
+
+	let updatedFeature = updateFromOldProperties({ features: [feature] }).features[0];
 
+	feature = {
+		...updatedFeature
+	};
```
joshspicer, line 938: I don't think this is any need to re-assign `feature` here AND manually assign it to `featureSet.feature[0]` below

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -926,19 +925,19 @@ async function applyFeatureConfigToFeature(output: Log, featureSet: FeatureSet,
 	const jsonString: Buffer = await readLocalFile(innerJsonPath);
 	const featureJson = jsonc.parse(jsonString.toString());
 
-	// TODO: Use spread operator {..., } to avoid needing to explicitly set each property.
 
-	feature.containerEnv = featureJson.containerEnv;
-	feature.buildArg = featureJson.buildArg;
-	feature.options = featureJson.options;
```
joshspicer, line 933: I feel like we should never been assigning anything to `featureJson.options` here, as that is kind of a special case in the `devcontainer-feature.json` as UI metadata (https://github.com/devcontainers/features/blob/main/src/docker-from-docker/devcontainer-feature.json#L6-L31).

I think this is just ignored later on, and the options are passed along separately.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -961,15 +960,18 @@ async function parseDevContainerFeature_v1Impl(output: Log, featureSet: FeatureS
...
+	feature = {
+		...seekedFeature,
+		...feature
+	};
+
+	let updatedFeature = updateFromOldProperties({ features: [feature] }).features[0];
+
+	feature = {
+		...updatedFeature
+	};
```
joshspicer, line 972: same as above

joshspicer (approved):
Would be great to also get @chrmarti & @edgonmsft to take a peek, but i think this is 👍

src/test/container-features/e2e.test.ts
```diff
@@ -68,6 +72,16 @@ describe('Dev Container Features E2E (remote)', function () {
...
                 assert.match(res.stderr, /CONTAINER ID/);
             });
+
+            it('should read configuration with features', async () => {
+                const res = await shellExec(`${cli} read-configuration --workspace-folder ${testFolder} --include-features-configuration  --skip-feature-auto-mapping`);
+                const response = JSON.parse(res.stdout);
+                console.log(res.stderr);
+                assert.equal(response?.featuresConfiguration?.featureSets[0]?.features[0]?.id, 'docker-in-docker');
+                assert.equal(response?.featuresConfiguration?.featureSets[0]?.features[0]?.customizations?.vscode?.extensions[0], 'ms-azuretools.vscode-docker');
+                assert.equal(response?.featuresConfiguration?.featureSets[2]?.features[0]?.id, 'node');
```
joshspicer, line 82: Note that i do not think the order of this is stable, and might cause this test to incorrectly fail in the future

joshspicer: Perhaps `response?.featuresConfiguration?.find(f => ...)` for each feature, and also adding an assert that the total size of `featureSets` is 3?

joshspicer:
I think it would be worth adding tests here as well https://github.com/devcontainers/cli/blob/main/src/test/container-features/generateFeaturesConfig.test.ts

src/test/container-features/e2e.test.ts
samruddhikhandale: Modified 👍

joshspicer (dismissed):

joshspicer:
One more quick note - with https://github.com/devcontainers/features/pull/97 we will lose the regression testing for top-level `extensions`/`settings`, etc.. 

Would be cool to add one more test (maybe with local features) to test that top level vscode attributes continue to work indefinitely.

alexdima (approved):
Looks good!

PR #128, opened 2022-08-17
Improve CLI examples to illustrate application of tool specific settings

Chuxel:
In this case, VS Code extensions and settings.


samruddhikhandale (approved):
👏 Thank you! LGTM

Issue #131, opened 2022-08-19
"ENV" randomly added to environment variables that originate from features can prevent container from starting


Chuxel:
Currently if you add both the node and python features, the container will fail to start properly in Remote - Containers stable.
 

In CLI 0.10.1, try building using this devcontainer.json...

```json
{
    "image": "mcr.microsoft.com/vscode/devcontainers/base:0-bullseye",
    "features": {
       "node": "latest",
       "python": "os-provided"
    } 
}
```

Now, start the container using the image that was built but specify `/bin/bash` as the entrypoint:

```
docker run -it --rm <image-id> /bin/bash -c 'echo ${PATH}'
```

You will get...

```
/usr/local/python/bin:/usr/local/share/nvm/current/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/binENV:/usr/local/py-utils/bin:/root/.local/bin
```

The `ENV` is smashed up against `/bin` (becoming `/binENV`) which prevents anything under the `/bin` folder, including bash, from being found.

With 0.11.1 or using the new dev container features via OCI, you will instead see a hang after the installation of Python followed  eventually by lots of errors about `chown`... possibly caused by the same type of thing.

```
{
    "image": "mcr.microsoft.com/vscode/devcontainers/base:0-bullseye",
    "features": {
       "ghcr.io/devcontainers/features/node": "latest",
       "ghcr.io/devcontainers/features/python": "os-provided"
    } 
}
```

Build result:
```
...
#0 46.35 Updating /etc/bash.bashrc and /etc/zsh/zshrc...
#0 141.7 chmod: cannot access '/usr/local/share/nvm/test/fast/Listing': No such
file or directory
#0 141.7 chmod: cannot access 'versions': No such file or directory
#0 141.7 chmod: cannot access '/usr/local/share/nvm/test/fast/Listing': No such
file or directory
...
```

joshspicer:
Re-opening to track releasing updating OSS CLI, as well as a Remote-Containers fix

chrmarti:
Fix is available in Dev Containers CLI 0.10.2, 0.12.1 and Remote-Containers 0.245.1 and 0.248.0-pre-release. @joshspicer Are you tracking the rollout for Codespaces here or should we close this issue?

Chuxel:
Also - the v2 hang in this scenario is still there, correct? Do we want to fork this issue for that part?

chrmarti:
[...]

PR #132, opened 2022-08-20
Emit newline in between ENV statements in features dockerfile (v1 only)

joshspicer:
fixes https://github.com/devcontainers/cli/issues/131


samruddhikhandale (approved):
✨ 🚀

PR #137, opened 2022-08-22
`devcontainer features info <..>` command

joshspicer:
A simple command that uses the local OCI implementation to fetch information on a dev container feature published to an OCI repository.

Today, this command returns all published tags for a given `featureId`.  The output can either be human-formatted (Default), or JSON.

image

### DIFF
--- a/src/spec-node/featuresCLI/info.ts
+++ b/src/spec-node/featuresCLI/info.ts
@@ -1,6 +1,5 @@
 import path from 'path';
 import { Argv } from 'yargs';
-import { CLIHost } from '../../spec-common/cliHost';
 import { getFeatureRef, getPublishedVersions } from '../../spec-configuration/containerFeaturesOCI';
 import { Log, LogLevel, mapLogLevel } from '../../spec-utils/log';
 import { createLog } from '../devContainers';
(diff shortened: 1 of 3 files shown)


src/spec-node/featuresCLI/info.ts
```diff
@@ -0,0 +1,85 @@
...
+		if (asJson) {
+			output.raw(JSON.stringify({}), LogLevel.Info);
+		} else {
+			output.raw(`No published versions found for feature '${featureId}'\n`, LogLevel.Error);
+		}
+		process.exit(1);
+	}
+
+	const data: { publishedVersions: string[] } = {
+		publishedVersions
```
samruddhikhandale, line 66: Can we sort the `publishedVersions` in descending order to make it more readable ? 😄 
 
https://www.npmjs.com/package/semver 's `rsort` can help, provided the input `list` does not contain `latest` as it only expects semvers.

src/spec-node/featuresCLI/info.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function featuresInfoOptions(y: Argv) {
+	return y
+		.options({
+			'log-level': { choices: ['info' as 'info', 'debug' as 'debug', 'trace' as 'trace'], default: 'info' as 'info', description: 'Log level.' },
+			'json': { type: 'boolean', default: false, description: 'Output in JSON format' },
+		})
+		.positional('featureId', { type: 'string', demandOption: true, description: 'Feature Id' })
+		.check(_argv => {
+			return true;
```
chrmarti, line 18: Is `check` needed when it always returns true?

src/spec-node/featuresCLI/info.ts
```diff
@@ -0,0 +1,85 @@
...
+import { Log, LogLevel, mapLogLevel } from '../../spec-utils/log';
+import { createLog } from '../devContainers';
+import { UnpackArgv } from '../devContainersSpecCLI';
+import { getPackageConfig } from '../utils';
+
+export function featuresInfoOptions(y: Argv) {
+	return y
+		.options({
+			'log-level': { choices: ['info' as 'info', 'debug' as 'debug', 'trace' as 'trace'], default: 'info' as 'info', description: 'Log level.' },
+			'json': { type: 'boolean', default: false, description: 'Output in JSON format' },
```
chrmarti, line 14: `up` is using `--log-format json`. Should we use `--output-format json` here? (`--log-format json` might make sense here too, we can add it later if needed.)

chrmarti (changes requested):

src/spec-node/featuresCLI/info.ts
joshspicer: What is the correct sort? 

```
 latest
 1
 1.0
 1.0.0
 1.0.1
```

or 

```
1.0.0
1.0.1
```

resort only works on fully resolved versions (1.0.0 and 1.0.1)

joshspicer: I landed on this output:
```
Published Versions: 
   latest
   1
   1.0
   1.0.0
   1.0.1
```

joshspicer:
updated: 
image

chrmarti (approved):

PR #142, opened 2022-08-26
use preferred syntax for docker-in-docker feature

joshspicer:
We prefer using the full OCI reference to the major version of the feature, over the older style syntax (that we are phasing out).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,7 +20,7 @@
 	"remoteUser": "node",
 	
 	"features": {
-		"docker-in-docker": "latest"
+		"ghcr.io/devcontainers/features/docker-in-docker:1": {}
 	},
 	
 	"customizations": {


samruddhikhandale (approved):

.devcontainer/devcontainer.json
```diff
@@ -20,7 +20,7 @@
 	"remoteUser": "node",
 	
 	"features": {
-		"docker-in-docker": "latest"
+		"ghcr.io/devcontainers/features/docker-in-docker:1": {}
```
chrmarti, line 23: `"latest"` would still work, wouldn't it?

chrmarti (changes requested):

.devcontainer/devcontainer.json
joshspicer: `latest` is implied as the [default value from the `devcontainer-feature.json`](https://github.com/devcontainers/features/blob/main/src/docker-in-docker/devcontainer-feature.json#L15).  It is redundant to add it here.  (And additionally, to a new user it is not clear that this value maps to the version of _docker_).  I prefer the empty object as it indicates "use the defaults the author has set for me"

chrmarti (approved):

Issue #148, opened 2022-08-29
Spaces not properly escaped in Feature option values


Chuxel:
Having a space in a Feature option value can cause a build to fail or only the part before the space to be picked up.

Repro:

1. Create a repo from https://github.com/devcontainers/feature-template
2. Add the following `.devcontainer.json`: 
    ```
    {
        "image": "mcr.microsoft.com/devcontainers/base:ubnutu",
        "features": {
            "./src/hello": {
               "greeting": "Hello there"
           }
        }
    }
    ```
4. Reopen in container
5. Go to a terminal and type `hello`.

**Expected:** Command outputs "Hello there, vscode!"
**Actual:** Command outputs "Hello, vscode!"

I can repro this in more complex features (e.g. [this one](https://github.com/Chuxel/feature-library/tree/main/src/nix)) as well - it seems to be across the board.

PR #149, opened 2022-08-30
Properly source feature options

joshspicer:
Fixes https://github.com/devcontainers/cli/issues/148

Instead of trying to use `xargs` to parse the .env file, directly source the .env file with `set -a` enabled.

```
-a
When this option is on, the export attribute shall be set for each variable to which an assignment is performed;

```

In the generated Dockerfile, this env file is simply sources with the `.` operator (The most portable/POSIX-y way to source https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot)

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -244,7 +244,7 @@ export function getFeatureLayers(featuresConfig: FeaturesConfig) {
 			result += `
 				
 RUN cd /tmp/build-features/${feature.consecutiveId} \\
-&& export $(cat devcontainer-features.env | xargs) \\
+&& . ./devcontainer-features.env \\
 && chmod +x ./install.sh \\
 && ./install.sh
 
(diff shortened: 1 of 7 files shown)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -244,7 +244,7 @@ export function getFeatureLayers(featuresConfig: FeaturesConfig) {
 			result += `
 				
 RUN cd /tmp/build-features/${feature.consecutiveId} \\
-&& export $(cat devcontainer-features.env | xargs) \\
+&& . ./devcontainer-features.env \\
```
chrmarti, line 247: You could use `set -a` to export all assignments and then `set +a` to turn that off as in https://unix.stackexchange.com/a/79084. That would keep the env file simpler. Also seems to work with /bin/sh. @Chuxel used this in some of the scripts.

chrmarti (changes requested):

src/spec-configuration/containerFeaturesConfiguration.ts
Chuxel: Yeah, it's the simplest method I've found.

chrmarti (approved):
Thanks!

Issue #155, opened 2022-09-02
Unable to build multi-platform images without also pushing them


natescherer:
Hello! Looks like when the --no-push parameter on devcontainer build was dropped it stopped being possible to build a multi-platform container without pushing it. The current code sets the behavior when you don't push to automatically add --load to the docker buildx build call, which is incompatible with manifest based Docker images (which multi-platform images must be).

I will cook up a PR to fix this.

Chuxel:
@natescherer With buildx, you can specify multiple platforms to build in a single call. Are you looking to manually construct the manifest?

natescherer:
So what I'm trying to do is build multi-platform image support into https://github.com/devcontainers/ci. Currently that's set up with three stages, build, run an (optional) command inside the image, and then push. I was trying to maintain the same workflow for multi-container images. Current thinking is adding `--output-type` and `--output-path` options to the CLI that map to the `docker buildx --output`, which would let me dump the container in OCI format out, then push it to GHCR.

Does that make sense?

Chuxel:
Given the dev container CLI supports `--platform` which like buildx supports a list, so wouldn't wiring that up be a bit easier?

e.g., this should work: `devcontainer build --platform "linux/amd64,linux/arm64"`

natescherer:
Oh yeah, that's what I'm doing. But given the only way that call works now is with `--push`, there's no way to run the intermediate command (unless I'm mistaken). That's what I'm trying to preserve: the ability to build, test, and then push that's currently in the Action.

Chuxel:
Do you mean the `runCmd`?  Yeah, that would require `--load` unfortunately since it ultimately is a "docker run + exec".  So you'd end up having to do an image pull or something after it was created anyway - another single architecture build that hopefully reuses the cache.  So you're thinking that having an output type would allow you to export and then import and push?

natescherer:
Yeah, `runCmd` is what I was talking about, sorry for not specifying. That part actually still works, because the `devcontainer up` that `runCmd` does recognizes it doesn't have a local image it can use an builds it again just for the local platform (not super efficient, obviously but gets the job done). (If you want to see for yourself, check this out: https://github.com/natescherer/sre-devcontainer/runs/8139243944?check_suite_focus=true)

I'm still trying to figure out how to push the OCI tarball to GHCR, but, if I can figure that out, I do think that adding output options to the CLI would solve the problem.

natescherer:
[...]

PR #166, opened 2022-09-08
Add --output option to devcontainer build

natescherer:
This is related to what was discussed in #155.

This PR adds new options `--output-type` and `--output-dest` that pass through to `docker buildx build --output`.

If there's anything that needs added or changed, please let me know!


natescherer:

src/test/cli.build.test.ts
```diff
@@ -170,5 +170,61 @@ describe('Dev Containers CLI', function () {
 			assert.equal(response.imageName[0], image1);
 			assert.equal(response.imageName[1], image2);
 		});
+
+		it('should fail with only --output-type and no --output-dest', async () => {
```
chrmarti, line 174: Would it make sense to have a test that checks a tar is created at the output location when the cli options are correct?

src/spec-node/singleContainer.ts
```diff
@@ -174,7 +174,11 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 		if (buildParams.buildxPush) {
 			args.push('--push');
 		} else {
-			args.push('--load'); // (short for --output=docker, i.e. load into normal 'docker images' collection)
+			if (buildParams.buildxOutputType) { 
+				args.push('--output', `type=${buildParams.buildxOutputType},dest=${buildParams.buildxOutputDest}`);
```
chrmarti, line 178: buildx seems to have a few variations on what options an output type can take: https://docs.docker.com/engine/reference/commandline/buildx_build/#output

Should we have just a single option that is passed directly to buildx's `--output` to cover all cases? This would also be useful if/when buildx adds more options to `--output`.

chrmarti (changes requested):
Thanks for the PR! Left a few comments for changes / discussion.

src/spec-node/singleContainer.ts
natescherer: I'm fine with doing that, wasn't sure which way was more in the spirit of the current option structure. Def a case to be made that a transparent `--output` would be less upkeep as Docker changes. Will do it.

src/test/cli.build.test.ts
natescherer: Good call. On it.

natescherer:
Both requested changes have been committed. Hopefully the test works; I was not able to get yarn test to work in my local env for reasons I couldn't quite figure out.

natescherer:
Previous test failed due to unbalanced quotes, looks like. Latest commit should fix it.

chrmarti:
@natescherer There are some test failures. Could you look into these? The change looks good otherwise.

natescherer:
Looks like I was using a non-Dockerfile config to try and run the test which wouldn't work. Changed it and hopefully the test will pass now!

natescherer:
@chrmarti Could you please trigger the tests again when you get a minute? Looks like they don't automatically run, probably since I'm a first-time contributor to this repo.

natescherer:
Looks like synchronizing from main caused there to be a missing comma. Last commit should fix.

chrmarti (approved):
LGTM!

src/test/cli.build.test.ts
```diff
@@ -170,5 +171,29 @@ describe('Dev Containers CLI', function () {
...
+			} catch (error) {
+				assert.equal(error.error.code, 1, 'Should fail with exit code 1');
+				const res = JSON.parse(error.stdout);
+				assert.equal(res.outcome, 'error');
+				assert.match(res.message, /cannot be used with/);
+			}
+			assert.equal(success, false, 'expect non-successful call');
+		});
+
+		it('file /tmp/output.tar should exist when using --output type=oci,dest=/tmp/output.tar', async () => {
```
chrmarti, line 189: The `docker` builder doesn't support that option it seems. Could you change the test to first create a builder and remove that builder at the end of the test?

Will you need to do the same for the GitHub action?

chrmarti (changes requested):

src/test/cli.build.test.ts
natescherer: Also I'm pretty sure the GitHub action environment uses the non-default buildx driver by default but I'll make sure to test that out!

natescherer: I liked your idea of removing the builder at test end rather than leaving it sit. Done.

src/test/cli.build.test.ts
```diff
@@ -170,5 +171,33 @@ describe('Dev Containers CLI', function () {
...
+			const testFolder = `${__dirname}/configs/dockerfile-with-target`; 
+			const outputPath = `/tmp/output.tar`;
+			await shellExec(`docker buildx create --name ocitest'`);
```
chrmarti, line 192: Seeing an extra single quote at the end of the string.

Could you replace the use of `/tmp` with `os.tmpdir()`? That will make it work on macOS too. Thanks!

chrmarti (changes requested):

src/test/cli.build.test.ts
natescherer: Sorry about the messed up quotes, and good call on using os.tmpdir(). Both should be fixed.

src/test/cli.build.test.ts
```diff
@@ -170,5 +172,33 @@ describe('Dev Containers CLI', function () {
...
+			await shellExec('docker buildx create --name ocitest');
+			await shellExec('docker buildx use ocitest');
+			const res = await shellExec(`${cli} build --workspace-folder ${testFolder} --output 'type=oci,dest=${outputPath}'`);
+			await shellExec('docker buildx use default');
+			await shellExec('docker buildx rm ocitest');
+			const response = JSON.parse(res.stdout);
+			assert.equal(response.outcome, 'success');
+			assert.equal(fs.existsSync(outputPath), true);
+		});
+
```
samruddhikhandale, line 202: super nit: Can we remove the extra line?

natescherer: Done!

samruddhikhandale (commented):
@natescherer looks like there are some merge conflicts. Could you help resolve them? thanks!

natescherer:
@samruddhikhandale I re-merged main with my branch, should be good now hopefull!

chrmarti (approved):

chrmarti:
@natescherer Merged, thanks!

PR #171, opened 2022-09-09
Add --additional-features option

stuartleeks:
Initial work to enable additional features to be specified via the CLI

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -435,7 +435,7 @@ function updateFromOldProperties<T extends { features: (Feature & { extensions?:
 
 // Generate a base featuresConfig object with the set of locally-cached features, 
 // as well as downloading and merging in remote feature definitions.
-export async function generateFeaturesConfig(params: { extensionPath: string; cwd: string; output: Log; env: NodeJS.ProcessEnv; skipFeatureAutoMapping: boolean }, dstFolder: string, config: DevContainerConfig, getLocalFeaturesFolder: (d: string) => string, additionalFeatures?: Record<string, string | boolean | Record<string, string | boolean>>) {
+export async function generateFeaturesConfig(params: { extensionPath: string; cwd: string; output: Log; env: NodeJS.ProcessEnv; skipFeatureAutoMapping: boolean }, dstFolder: string, config: DevContainerConfig, getLocalFeaturesFolder: (d: string) => string, additionalFeatures: Record<string, string | boolean | Record<string, string | boolean>>) {
 	const { output } = params;
 
 	const workspaceRoot = params.cwd;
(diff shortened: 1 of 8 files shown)


src/spec-node/devContainersSpecCLI.ts
```diff
@@ -735,8 +746,9 @@ async function readConfiguration({
 			};
 		}
 
-		const needsFeaturesConfig = includeFeaturesConfig || (includeMergedConfig && (!container || !experimentalImageMetadata));
-		const featuresConfiguration = needsFeaturesConfig ? await readFeaturesConfig(params, pkg, configuration.config, extensionPath, skipFeatureAutoMapping) : undefined;
+		const additionalFeatures = additionalFeaturesJson ? jsonc.parse(additionalFeaturesJson) as Record<string, string | boolean | Record<string, string | boolean>> : undefined;
+		const needsFeaturesConfig = includeFeaturesConfig || additionalFeatures || (includeMergedConfig && (!container || !experimentalImageMetadata));
```
chrmarti, line 750: The features config shouldn't be needed when the container already exists. The container will have the config from the additional features on its devcontainers.metadata label. (Fetching the features config can be expensive.) Please remove the `additionalFeatures` from the condition.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -491,19 +491,38 @@ async function prepareOCICache(dstFolder: string) {
 	return ociCacheDir;
 }
 
-function featuresToArray(config: DevContainerConfig): DevContainerFeature[] | undefined {
-	if (!config.features) {
+function featuresToArray(config: DevContainerConfig, additionalFeatures?: Record<string, string | boolean | Record<string, string | boolean>>): DevContainerFeature[] | undefined {
```
chrmarti, line 494: Please make `additionalFeatures` a mandatory argument here and everywhere else where we want to make sure we don't forget pass it on.

chrmarti (changes requested):
Looks great. Left two comments for change / discussion. Thanks!

src/spec-node/devContainersSpecCLI.ts
stuartleeks: Good point - updated

src/spec-configuration/containerFeaturesConfiguration.ts
stuartleeks: Done 😄

chrmarti:
> @chrmarti - I've pushed an update that hopefully addresses your feedback

@stuartleeks Still seeing the additional clause in `readConfiguration`. I think we do without it to avoid reading the entire features config when the container already exists:
image

stuartleeks:
Oops! I'm not sure what happened there - sorry!

I have reapplied the change to remove `additionalFeatures` from the `needsFeaturesConfig` expression.

PR #173, opened 2022-09-13
GPU requirement and auto-detect NVIDIA extensions

chris-major-improbable:
This PR builds upon the proposal in https://github.com/devcontainers/spec/issues/82
When launching a devcontainer requesting GPUs it also detects the presence of NVIDIA docker extensions, and adds the `--gpus all` flag.

This means that the same devcontainer can be used by developments that don't all have GPU's.

It also does the equivalent for docker-compose scripts, adding the GPU access to the service specified in the `devcontainer.json` file


src/spec-node/singleContainer.ts
```diff
@@ -291,6 +291,17 @@ export async function findDevContainer(params: DockerCLIParameters | DockerResol
 	return details.filter(container => container.State.Status !== 'removing')[0];
 }
 
+export async function extraRunArgs(params: DockerCLIParameters | DockerResolverParameters, config: DevContainerFromDockerfileConfig | DevContainerFromImageConfig) {
```
chrmarti, line 294: Can the same be done for the Docker Compose case? (dockerCompose.ts)

chrmarti (changes requested):
Thanks for the PR! I have left a comment.

src/spec-node/singleContainer.ts
chris-major-improbable: @chrmarti that's a really good spot, I hadn't seen there was a separate code path there, I'll take a look.

chris-major-improbable: @chrmarti I've had a look through the code and the documentation for `docker compose`, and I'm nor sure what the most sensible approach would be here.

The 'normal' `docker-compose.yml` file can specify that a specific service needs a gpu capability, and will fail to start if one isn't available. This is the same situation as when using `Dockerfile`, and specifying the extra arguments, and so not a great experience for developers on mixed hardware.

The only way I can see to fix this is to modify the `docker-compose.yml` before it is used - either adding or removing the gpu capability. Removing is seems slightly more logical, as then it's clear which service should have gpu access if it's available.

So my current thoughts are that the logic when using docker-compose should be:
1. If the devontainer.json lists gpu's as required,
2. and the docker-compose.yml has services requiring gpu capabilities,
3. and the nvidia extensions are not available,
4. then create a temporary copy of the yaml without the gpu capabilities, and launch that instead.

How does that sound?

chris-major-improbable: I've implemented the flow described above - let me know hoe it looks

chrmarti: We use overlay docker-compose.yml files in other cases. So you could probably (I haven't tried) enable gpu support in `generateFeaturesComposeOverrideContent()` if required instead of removing it if it is already there. Being additive this would be closer to the single container case.

src/spec-node/dockerCompose.ts
```diff
@@ -244,6 +244,25 @@ ${cacheFromOverrideContent}
 		args.push('-f', composeOverrideFile);
 	}
 
+	const hasGpuRequirement = config.hostRequirements?.gpu;
+	const addGpuCapability = hasGpuRequirement && await checkDockerSupportForGPU(params);
+	if (addGpuCapability) {
+		const composeOverrideContent = `services:
+  ${config.service}:
```
chrmarti, line 251: `generateFeaturesComposeOverrideContent()` would be the better spot for adding this. The function name is increasingly misleading, but that's the override file for everything we need to create the container. (The code here takes care of the building.)

src/spec-configuration/configuration.ts
```diff
@@ -21,10 +21,17 @@ export type UserEnvProbe = 'none' | 'loginInteractiveShell' | 'interactiveShell'
...
+	cores?: number;
+	memory?: string;
+}
+
+
 export interface HostRequirements {
 	cpus?: number;
 	memory?: string;
 	storage?: string;
+	gpu?: boolean | HostGPURequirements;
```
chrmarti, line 34: It would make more sense to make the "requirement" fail if there is no GPU. E.g., if the `"gpu"` property is set, you would check if a GPU is available and fail if it is not. Only when `"gpu"` is set and a GPU is available would you enable the GPU for the dev container.

This will make sure that dev containers requiring a GPU can count on one being available. (To make the GPU optional but still automatically add it, maybe `"gpu": "optional"` could be added.)

src/spec-configuration/configuration.ts
```diff
@@ -21,10 +21,17 @@ export type UserEnvProbe = 'none' | 'loginInteractiveShell' | 'interactiveShell'
 
 export type DevContainerConfigCommand = 'initializeCommand' | 'onCreateCommand' | 'updateContentCommand' | 'postCreateCommand' | 'postStartCommand' | 'postAttachCommand';
 
+export interface HostGPURequirements {
```
chrmarti, line 24: I recently added `mergeHostRequirements()` in `imageMetadata.ts` that needs to be updated to also merge the GPU requirements. (The merge logic is to go with the max values.)

chrmarti (commented):
Thanks for the update! I have added a few more comments.

src/spec-configuration/configuration.ts
chrmarti: @jkeech @Chuxel I realize, we don't enforce any of the other host requirements in the CLI. The thinking was that we don't want to prevent anyone from using the dev container. In the Dev Containers extension we show a warning dialog in the UI if the host requirements are not matched and the user can choose to still continue.

chris-major-improbable: @chrmarti Thanks - my motivation for this is all about making it optional, so I'll try and include that.

src/spec-configuration/configuration.ts
chris-major-improbable: I think I've updated - however not 100% sure where that is used, so I'm not very confident about it.

src/spec-node/dockerCompose.ts
chris-major-improbable: I've moved the logic into that function - it reduces the complexity significantly - thanks :)

src/spec-configuration/configuration.ts
chris-major-improbable: I've added to the tests to check the functioning of the `mergeHostRequirements`

src/spec-node/dockerCompose.ts
```diff
@@ -499,6 +508,20 @@ async function generateFeaturesComposeOverrideContent(
...
 
+
+	const hasGpuRequirement = config.hostRequirements?.gpu;
+	const addGpuCapability = hasGpuRequirement && await checkDockerSupportForGPU(params);
+	if (hasGpuRequirement && hasGpuRequirement !== 'optional' && !addGpuCapability) {
+		throw Error('Unable to detect GPU yet a GPU was required - consider marking it as "optional"');
+	}
+	const gpuResources = addGpuCapability ? '' : `
+	deploy:
+	  resources:
```
chrmarti, line 519: Indentation seems to be mixed here. Maybe only use spaces with 2 per indent.

src/spec-configuration/configuration.ts
```diff
@@ -21,10 +21,17 @@ export type UserEnvProbe = 'none' | 'loginInteractiveShell' | 'interactiveShell'
...
+	cores?: number;
+	memory?: string;
+}
+
+
 export interface HostRequirements {
 	cpus?: number;
 	memory?: string;
 	storage?: string;
+	gpu?: boolean | string | HostGPURequirements; // The 'string' option allows us to specify "optional"
```
chrmarti, line 34: Use `'optional'` instead of `string`. (String literal types of TypeScript.)

src/spec-node/dockerCompose.ts
```diff
@@ -499,6 +508,20 @@ async function generateFeaturesComposeOverrideContent(
 	const userCommand = overrideCommand ? [] : composeCommand /* $ already escaped. */
 		|| (composeEntrypoint ? [/* Ignore image CMD per docker-compose.yml spec. */] : ((await imageDetails()).Config.Cmd || []).map(c => c.replace(/\$/g, '$$$$'))); // $ > $$ to escape docker-compose.yml's interpolation.
 
+
+	const hasGpuRequirement = config.hostRequirements?.gpu;
+	const addGpuCapability = hasGpuRequirement && await checkDockerSupportForGPU(params);
+	if (hasGpuRequirement && hasGpuRequirement !== 'optional' && !addGpuCapability) {
+		throw Error('Unable to detect GPU yet a GPU was required - consider marking it as "optional"');
```
chrmarti, line 515: We realized that we don't enforce other requirements. Having a warning would still be valuable, could you write this message to `output.write` with `LogLevel.Warning` instead of throwing an error?

chrmarti (changes requested):
Looks good, left a few more comments. Thanks!

src/spec-node/dockerCompose.ts
```diff
@@ -470,6 +471,12 @@ async function writeFeaturesComposeOverrideFile(
 	}
 }
 
+async function checkDockerSupportForGPU(params: DockerCLIParameters | DockerResolverParameters): Promise<Boolean> {
```
chrmarti, line 474: Move to `utils.ts` and reuse in `singleContainer.ts`?

src/spec-node/dockerCompose.ts
```diff
@@ -501,6 +509,19 @@ async function generateFeaturesComposeOverrideContent(
 	const userCommand = overrideCommand ? [] : composeCommand /* $ already escaped. */
 		|| (composeEntrypoint ? [/* Ignore image CMD per docker-compose.yml spec. */] : ((await imageDetails()).Config.Cmd || []).map(c => c.replace(/\$/g, '$$$$'))); // $ > $$ to escape docker-compose.yml's interpolation.
 
+	const hasGpuRequirement = config.hostRequirements?.gpu;
+	const addGpuCapability = hasGpuRequirement && await checkDockerSupportForGPU(params);
+	if (hasGpuRequirement && hasGpuRequirement !== 'optional' && !addGpuCapability) {
+		throw Error('Unable to detect GPU yet a GPU was required - consider marking it as "optional"');
```
chrmarti, line 515: Could you log a warning instead? (Looks like you missed checking in this change.)

chrmarti (changes requested):
2 more comment. :)

src/spec-node/dockerCompose.ts
chris-major-improbable: Sorry, must have lost that when rebasing to the new main - I'll fix it now

chrmarti (approved):
LGTM! (Looping in our partner team for cross-review. Will merge after that. Thanks!)

chris-major-improbable:
I've taken a bit of a closer look, and rebased onto the latest main - and I see the same failures running the tests locally, so I don't think this is a regression from this PR.

chrmarti:
@chris-major-improbable I have merged in the main branch and fixed an issue with the PR I found when looking into the test failures. Please check the above comment from the CLA bot, we must have added that after you submitted the PR. Thanks!

chris-major-improbable:
@microsoft-github-policy-service agree [company="{Improbable Worlds Ltd}"]

chris-major-improbable:
@microsoft-github-policy-service agree company="{Improbable Worlds Ltd}"

chris-major-improbable:
@microsoft-github-policy-service agree company="Improbable Worlds Ltd"

chrmarti (approved):
LGTM! (Will get a second review now.) Thanks!

joshspicer (approved):
Looks great! Thanks for the contribution.

samruddhikhandale (approved):
Looks great. Thanks for the contribution! ✨

PR #174, opened 2022-09-13
Add `--skip-post-attach` to skip running the post attach command

joshaber:
Teach `up` and `run-user-commands` the `--skip-post-attach` option so that attaching clients can take ownership for running post attach themselves.


src/spec-common/injectHeadless.ts
```diff
@@ -319,7 +320,9 @@ export async function runPostCreateCommands(params: ResolverParameters, containe
 		return 'skipNonBlocking';
 	}
 
-	await runPostAttachCommand(params, containerProperties, config, remoteEnv);
+	if (!params.skipPostAttach) {
+		await runPostAttachCommand(params, containerProperties, config, remoteEnv);
```
joshaber, line 324: Should we return a different status from this method if we skipped post attach? It's not clear to me how the return value is used.

chrmarti: We could return 'skipPostAttach' and handle it [here](https://github.com/microsoft/vscode-remote-containers/blob/19a2dc727aa2d1feb3983ffff345ed7341dc2950/src/common/injectHeadless.ts#L596). It wouldn't change what we do there (except for accepting the new value) though and we can add the additional return value later if we need.

chrmarti (approved):
LGTM!

src/spec-common/injectHeadless.ts
joshaber: Sounds good on waiting until we need it 👍

edgonmsft (approved):
LGTM

Issue #170, opened 2022-09-08
Feature installation logs should clearly indicate when a feature starts/stops installing


jkeech:
Currently, when you install a set of devcontainer features, the output of each one is in the middle of the Dockerfile build and not clearly attributable to the feature that caused it. This makes it hard to see which features are running, and if there's an error, to see which feature failed.

It would be nice if the logs showed the feature that is executing along with the inputs and contact info for the feature if the user has any questions or runs into issues using it. 

### Current CLI experience

In the screenshot below, it's hard to make out that the `docker-in-docker` feature starts on line 122 and it's easy to miss when it ends and switches to another feature. It also doesn't show which options were specified for the feature. 

![devcontainer-logs](https://user-images.githubusercontent.com/746020/189233562-e6f86d79-c531-48b8-919e-68399c0c418f.png)

### Other solutions 
CI/CD systems often do a good job of this so you can see which steps are running and what their inputs are.

GitHub Actions:
![image](https://user-images.githubusercontent.com/746020/189233166-af490b11-207e-4d1d-a752-84b4cef32392.png)

Azure Pipelines:
![image](https://user-images.githubusercontent.com/746020/189233263-7013b0cb-f371-4bc2-ac10-8ac3e8f73037.png)

I propose that we adopt a similar pattern in the devcontainer CLI. This could be accomplished by injecting a custom wrapper script into the Dockerfile which in turn invokes the feature's `install.sh` instead of directly injecting the install script into the Dockerfile. The wrapper script would print out the feature being invoked, the options used, and the contact info for the feature in a clear header before invoking the feature's install script. If the install script exits with a non-zero exit code, we could print out a custom error message saying which feature failed to install.

PR #178, opened 2022-09-17
Improve feature installation logs

jkeech:
Add a header to the output logs for each feature that is executed to provide info about the feature

Example:
```
#15 [dev_containers_target_stage 4/4] RUN cd /tmp/build-features/docker-in-docker_2 && chmod +x ./devcontainer-features-install.sh && ./devcontainer-features-install.sh
#15 0.521 ===========================================================================
#15 0.521 Feature       : Docker (Docker-in-Docker)
#15 0.521 Description   : Create child containers *inside* a container, independent from the host's docker instance. Installs Docker extension in the container along with needed CLIs.
#15 0.521 Id            : ghcr.io/devcontainers/features/docker-in-docker
#15 0.521 Version       : 1.0.4
#15 0.521 Documentation : https://github.com/devcontainers/features/tree/main/src/docker-in-docker
#15 0.521 Options       :
#15 0.521     VERSION=latest
#15 0.521     MOBY=true
#15 0.521     DOCKERDASHCOMPOSEVERSION=v1
#15 0.521 ===========================================================================
#15 0.669 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES=buster bullseye bionic focal jammy
#15 0.669 Distro codename  'bullseye'  matched filter  'buster bullseye bionic focal jammy'
#15 0.675 Running apt-get update...
#15 0.714 Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]
...
```

In order to print out feature information in the middle of the docker build, we are now injecting a wrapper script as the command to `RUN` instead of directly invoking the feature's `install.sh`. The wrapper script prints the header, sources the `devcontainer-features.env` and then invokes the feature's `install.sh`. This simplifies the `RUN` command in the Dockerfile and gives us a nice spot to handle both before and after a feature executes.

If a feature fails to install, a clear error message is printed saying which feature failed with a pointer to the documentation for that feature. This is accomplished with `trap on_exit EXIT` that runs the `on_exit()` function before the script exits.

Added unit and e2e tests to cover the v2 feature install path.

Fixes: https://github.com/devcontainers/cli/issues/170

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -225,25 +225,37 @@ USER $_DEV_CONTAINERS_IMAGE_USER
 `;
 }
 
-function getFeatureInstallWrapperScriptTemplate() {
-    return `#!/bin/sh
+export function getFeatureInstallWrapperScript(feature: Feature, featureSet: FeatureSet, options: string[]): string {
+	const id = escapeQuotesForShell(featureSet.sourceInformation.userFeatureIdWithoutVersion ?? 'Unknown');
+	const name = escapeQuotesForShell(feature.name ?? 'Unknown');
+	const description = escapeQuotesForShell(feature.description ?? '');
+	const version = escapeQuotesForShell(feature.version ?? '');
+	const documentation = escapeQuotesForShell(feature.documentationURL ?? '');
+	const optionsIndented = escapeQuotesForShell(options.map(x => `    ${x}`).join('\n'));
+
+	const errorMessage = `ERROR: Feature "${name}" (${id}) failed to install!`;
+	const troubleshootingMessage = documentation
+		? ` Look at the documentation at ${documentation} for help troubleshooting this error.`
+		: '';
+
+	return `#!/bin/sh
 set -e
 
 on_exit () {
 	[ $? -eq 0 ] && exit
-	echo 'ERROR: Feature "#{name}" (#{id}) failed to install! Look at the documentation at #{documentation} for help troubleshooting this error.'
+	echo '${errorMessage}${troubleshootingMessage}'
 }
 
 trap on_exit EXIT
 
 echo ===========================================================================
-echo 'Feature       : #{name}'
-echo 'Description   : #{description}'
-echo 'Id            : #{id}'
-echo 'Version       : #{version}'
-echo 'Documentation : #{documentation}'
+echo 'Feature       : ${name}'
+echo 'Description   : ${description}'
+echo 'Id            : ${id}'
+echo 'Version       : ${version}'
+echo 'Documentation : ${documentation}'
 echo 'Options       :'
-echo '#{options}'
+echo '${optionsIndented}'
 echo ===========================================================================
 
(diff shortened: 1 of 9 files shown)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -225,6 +225,54 @@ USER $_DEV_CONTAINERS_IMAGE_USER
...
+. ./devcontainer-features.env
+set +a
+
+chmod +x ./install.sh
+./install.sh
+`;
+}
+
+export function getFeatureInstallWrapperScript(feature: Feature, featureSet: FeatureSet, options: string[]): string {
+	return getFeatureInstallWrapperScriptTemplate()
```
chrmarti, line 259: You could use the `${expression}` syntax inside the above template string to simplify the replacement.

chrmarti (approved):
LGTM! Left a suggestion.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -225,6 +225,54 @@ USER $_DEV_CONTAINERS_IMAGE_USER
 `;
 }
 
+function getFeatureInstallWrapperScriptTemplate() {
+    return `#!/bin/sh
+set -e
+
+on_exit () {
+	[ $? -eq 0 ] && exit
+	echo 'ERROR: Feature "#{name}" (#{id}) failed to install! Look at the documentation at #{documentation} for help troubleshooting this error.'
```
joshspicer, line 234: If the feature doesn't have a `documentationUrl`, this sentence will read a little bit weird.

joshspicer: Same with `name`, which this PR makes optional

joshspicer (dismissed):
woo!

src/spec-configuration/containerFeaturesConfiguration.ts
jkeech: Agreed. I can tweak the error message in those cases.

src/spec-configuration/containerFeaturesConfiguration.ts
jkeech: Good call. We still need to do the quote escaping, but I think it will be cleaner with the string interpolation. I'll play around with it and see how it looks.

src/spec-configuration/containerFeaturesConfiguration.ts
jkeech: Updated `name` and `id` to default to "Unknown" if they are undefined to make it read a bit better. I also removed the second sentence from the message if no documentation url is defined.

src/spec-configuration/containerFeaturesConfiguration.ts
jkeech: Updated to use string interpolation. I agree it's cleaner this way, and it actually makes it easier to customize the header a bit more when some optional properties are undefined. 👍

chrmarti (approved):

joshspicer (approved):

PR #188, opened 2022-09-23
Image metadata

chrmarti:
Implements https://github.com/devcontainers/spec/pull/95.


src/spec-common/injectHeadless.ts
```diff
@@ -125,6 +126,37 @@ export interface CommonDevContainerConfig {
 	userEnvProbe?: UserEnvProbe;
 }
 
+export interface CommonContainerMetadata {
```
joshspicer, line 129: For my future reference, what does the `Common` prefix signify?

Could this interface be included in `CommonDevContainerConfig`?

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -872,8 +919,10 @@ export async function doExec({
 		if (!container) {
 			bailOut(common.output, 'Dev container not found.');
 		}
-		const containerProperties = await createContainerProperties(params, container.Id, workspaceConfig.workspaceFolder, config.remoteUser);
-		const updatedConfig = containerSubstitute(cliHost.platform, config.configFilePath, containerProperties.env, config);
+		const imageMetadata = getImageMetadataFromContainer(container, config, [], experimentalImageMetadata, output).config;
+		const mergedConfig = mergeConfiguration(config.config, imageMetadata);
+		const containerProperties = await createContainerProperties(params, container.Id, workspaceConfig.workspaceFolder, mergedConfig.remoteUser);
+		const updatedConfig = containerSubstitute(cliHost.platform, config.config.configFilePath, containerProperties.env, mergedConfig);
```
samruddhikhandale, line 925: should it be `config.configFilePath` instead ?

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -872,8 +919,10 @@ export async function doExec({
 		if (!container) {
 			bailOut(common.output, 'Dev container not found.');
 		}
-		const containerProperties = await createContainerProperties(params, container.Id, workspaceConfig.workspaceFolder, config.remoteUser);
-		const updatedConfig = containerSubstitute(cliHost.platform, config.configFilePath, containerProperties.env, config);
+		const imageMetadata = getImageMetadataFromContainer(container, config, [], experimentalImageMetadata, output).config;
+		const mergedConfig = mergeConfiguration(config.config, imageMetadata);
```
samruddhikhandale, line 923: should it be `config` instead ?

src/spec-common/injectHeadless.ts
chrmarti: `common` is for "any container engine" (sources are in a `spec-common` folder). Maybe not the best naming, I introduced it when we added support for attaching to k8s containers in the extension. Currently this distinction is not really used in the CLI, but we might revive it in the future.

src/spec-node/devContainersSpecCLI.ts
chrmarti: `config` here is this combination of a config object with variable substitution applied (`config.config`) and raw config (`config.raw`). So `config.config.configFilePath` is correct.

src/spec-node/devContainersSpecCLI.ts
chrmarti: (Like above:) `config` has `config` and `raw` properties. I introduced an interface to carry the config with variables substituted and the raw config around:
```ts
interface SubstitutedConfig<T> {
	config: T;
	raw: T;
	substitute: SubstituteConfig;
}
```

samruddhikhandale (approved):
LGTM ✨

joshspicer (approved):

joshspicer (approved):

src/spec-node/imageMetadata.ts
```diff
@@ -0,0 +1,375 @@
...
+	entrypoints?: string[];
+	onCreateCommands?: (string | string[])[];
+	updateContentCommands?: (string | string[])[];
+	postCreateCommands?: (string | string[])[];
+	postStartCommands?: (string | string[])[];
+	postAttachCommands?: (string | string[])[];
+	shutdownAction?: 'none' | 'stopContainer' | 'stopCompose';
+}
+
+export function mergeConfiguration(config: DevContainerConfig, imageMetadata: ImageMetadataEntry[]): MergedDevContainerConfig {
```
greggroth, line 108: What is the purpose or advantage of having a type `MergedDevContainerConfig` that's distinct form `DevContainerConfig`?

PR #193, opened 2022-09-28
adds `templates package` command

samruddhikhandale:
Adds `template package` command according to the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-templates.md)

image

Sample - Packaging single Template
image

Sample - Packaging collection
image

### DIFF
--- a/src/spec-node/collectionCommonUtils/packageCommandImpl.ts
+++ b/src/spec-node/collectionCommonUtils/packageCommandImpl.ts
@@ -1,0 +1,140 @@
+import tar from 'tar';
+import { PackageCommandInput } from './package';
+import { isLocalFile, isLocalFolder, mkdirpLocal, readLocalDir, readLocalFile, rmLocal } from '../../spec-utils/pfs';
+import { LogLevel } from '../../spec-utils/log';
+import path from 'path';
+
+export interface SourceInformation {
+	source: string;
+	owner?: string;
+	repo?: string;
+	tag?: string;
+	ref?: string;
+	sha?: string;
+}
+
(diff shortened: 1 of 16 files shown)


src/spec-node/collectionCommonUtils/packageCommandImpl.ts
```diff
@@ -0,0 +1,140 @@
...
+export interface SourceInformation {
+	source: string;
+	owner?: string;
+	repo?: string;
+	tag?: string;
+	ref?: string;
+	sha?: string;
+}
+
+export const OCICollectionFileName = 'devcontainer-collection.json';
```
joshspicer, line 16: I think style-wise typescript constants should be BIG_LIKE_THIS

joshspicer (approved):
looks good to me!  Thanks for splitting out that common packaging utility.

src/spec-node/collectionCommonUtils/packageCommandImpl.ts
samruddhikhandale: I tried to match the casing with the already existing constants 😅 

image

we do have some caps with _ variables but only for Features stuff 🤔 

image

I don't have a preference though.

chrmarti: Camel-case is fine. I often lower-case abbreviations when they are at the start (but upper-case in the middle). E.g., `ociCollectionFileName`, but I don't have a strong preference on that.

chrmarti (approved):
LGTM!

PR #198, opened 2022-09-29
adds `templates publish` command

samruddhikhandale:
Based on top of [samruddhikhandale/templates-package](https://github.com/devcontainers/cli/pull/193) branch.

Adds a `template publish` CLI command according to the [spec](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-templates.md)

image

Sample - publishing templates

image

image

image


src/spec-node/templatesCLI/publish.ts
```diff
@@ -0,0 +1,96 @@
...
+    const cliHost = await getCLIHost(cwd, loadNativeModule);
+    const output = createLog({
+        logLevel: mapLogLevel(inputLogLevel),
+        logFormat: 'text',
+        log: (str) => process.stdout.write(str),
+        terminalDimensions: undefined,
+    }, pkg, new Date(), disposables);
+
+    // Package templates
+    const outputDir = path.join(os.tmpdir(), '/templates-output');
```
chrmarti, line 53: Make this a unique folder name. Appending `Date.now()` might be enough, but I have started using the `uuid` package in vscode-remote-containers to be on the safe side, e.g.: `/templates-output-${uuid.v4()}`.

Could you apply the same fix in the features code?

chrmarti (changes requested):
LGTM! Left one comment to avoid tmp folder conflicts.

src/spec-node/templatesCLI/publish.ts
samruddhikhandale: Sure, added changes. 👍

chrmarti (approved):
LGTM!

PR #201, opened 2022-10-03
Template: Updates package/publish commands according to spec

samruddhikhandale:
Updates according to https://github.com/devcontainers/spec/pull/105

For templates - 
- Removes `replaceIn` prop
- Adds code to configure `type` prop automatically
- Adds `fileCount` prop


chrmarti (approved):
👍

PR #206, opened 2022-10-06
Templates & Features Packaging - Throw warning of a missing JSON file and continue

samruddhikhandale:
When packaging a collection of templates with `templates publish` or `feature publish` cmd , it errors out when a sub-folder does not contain files meant for templates publishing (eg. `devcontainer-*.json`).

This PR warns about empty directories, but still continue. That would make the user aware of a missing JSON file, but still support the case where this is expected.

### DIFF
--- a/src/spec-node/collectionCommonUtils/packageCommandImpl.ts
+++ b/src/spec-node/collectionCommonUtils/packageCommandImpl.ts
@@ -162,19 +162,21 @@ export async function packageCollection(args: PackageCommandInput, collectionTyp
 		output.write(`Processing ${collectionType}: ${c}...`, LogLevel.Info);
 		if (!c.startsWith('.')) {
 			const folder = path.join(srcFolder, c);
+
+			// Validate minimal folder structure
+			const devcontainerJsonName = `devcontainer-${collectionType}.json`;
+
+			if (!(await isLocalFile(path.join(folder, devcontainerJsonName)))) {
+				output.write(`(!) WARNING: ${collectionType} '${c}' is missing a ${devcontainerJsonName}. Skipping... `, LogLevel.Error);
+				continue;
+			}
+
 			const tmpSrcDir = path.join(os.tmpdir(), `/templates-src-output-${Date.now()}`);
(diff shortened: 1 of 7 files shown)


joshspicer:
I feel like we shouldn't need a flag for this, but rather just (1) not allow empty directories or (2) always allow empty directories and automatically ignore. 

Perhaps there is an example you have where this indeed the best way to do it?

chrmarti:
> I feel like we shouldn't need a flag for this, but rather just (1) not allow empty directories or (2) always allow empty directories and automatically ignore.
> 
> Perhaps there is an example you have where this indeed the best way to do it?

Or: Warn about empty directories, but still continue. That would make the user aware of a missing JSON file, but still support the case where this is expected.

src/spec-node/collectionCommonUtils/packageCommandImpl.ts
```diff
@@ -171,8 +171,9 @@ export async function packageCollection(args: PackageCommandInput, collectionTyp
 			const devcontainerJsonName = `devcontainer-${collectionType}.json`;
 			const jsonPath = path.join(tmpSrcDir, devcontainerJsonName);
 			if (!(await isLocalFile(jsonPath))) {
-				output.write(`${collectionType} '${c}' is missing a ${devcontainerJsonName}`, LogLevel.Error);
-				return;
+				output.write(`(!) WARNING: ${collectionType} '${c}' is missing a ${devcontainerJsonName}. Skipping... `, LogLevel.Error);
+				await rmLocal(tmpSrcDir, { recursive: true, force: true });
```
chrmarti, line 175: Would it make sense to check for the file before copying the folder? You could avoid copying and removing it.

chrmarti (changes requested):

src/spec-node/collectionCommonUtils/packageCommandImpl.ts
samruddhikhandale: Good point! Modified

chrmarti (approved):
👍

PR #218, opened 2022-10-10
Configure proxy-agent (microsoft/vscode-remote-release#6995)

chrmarti:


src/spec-utils/httpRequest.ts
```diff
@@ -42,12 +45,13 @@ export function request(options: { type: string; url: string; headers: Record<st
 export function headRequest(options: { url: string; headers: Record<string, string> }, output?: Log) {
 	return new Promise<number>((resolve, reject) => {
 		const parsed = new url.URL(options.url);
-		const reqOptions = {
+		const reqOptions: RequestOptions = {
```
joshspicer, line 48: I wonder if we should consider using an HTTP library (axios) instead of doing this plumbing here manually.

joshspicer (approved):

src/spec-utils/httpRequest.ts
chrmarti: Maybe. :)

chrmarti: Not sure how much a wrapper library would help. There is a chance that we can improve on the defaults here (drawing on learnings from VS Code) - though that might also be possible with the wrapper library.

Issue #217, opened 2022-10-09
`scenarios.json` does not support running with existing features


AlexanderLanin:
# Observation
When `ghcr.io/devcontainers/features/python:1` is added as a feature to scenarios.json, it attempts to read local file `path: '/workspaces/my-feature/src/ghcr.io/devcontainers/features/python:1'`.

# Expectation
It will use python feature for testing. Therefore I would see that it behaves correctly regarding `installsAfter` and my stuff can indeed use python.

# Background:
`my-feature` requires python 3.7, therefore I implemented a version detection in case the available version is too old. However as there is a python feature already, I don't want to add too much complex logic to my script. I have an error message which tells people to use https://github.com/devcontainers/features/tree/main/src/python in addition to `my-feature`.
However I cannot verify that this would indeed work!

joshspicer:
Yes, this was also raised in https://github.com/devcontainers/cli/issues/217 and is good to point out. The command currently injects the appropriate relative path in for all Features in the scenario, so this is not expected to work (But I agree would be very handy!)

Issue #163, opened 2022-09-06
Feature Dependencies - installsAfter not respected


virtualstaticvoid:
Hello,

I've written a couple of features for a project I'm working on and I've ran into situations where I'm not sure what the best approach to take is.

For example, I have a feature which involves installing a CLI tool using `npm`. Now, I'm not sure whether this feature should also install Node.js so that it is able to install the CLI tool, since I have no control over the base image, nor the included features in the `devcontainer.json` file.

The power of features, as I understand, is that they can be composed together; making each serve one purpose (installing the given feature) thus providing a simple way to handle all the tooling permutations for many different developers and/or teams whilst avoiding "golden container images"; and similar perhaps in concept to [buildpacks](https://buildpacks.io/).

So it occurred to me that since [`devcontainers/features`](https://github.com/devcontainers/features) already provide features for installing Node.js (and other language runtimes), it would be beneficial to make _my feature_ dependant on the `node` feature. Then I wouldn't need to worry about which base image a developer is using, and the `node` feature would be automatically included before mine, so my feature would install reliably.

Additional scenarios include:

* a Ruby gem which needs to build a native extension and thus requires `build-essential` to be installed.
* a Python package, installed via `pip`, needing to ensure Python and pip are installed.
* a Go CLI tool, which doesn't have a binary distribution, installed using `go install`, needing `go` to be installed.
* a `tflint` ruleset for AWS, which requires Terraform and `tflint` to be installed.

I realise installing packages using a `postCreateCommand` script can work, but it places the onus on the developer to know what features to include to install the given package. For example, I'm building features which install Python packages (e.g. `mkdocs`) for C# .Net developers, who wouldn't know to include the `python` feature.

I imagine a `devcontainer-feature.json` file including a `dependsOn` property, in the same schema as the `devcontainer.json` [`features` object](https://github.com/devcontainers/spec/blob/main/proposals/devcontainer-features.md#devcontainerjson-properties).

I.e.

```json
"dependsOn": {
  "ghcr.io/user/repo/go": {},
  "ghcr.io/user/repo1/go:1": {},
  "ghcr.io/user/repo2/go:latest": {},
  "https://github.com/user/repo/releases/devcontainer-feature-go.tgz": { 
        "optionA": "value" 
  },
  "./myGoFeature": { 
        "optionA": true,
        "optionB": "hello",
        "version" : "1.0.0"
  }
}
```

Chuxel:
[...]

Issue #139, opened 2022-08-24
`devcontainer features test` improvements


joshspicer:
An issue collecting various improvements that can be made for the `features test...` command.  The command is still experimental and under active development.

- [ ] `remoteUser` is failing with a permissions issue when included in a scenario
- [ ] Document all command functionality 
- [ ] If there is a scenario.json file and no test.sh, the test will not run unless `--skip-autogenerated` is set.  We should be smarter about this
- [ ] The one positional argument (path to collection) is required and doesn't use the default value (cwd)

eitsupi:
Hello.
I developed a feature using the feature template and noticed that the test for option selection was not covered.

Is it possible to run multiple installations by referencing a pre-created options matrix (json file) when running an installation as a test?
Or the user may want to give options as CLI arguments to parallelize using the matrix on GitHub Actions.

Similarly, it would be useful to be able to specify the platform (CPU architecture).

Clockwork-Muse:
Can we add another task?  Currently, there isn't a way to supply `--docker-path`/`--docker-compose-path` to any of the features commands, which causes issues if you're using podman in ci scenarios.

joshspicer:
@Clockwork-Muse  Do you mean a way to specify the path to the docker/docker-compose CLIs? Yes, I'll convert your request into an issue 👍  (If you're interested, we'd also accept a PR on this)

Clockwork-Muse:
Yes, that's what I meant.

jcbhmr:
> If there is a scenario.json file and no test.sh, the test will not run unless --skip-autogenerated is set. We should be smarter about this

👍 id very much like to see this so that i can do this in something like gh actions:
```sh
devcontainer features test -f ${{ matrix.id }}
```
and NOT fail like this:
![image](https://github.com/devcontainers/spec/assets/61068799/967fef0b-2470-4ace-b754-f3c9723155e2)

as a workaround I'm currently doing this:
```sh
if [[ -e test/${{ matrix.id }}/test.sh ]]; then
  devcontainer features test -f ${{ matrix.id }}
else
  devcontainer features test -f ${{ matrix.id }} --skip-autogenerated
fi
```
accidentally made https://github.com/devcontainers/spec/issues/294 🙄

PR #220, opened 2022-10-11
allow referencing external Features in `devcontainer features test` command

joshspicer:
closes https://github.com/devcontainers/cli/issues/217
related: https://github.com/devcontainers/cli/issues/163
tracking: https://github.com/devcontainers/cli/issues/139

Does not replace external Features in a Feature's `scenarios.json`.  For example, this lets people test `installsAfter` with an already published Feature.  This PR also adds a test with an external Feature combined with a local Feature.

Also adds two tests for `installsAfter` - one that must install `A -> B` and one must install `B -> A`.


eitsupi:
This is a long-awaited feature in using the feature template!
(For now, it is not possible to run tests in combination with external features like devcontainers/features)

PR #223, opened 2022-10-11
Handle old and otherwise started containers (microsoft/vscode-remote-release#7307)

chrmarti:


samruddhikhandale (approved):
Looks good. 👍

PR #225, opened 2022-10-12
Wire up fetching a Template from an OCI registry

joshspicer:
Use the existing OCI implementation to fetch Templates.

### DIFF
--- a/src/spec-configuration/containerTemplatesOCI.ts
+++ b/src/spec-configuration/containerTemplatesOCI.ts
@@ -20,9 +20,9 @@ export async function fetchTemplate(output: Log, identifier: string, templateCac
 	output.write(`blob url: ${blobUrl}`, LogLevel.Trace);
 
 	const tmpDir = os.tmpdir();
-	const success = await getBlob(output, process.env, blobUrl, tmpDir, templateCachePath, templateRef);
+	const files = await getBlob(output, process.env, blobUrl, tmpDir, templateCachePath, templateRef, undefined, ['devcontainer-template.json', 'README.md']);
 
-	if (!success) {
+	if (!files) {
 		throw new Error(`Failed to download package for ${templateRef.resource}`);
 	}
 
--- a/src/spec-configuration/containerTemplatesOCI.ts
+++ b/src/spec-configuration/containerTemplatesOCI.ts
@@ -45,6 +45,8 @@ export async function fetchTemplate(output: Log, selectedTemplate: SelectedTempl
 			const fileContents = await readLocalFile(filePath);
 			const fileContentsReplaced = replaceTemplatedValues(output, fileContents.toString(), options);
 			await writeLocalFile(filePath, Buffer.from(fileContentsReplaced));
+		} else {
+			output.write(`Could not find templated file '${f}'.`, LogLevel.Error);
 		}
 	}
 
--- a/src/spec-configuration/containerTemplatesOCI.ts
+++ b/src/spec-configuration/containerTemplatesOCI.ts
@@ -1,0 +1,36 @@
+import { Log, LogLevel } from '../spec-utils/log';
+import * as os from 'os';
+import { fetchOCIManifestIfExists, getBlob, getRef, OCIManifest } from './containerCollectionsOCI';
+
+
+export async function fetchTemplate(output: Log, identifier: string, templateCachePath: string) {
+	const templateRef = getRef(output, identifier);
+	if (!templateRef) {
+		output.write(`Failed to parse template ref for ${identifier}`, LogLevel.Error);
+		return false;
+	}
+
+	const ociManifest = await fetchOCITemplateManifestIfExistsFromUserIdentifier(output, process.env, identifier);
+	if (!ociManifest) {
+		output.write(`Failed to fetch template manifest for ${identifier}`, LogLevel.Error);
+		return false;
+	}
+
+	const blobUrl = `https://${templateRef.registry}/v2/${templateRef.path}/blobs/${ociManifest?.layers[0].digest}`;
+	output.write(`blob url: ${blobUrl}`, LogLevel.Trace);
+
+	const tmpDir = os.tmpdir();
+	const success = await getBlob(output, process.env, blobUrl, tmpDir, templateCachePath, templateRef);
+
+	if (!success) {
+		throw new Error(`Failed to download package for ${templateRef.resource}`);
+	}
+
+	return true;
+}
+
+
+async function fetchOCITemplateManifestIfExistsFromUserIdentifier(output: Log, env: NodeJS.ProcessEnv, identifier: string, manifestDigest?: string, authToken?: string): Promise<OCIManifest | undefined> {
+	const templateRef = getRef(output, identifier);
+	return await fetchOCIManifestIfExists(output, env, templateRef, manifestDigest, authToken);
+}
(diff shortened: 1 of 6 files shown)


src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -0,0 +1,36 @@
...
+	const ociManifest = await fetchOCITemplateManifestIfExistsFromUserIdentifier(output, process.env, identifier);
+	if (!ociManifest) {
+		output.write(`Failed to fetch template manifest for ${identifier}`, LogLevel.Error);
+		return;
+	}
+
+	const blobUrl = `https://${templateRef.registry}/v2/${templateRef.path}/blobs/${ociManifest?.layers[0].digest}`;
+	output.write(`blob url: ${blobUrl}`, LogLevel.Trace);
+
+	const tmpDir = os.tmpdir();
```
chrmarti, line 22: Add a unique folder name to it to avoid collisions?

chrmarti (dismissed):
LGTM, left one comment.

src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -2,31 +2,52 @@ import { Log, LogLevel } from '../spec-utils/log';
...
 
+export interface TemplateOptions {
+	[name: string]: string;
+}
 
-export async function fetchTemplate(output: Log, identifier: string, templateCachePath: string): Promise<string[] | undefined> {
-	const templateRef = getRef(output, identifier);
+export interface SelectedTemplate {
+	id: string;
+	options: TemplateOptions;
```
samruddhikhandale, line 13: Should this be an `options[]` ? 
A template can have multiple options.

src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -2,31 +2,52 @@ import { Log, LogLevel } from '../spec-utils/log';
...
 	const tmpDir = path.join(os.tmpdir(), 'vsch-template-temp', `${Date.now()}`);
-	const files = await getBlob(output, process.env, blobUrl, tmpDir, templateCachePath, templateRef, undefined, ['devcontainer-template.json', 'README.md', 'NOTES.md']);
+	const files = await getBlob(output, process.env, blobUrl, tmpDir, templateDestPath, templateRef, undefined, ['devcontainer-template.json', 'README.md', 'NOTES.md']);
 
 	if (!files) {
 		throw new Error(`Failed to download package for ${templateRef.resource}`);
 	}
 
+	// Scan all template files and replace any templated values.
+	for (const f of files) {
```
samruddhikhandale, line 41: Would this scan and replace recursively? (ie through the sub-folders)

src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -2,31 +2,52 @@ import { Log, LogLevel } from '../spec-utils/log';
...
+		output.write(`Scanning file '${f}'`, LogLevel.Trace);
+		const filePath = path.join(templateDestPath, f);
+		if (await isLocalFile(filePath)) {
+			const fileContents = await readLocalFile(filePath);
+			const fileContentsReplaced = replaceTemplatedValues(output, fileContents.toString(), options);
```
samruddhikhandale, line 46: Should we loop through the `options` ?

src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -35,3 +56,11 @@ async function fetchOCITemplateManifestIfExistsFromUserIdentifier(output: Log, e
 	const templateRef = getRef(output, identifier);
 	return await fetchOCIManifestIfExists(output, env, templateRef, manifestDigest, authToken);
 }
+
+function replaceTemplatedValues(output: Log, template: string, options: TemplateOptions) {
+	const pattern = /\${templateOption:\s*(\w+?)\s*}/g; // ${templateOption:XXXX}
+	return template.replace(pattern, (_, token) => {
+		output.write(`Replacing ${token} with ${options[token]}`);
+		return options[token] || '';
```
samruddhikhandale, line 64: The options's value may contain escape chars (eg - `nodeVersion: "lts/*"`).
Would this function handle that or the should the options be passed with escaped value?

samruddhikhandale (commented):
left few comments/questions!

src/spec-configuration/containerTemplatesOCI.ts
samruddhikhandale: Never mind. Understood why don't need `options[]`!

src/spec-configuration/containerTemplatesOCI.ts
joshspicer: Yes it does (anaconda test proves this)

src/spec-configuration/containerTemplatesOCI.ts
joshspicer: I think this regex is ok? I've updated the test to add in `lts/*` to validate your example.  Is there another example you can think of that would be good to test?

src/test/container-templates/containerTemplatesOCI.test.ts
```diff
@@ -0,0 +1,55 @@
...
+		const dest = path.relative(process.cwd(), path.join(__dirname, 'tmp2'));
+		const files = await fetchTemplate(output, selectedTemplate, dest);
+		assert.ok(files);
+		// Expected:
+		// ./environment.yml, ./.devcontainer/.env, ./.devcontainer/Dockerfile, ./.devcontainer/devcontainer.json, ./.devcontainer/docker-compose.yml, ./.devcontainer/noop.txt
+		assert.strictEqual(files.length, 6);
+
+		// Read File
+		const file = (await readLocalFile(path.join(dest, '.devcontainer', 'Dockerfile'))).toString();
+		assert.match(file, /ARG NODE_VERSION="lts\/\*"/);
```
samruddhikhandale, line 52: Should this be ⬇️  instead?

```suggestion
		assert.match(file, /ARG NODE_VERSION="lts/*"/);
```

I was mentioning that we probably need to escape chars when we are passing this option.value to the replacement engine. However, the final files should have normal values (ie `lts/*`) ?

Because I am not sure how the dockerfile might parse this?

image

For tests action [here](https://github.com/devcontainers/templates/blob/main/.github/workflows/test-pr.yaml#L94), we pass escaped value to `sed` but it ends up replacing without the esc chars - which seems correct to me

joshspicer: the second argument to `match` above is itself a regex. 

Here are the CLI trace logs substituting this value:
```
2022-10-13T20:10:12.208Z] Scanning file './.devcontainer/.env'
[2022-10-13T20:10:12.209Z] Scanning file './.devcontainer/Dockerfile'
[2022-10-13T20:10:12.209Z] Replacing nodeVersion with lts/*
[2022-10-13T20:10:12.210Z] Scanning file './.devcontainer/devcontainer.json'
[2022-10-13T20:10:12.211Z] Scanning file './.devcontainer/docker-compose.yml'
[2022-10-13T20:10:12.212Z] Scanning file './.devcontainer/noop.txt'
```

And the output of the this test looks good to me:
image

samruddhikhandale: Awesome! 🎉

src/spec-configuration/containerTemplatesOCI.ts
samruddhikhandale: looks good. thanks!

samruddhikhandale (approved):
looks great 🎉

PR #228, opened 2022-10-14
User env variables for features

chrmarti:
devcontainers/spec#91

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -258,6 +258,7 @@ echo '${optionsIndented}'
 echo ===========================================================================
 
 set -a
+. ../devcontainer-features.builtin.env
 . ./devcontainer-features.env
 set +a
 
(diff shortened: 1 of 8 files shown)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -274,8 +275,12 @@ function escapeQuotesForShell(input: string) {
 	return input.replace(new RegExp(`'`, 'g'), `'\\''`);
 }
 
-export function getFeatureLayers(featuresConfig: FeaturesConfig) {
-	let result = '';
+export function getFeatureLayers(featuresConfig: FeaturesConfig, containerUser: string, remoteUser: string) {
+	let result = `RUN \\
+echo "_CONTAINER_USER_HOME=$(getent passwd ${containerUser} | cut -d: -f6)" >> /tmp/build-features/devcontainer-features.builtin.env && \\
```
joshspicer, line 280: I wasn't familiar with `getent`. This is it working successfully on an `alpine` container 👍 

image

joshspicer (approved):
👍

src/spec-configuration/containerFeaturesConfiguration.ts
chrmarti: Did some testing and it looks like it is working across distributions.

PR #233, opened 2022-10-17
Normalize feature permissions in an intetermediate stage before copying them to the final container

### LINKED ISSUE #153: "cacheFrom" value not considered when building dev container features
The `build.cacheFrom` value allows you to publish an image, and then use it as a local cache for subsequent builds on another machine (as does the Docker Compose equivalent).  Currently contents of the image in `cacheFrom` do not appear to be being used during the dev container features build step which reduces the utility of it pretty significantly.

Repro:

1. Go to https://github.com/chuxel/feature-library and create a Codespace, or clone the project locally and create a dev container from it
2. Watch the Dockerfile build

**Expected:** Since there is a pre-built image with the "docker-in-docker" feature inside it (see https://github.com/Chuxel/feature-library/blob/main/.github/workflows/devcontainer-image.yaml), the feature docker build reuses the cached result.
**Actual:** The image contents are only used in during the core Dockerfile build not during the feature build

davidwallacejackson:
Resolves #153.

Since the underlying problem was that the permissions for features sources became corrupted and invalidated the cache, I added a new build stage that serves only to `COPY` the features sources and `chmod` them to `0600`. This way, by the time the actual work of installing the features is performed the permissions will have been normalized, allowing subsequent runs to use the install layers as cache.

I tested this manually by

1. Building a devcontainer image with this version of the CLI and pushing it to Docker Hub
2. Clearing the buildkit cache (`docker buildx prune -af`)
3. `docker pull`ing the built image back onto my machine from Docker Hub
4. Building the same image with the **release version** of the CLI and a `--cache-from` on the image from Docker Hub, to validate that the cache is not used by the release version. I canceled this build before it could finish, once it was clear the cache was not being used.
5. Building the image with a `--cache-from` again, but using this version of the CLI (the cache was used successfully)

This might be worth adding to tests, but it'd be a bit awkward -- you'd have to basically recreate the process above and then run regexes on the output to see if you got a cache hit. I can take a crack at that, if that's a blocker for merging.

Side note: in case anybody's curious why I used an additional stage instead of simply adding a `RUN chmod -R 0600 /tmp/build-features` after the `COPY`, it's because you don't get the cache without the extra stage -- BuildKit is apparently smart enough to identify when two contexts are equal, but doesn't seem to have the same capability at the individual layer level. [This demo](https://github.com/davidwallacejackson/docker-cache-merge-test) shows this in action.

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -205,11 +205,15 @@ export function getContainerFeaturesBaseDockerFile() {
 
 #{nonBuildKitFeatureContentFallback}
 
+FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
+COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
+RUN chmod -R 0600 /tmp/build-features
+
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_target_stage
 
 USER root
 
-COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
+COPY --from=dev_containers_feature_content_normalize /tmp/build-features /tmp/build-features
 
 #{featureLayer}
 
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -207,7 +207,7 @@ export function getContainerFeaturesBaseDockerFile() {
 
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
 COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
-RUN chmod -R 0600 /tmp/build-features
+RUN chmod -R 0700 /tmp/build-features
 
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_target_stage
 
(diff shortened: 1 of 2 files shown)


davidwallacejackson:
@microsoft-github-policy-service agree [company="Weights and Biases"]

davidwallacejackson:
@microsoft-github-policy-service agree company="Weights and Biases"

edgonmsft (approved):
Hey! Sorry for the delay but it does look ok to me!

davidwallacejackson:
> Hey! Sorry for the delay but it does look ok to me!

Thank you! Would you mind merging it for me?

Chuxel:
@chrmarti wanted to take a quick look now that he is back.  This look good Christof?

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -205,11 +205,15 @@ export function getContainerFeaturesBaseDockerFile() {
 
 #{nonBuildKitFeatureContentFallback}
 
+FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
+COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
+RUN chmod -R 0600 /tmp/build-features
```
chrmarti, line 210: Would this break features that come with additional executable files (beyond install.sh) because these won't be executable anymore? A simple fix could be to use `700`.

Another approach could be to copy the tar files to the build stage and untar there. That would preserve the original permissions.

chrmarti (changes requested):
Thanks @davidwallacejackson! This looks very promising. I have left a comment for discussion.

src/spec-configuration/containerFeaturesConfiguration.ts
davidwallacejackson: I think you're right. That's my bad; I didn't realize features were allowed to contain additional files. I've changed the `chmod` to `700`, since that's easiest and (I think?) harmless, but I can see how untaring inside docker would be more correct. If you want me to take a crack at that, let me know -- your call.

chrmarti: We can go with this and look into copying the tars separately.

chrmarti (approved):
Thanks!

edgonmsft (approved):

PR #242, opened 2022-10-24
${devcontainerId} (devcontainers/spec#62)

chrmarti:


src/test/container-templates/templatesCLICommands.test.ts
```diff
@@ -43,9 +43,9 @@ describe('tests apply command', async function () {
 
 		assert.isTrue(success);
 		assert.isDefined(result);
-		assert.isTrue(result.stdout === '{"files":["./.devcontainer.json"]}\n');
+		assert.strictEqual(result.stdout.trim(), '{"files":["./.devcontainer/devcontainer.json"]}');
 
-		const file = (await readLocalFile(path.join(tmp, 'template-output', '.devcontainer.json'))).toString();
+		const file = (await readLocalFile(path.join(tmp, 'template-output', '.devcontainer', 'devcontainer.json'))).toString();
```
samruddhikhandale, line 48: Thanks for fixing this! 👍

src/spec-common/variableSubstitution.ts
```diff
@@ -32,6 +33,11 @@ export function substitute<T extends object>(context: SubstitutionContext, value
 	return substitute0(replace, value);
 }
 
+export function beforeContainerSubstitute<T extends object>(idLabels: Record<string, string>, value: T): T {
+	let devcontainerId: string | undefined;
+	return substitute0(replaceDevContainerId.bind(undefined, () => devcontainerId || (devcontainerId = devcontainerIdForLabels(idLabels))), value);
```
samruddhikhandale, line 38: Curious, would both code result into same output ? 

```suggestion
	return substitute0(replaceDevContainerId.bind(undefined, () => devcontainerId || devcontainerIdForLabels(idLabels))), value);
```

samruddhikhandale (approved):
Looks great ✅ 
Left one question

src/spec-common/variableSubstitution.ts
chrmarti: The output would be the same, but the variable `devcontainerId` wouldn't be used for caching because it would not get the id assigned when it is computed the first time. Performance would be worse when `${devcontainerId}` is used multiple times.

OATEF4313 (commented):
> 
[DO-IRPV3.0--2022-06-30.pdf](https://github.com/user-attachments/files/17171538/DO-IRPV3.0--2022-06-30.pdf)
[trialagreement.pdf](https://github.com/user-attachments/files/17171540/trialagreement.pdf)
[export.csv](https://github.com/user-attachments/files/17171541/export.csv)
[Comparison Sheet (2).pdf](https://github.com/user-attachments/files/17171542/Comparison.Sheet.2.pdf)
[Comparison Sheet (1).pdf](https://github.com/user-attachments/files/17171543/Comparison.Sheet.1.pdf)
`
[Comparison Sheet.pdf](https://github.com/user-attachments/files/17171544/Comparison.Sheet.pdf)
[documents-in-tsv.csv](https://github.com/user-attachments/files/17171545/documents-in-tsv.csv)
[pages-in-tsv.csv](https://github.com/user-attachments/files/17171546/pages-in-tsv.csv)
[rfc4844.txt.pdf](https://github.com/user-attachments/files/17171547/rfc4844.txt.pdf)

PR #243, opened 2022-10-24
Update examples to use mergedConfiguration so image labels work

Chuxel:
Takes advantage of the new mergedConfiguration section in config output to simplify the script and pick up image label info.

Also made a minor tweak to properly return the right exit code from the CI example.

### DIFF
--- a/example-usage/ci-app-build-script/build-application.sh
+++ b/example-usage/ci-app-build-script/build-application.sh
@@ -2,7 +2,6 @@
 set -e
 cd "$(dirname $0)"
 
-
 build_date="$(date +%s)"
 
 # Create a label for use during cleanup since the devcontainer CLI does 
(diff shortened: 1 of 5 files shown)


Chuxel:
The CI failure reported here does not seem to be related to this change.

bamurtaugh (commented):
Thanks for making these changes!

It's been a bit since I've run the examples, so perhaps you can help me refresh my memory on what/how I should be checking. Here's what I tried:

- Open devcontainers/cli on the example-tweaks branch locally
- `yarn` and `yarn compile` to ensure CLI is working properly
- Type `example-usage/tool-vscode-server/start.sh` in integrated terminal
- When I visit http://localhost:8000/, it can't be reached

Chuxel:
> * example-usage/tool-vscode-server/start.sh

I use the deployed version of the CLI for testing.

```
npm install -g @devcontainers/cli
```

Possible there is a bug with local install - or it's finding an out-of-date version of the CLI.  Does that work?  If not, are there any errors?

bamurtaugh (approved):
Thanks for the guidance, got it working now and LGTM!

PR #247, opened 2022-10-25
Add missing substitutions in run-user-commands (microsoft/vscode-remote-release#7412)

chrmarti:


edgonmsft (approved):
Can you add a description on what the issue was?

PR #248, opened 2022-10-25
Last value wins (microsoft/vscode-remote-release#7390)

chrmarti:


edgonmsft (approved):
LGTM, can you add a link to the bug and a description on what the issue was?

PR #254, opened 2022-10-31
Apply current configuration (#252)

### LINKED ISSUE #252: Update values in `devcontainer.json` are not respected after the container is created
Version 0.74.0

Updated values, such as `remoteEnv` in `devcontainer.json` after the container is created are not respected since the values from the container metadata are taking precedence.

This breaks flows where this values are updated, like secrets, to keep environments working.

The issue appears to be coming from here: https://github.com/devcontainers/cli/blob/dadf77bf0d0f7047fc935f4b56ceeda463860295/src/spec-node/imageMetadata.ts#L119-L149

chrmarti:
Fixes #252

### DIFF
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -39,6 +39,12 @@ const pickConfigProperties: (keyof DevContainerConfig & keyof ImageMetadataEntry
 	'hostRequirements',
 ];
 
+const pickUpdateableConfigProperties: (keyof DevContainerConfig & keyof ImageMetadataEntry)[] = [
+	'remoteUser',
+	'userEnvProbe',
+	'remoteEnv',
+];
+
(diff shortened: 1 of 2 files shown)


src/spec-node/imageMetadata.ts
```diff
@@ -39,6 +39,12 @@ const pickConfigProperties: (keyof DevContainerConfig & keyof ImageMetadataEntry
 	'hostRequirements',
 ];
 
+const pickUpdateableConfigProperties: (keyof DevContainerConfig & keyof ImageMetadataEntry)[] = [
```
samruddhikhandale, line 42:
```suggestion
const pickUpdatableConfigProperties: (keyof DevContainerConfig & keyof ImageMetadataEntry)[] = [
```

nit: spelling

samruddhikhandale (approved):
Left a comment. Besides that, looks great.

PR #258, opened 2022-11-01
Skip features code path without features (microsoft/vscode-remote-release#7305)

chrmarti:
microsoft/vscode-remote-release#7305


samruddhikhandale (approved):
🚀

PR #265, opened 2022-11-02
`features test...` command: Copy entire test folder on test execution and improve CLI command usage.

joshspicer:
Tracking: https://github.com/devcontainers/cli/issues/139

Some changes/improvements to the `devcontainers features test` command.

- Deprecates the required positional <target> in favor of a flag.  This fixes the bug with ordering options lets the command be run parameterless for all defaults
- Now copies entire `/test` folder over to each test run, so tests (scenarios, etc..) can share scripts between runs
- Avoids outputting test command-related output to standard error.


src/test/container-features/featuresCLICommands.test.ts
```diff
@@ -49,6 +49,31 @@ describe('CLI features subcommands', async function () {
...
+			assert.isDefined(result);
+
+			const expectedTestReport = `  ================== TEST REPORT ==================
+✅ Passed:      'util'
+✅ Passed:      'some_scenario'`;
+			const hasExpectedTestReport = result.stdout.includes(expectedTestReport);
+			assert.isTrue(hasExpectedTestReport);
+
+			assert.isTrue(result.stderr.includes('I AM A DIFFERENT SCRIPT'));
+			assert.isTrue(result.stderr.includes('I AM A HELPER SCRIPT FOR A SCENARIO'));
```
AlexanderLanin, line 74: Isn't this check redundant since you have `set -e` in the test scripts?

joshspicer: I agree it's a bit redundant. 

For a unit test's completeness I want to be 100% sure the test continues to function how i'd expect (if someone removes `set -e` in a later commit I would want changes to get caught)

src/spec-node/featuresCLI/test.ts
```diff
@@ -10,7 +10,8 @@ import { doFeaturesTestCommand } from './testCommandImpl';
 export function featuresTestOptions(y: Argv) {
 	return y
 		.options({
-			'features': { type: 'array', alias: 'f', describe: 'Feature(s) to test as space-separated parameters. Omit to run all tests.  Cannot be combined with \'--global-scenarios-only\'.', },
+			'target-project': { type: 'string', alias: 't', default: '.', description: 'Path to folder containing \'src\' and \'test\' sub-folders.' },
```
chrmarti, line 13: `--workspace-folder` or `--project-folder` might make it easier to read since it is more obvious that it is about a folder (and not a name or some other kind of reference).

chrmarti (approved):
LGTM, left a suggestion.

src/spec-node/featuresCLI/test.ts
joshspicer: I didn't want to conflate with the concept of `workspace-folder`, but I agree that `project-folder` is more clear - will change!

PR #267, opened 2022-11-04
Avoid image build (microsoft/vscode-remote-release#7378)

chrmarti:
Fixes microsoft/vscode-remote-release#7378.

### DIFF
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -15,7 +15,7 @@ import { readLocalFile } from '../spec-utils/pfs';
 import { includeAllConfiguredFeatures } from '../spec-utils/product';
 import { createFeaturesTempFolder, DockerResolverParameters, getCacheFolder, getFolderImageName, getEmptyContextFolder, SubstitutedConfig } from './utils';
 import { isEarlierVersion, parseVersion } from '../spec-common/commonUtils';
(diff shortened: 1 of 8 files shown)


src/spec-node/containerFeatures.ts
```diff
@@ -105,6 +106,13 @@ export async function getExtendImageBuildInfo(params: DockerResolverParameters,
 	// Processes the user's configuration.
 	const featuresConfig = await generateFeaturesConfig(params.common, dstFolder, config.config, getContainerFeaturesFolder, additionalFeatures);
 	if (!featuresConfig) {
+		if (canAddLabelsToContainer && !imageBuildInfo.dockerfile) {
```
joshspicer, line 109: I don't really understand `canAddLabelsToContainer`?  Is this strictly for compose?

joshspicer (approved):
This LGTM, mostly based on walking through the test.  A little more detail on the approach taken here (in the PR description or commit message, maybe) would be interesting to hear.

chrmarti:
The previous approach was to always use `LABEL` in an additional (or modified) Dockerfile to add the metadata. The alternative to this is to use `--label` (or Docker Compose's equivalent) when starting the container. The advantage of `LABEL` is that it has a higher upper limit on the size of the label than `--label` (1.3MB vs. 0.5MB based on testing, IIRC).

The fix is to use `--label` when adding metadata is the only reason for using an additional (or modified) Dockerfile. For `devcontainer build ...` we still want to have a new image with the metadata baked in, this is when `canAddLabelsToContainer` is passed in as `false`.

Hope that makes sense, let me know if anything needs more explanation. Thanks!

Issue #279, opened 2022-11-11
Bundle code for npm package


chrmarti:
We use the npm package as a CLI that is ready to run (as opposed to a source dependency). We will minify and bundle the code to optimize for that usage.

PR #292, opened 2022-11-21
Keep dependencies for 3rd-party notices generator (devcontainers/cli#279)

chrmarti:
We moved all dependencies but the node-pty native module to devdependencies with the change for bundling (so these don't get installed again when installing from NPM). This also removes them from the 3rd-party notices. This change reverts that part of the change and instead modifies the package.json only when producing the package for NPM.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ logs
 *.tgz
 tmp
 tmp[0-9]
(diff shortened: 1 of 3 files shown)


edgonmsft (approved):
LGTM

PR #293, opened 2022-11-21
downcase OCI identifiers and validate input of getRef()

joshspicer:
Downcases the identifier input string (context: https://github.com/oras-project/oras/issues/679)

Additionally, be more strict about what the Feature/Template OCI identifiers that we parse.  `getRef()` now returned `undefined` if the parsed components don't match [the regular expressions provided in the OCI distribution spec]( https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests).

Adds new tests to assert this behavior.

NOTE: This isn't perfect yet, but an improvement. Left two `TODO`s inline.

relevant: 
- https://github.com/oras-project/oras/issues/679
- https://github.slack.com/archives/C0431R35H37/p1669045129224189
- https://github.com/devcontainers/action/pull/105
- https://github.com/devcontainers/action/issues/103

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -11,6 +11,18 @@ export const DEVCONTAINER_MANIFEST_MEDIATYPE = 'application/vnd.devcontainers';
 export const DEVCONTAINER_TAR_LAYER_MEDIATYPE = 'application/vnd.devcontainers.layer.v1+tar';
 export const DEVCONTAINER_COLLECTION_LAYER_MEDIATYPE = 'application/vnd.devcontainers.collection.layer.v1+json';
 
+// Following Spec:   https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
+// Alternative Spec: https://docs.docker.com/registry/spec/api/#overview
+//
+// Entire path ('namespace' in spec terminology) for the given repository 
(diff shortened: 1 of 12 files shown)


samruddhikhandale (dismissed):

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -58,13 +58,38 @@ interface OCITagList {
...
-	// ex: ghcr.io/codspace/templates/ruby:1
-	const splitOnColon = resourceAndVersion.split(':');
-	const resource = splitOnColon[0];
-	const version = splitOnColon[1] ? splitOnColon[1] : 'latest';
+// Following Spec:   https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
+// Alternative Spec: https://docs.docker.com/registry/spec/api/#overview
+//
+// Entire path ('namespace' in spec terminology) for the given repository 
+// (eg: devcontainers/features/go)
+const regexForPath = /[a-z0-9]+([._-][a-z0-9]+)*(\/[a-z0-9]+([._-][a-z0-9]+)*)*/;
```
chrmarti, line 66: Add `^` and `$` to the two regexes to make sure they match the entire string and not just a substring?

chrmarti (changes requested):

chrmarti (approved):
LGTM!

PR #298, opened 2022-11-23
Emit response from registry on failed `postUploadSessionId`

joshspicer:
relevant: https://github.com/devcontainers/action/issues/103

This change stops throwing away the response body when using `requestResolveHeaders()`, and utilizes the response body to display any error message an OCI registry may send us on error (per: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes).  

This patch doesn't parse the (presumably) JSON response, just emits the entire string as an error log.

This output will be useful in cases like this example:
https://github.com/warrenbuckley/codespace-features/actions/runs/3515866436/jobs/5891670553#step:3:22


chrmarti (approved):
LGTM!

PR #305, opened 2022-11-30
emit a JSON summary of the result of the `features publish` command

joshspicer:
setup work for https://github.com/github/codespaces/issues/11247

Emits a JSON object to standard out at the end of a successful `devcontainer features publish` run.  Each Feature is listed and reports the tag(s) that were updated as a result of this publish.  

An empty array `[]` represents publishing being skipped for that Feature (for example if the Feature version was not updated)

image

### DIFF
--- a/src/spec-node/featuresCLI/publish.ts
+++ b/src/spec-node/featuresCLI/publish.ts
@@ -84,7 +84,7 @@ async function featuresPublish({
             process.exit(1);
         }
 
-        const publishResult = await doPublishCommand(f.version, featureRef, outputDir, output, collectionType)
+        const publishResult = await doPublishCommand(f.version, featureRef, outputDir, output, collectionType);
         if (!publishResult) {
             output.write(`(!) ERR: Failed to publish '${resource}'`, LogLevel.Error);
             process.exit(1);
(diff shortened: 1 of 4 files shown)


src/spec-node/featuresCLI/publish.ts
```diff
@@ -82,10 +84,16 @@ async function featuresPublish({
             process.exit(1);
         }
 
-        if (! await doPublishCommand(f.version, featureRef, outputDir, output, collectionType)) {
+        const publishResult = await doPublishCommand(f.version, featureRef, outputDir, output, collectionType);
```
samruddhikhandale, line 87: As `templates publish command` also uses the same `doPublishCommand()`. Wondering, if you'd also like to output the result for Templates as well?

You'd likely need to make similar changes like this file to [src/spec-node/templatesCLI/publish.ts](https://github.com/devcontainers/cli/blob/main/src/spec-node/templatesCLI/publish.ts)

samruddhikhandale (approved):
🔥 🔥  Approving with a suggestion.

src/spec-node/featuresCLI/publish.ts
joshspicer: Good point, might as well make those two consistent 👍

chrmarti (approved):

src/spec-node/featuresCLI/publish.ts
joshspicer: image

samruddhikhandale: 🎉

samruddhikhandale (approved):
🚀

PR #310, opened 2022-12-02
A more spec-compliant/resilient OCI distribution implementation

joshspicer:
Exercises the `devcontainer features publish` command against the [OCI registry specification's Reference Implementation ](https://hub.docker.com/_/registry), and updates our implementation to more closely follow the specification. 

More specifically, starts a `registry` container (with hardcoded auth generated with htpasswd).

```js
const startRegistryCmd = `docker run -d -p 5000:5000 \
   -v ${resolvedTmpPath}/auth.htpasswd:/etc/docker/registry/auth.htpasswd \
   -e REGISTRY_AUTH="{htpasswd: {realm: localhost, path: /etc/docker/registry/auth.htpasswd}}" \
   --name registry \
   registry`;
```

And then executing the following `publish` command against the registry server running on localhost.

```bash
DEVCONTAINERS_OCI_AUTH='localhost:5000|myuser|mypass' ./${cli} features publish --log-level trace -r localhost:5000 -n octocat/features ${collectionFolder}/src
```

A new class of unit tests will spin up an ephemeral `registry` container and exercise the push and pull operations of our OCI distribution implementation.

### Details on changes made to support Reference Implementation

- Create a new `postUploadSessionId` for each uploaded blob layer.  Previously, a single session ID was created for each manifest, but our implementation now matches the [specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#post-then-put) where it should be generated per blob.  GHCR tolerates re-using the upload ID, but the reference implementation does not.
- Fallback to `Basic` auth if a session token cannot be fetched from the registry.  For GHCR, operations are only accepted with a valid session token from the `/session` endpoint of the registry, whereas the reference implementation talks only with Basic auth credentials.  I don't see the session auth built in the official specification, so my implementation supports both methods, only using Basic auth if the /token endpoint does not return a success code and a valid token from it's JSON response body. 
- Accurately set  the `Content-Length` header when uploading blobs, per the specification. 

#### Additional Changes
- Previously, we were reading the `.tgz` files from disk several times throughout the process. Now, the file is read from disk into a buffer one time, and the buffer is passed around.  This is more efficient.
- The `DEVCONTAINERS_OCI_AUTH` environment variable was always broken (a condition would always result to `false`).  Since there is no way this could be used by any active users (and it is not yet documented), I changed the contract a bit to include a username field, which is necessary to get this variable working with the reference implementation (GHCR ignores this field, which is why it was originally omitted). This variable is exercised in the provided test.
- More clearly surfaces error responses from the registry in logs. Errors are parsed as [JSON](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes) per the specification.
image

- Break out existing `features info` command two new, simple commands `devcontainer info tags` and `devcontainer info manifest`.  These are useful utilities to break out for understanding published Features. They are used in the newly added test.
- The http(s) family of functions will now default to using `http` when communicating with `localhost`, or if the parsed protocol is `http`.
image

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -168,7 +168,8 @@ export function getCollectionRef(output: Log, registry: string, namespace: strin
 export async function fetchOCIManifestIfExists(output: Log, env: NodeJS.ProcessEnv, ref: OCIRef | OCICollectionRef, manifestDigest?: string, authToken?: string): Promise<OCIManifest | undefined> {
 	// Simple mechanism to avoid making a DNS request for 
 	// something that is not a domain name.
-	if (ref.registry.indexOf('.') < 0) {
+	if (ref.registry.indexOf('.') < 0 && !ref.registry.startsWith('localhost')) {
+		output.write(`ERR: Registry '${ref.registry}' is not a valid domain name or IP address.`, LogLevel.Error);
 		return undefined;
 	}
 
@@ -255,6 +256,7 @@ async function getBasicAuthCredential(output: Log, registry: string, env: NodeJS
 
 	let userToken: string | undefined = undefined;
 	if (!!env['GITHUB_TOKEN'] && registry === 'ghcr.io') {
+		output.write('Using environment GITHUB_TOKEN for auth', LogLevel.Trace);
 		userToken = `USERNAME:${env['GITHUB_TOKEN']}`;
 	} else if (!!env['DEVCONTAINERS_OCI_AUTH']) {
 		// eg: DEVCONTAINERS_OCI_AUTH=domain1|user1|token1,domain2|user2|token2
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -226,7 +226,7 @@ async function getBasicAuthCredential(output: Log, registry: string, env: NodeJS
 
 	let userToken: string | undefined = undefined;
 	if (!!env['GITHUB_TOKEN'] && registry === 'ghcr.io') {
-		userToken = env['GITHUB_TOKEN'];
+		userToken = `USERNAME:${env['GITHUB_TOKEN']}`;
 	} else if (!!env['DEVCONTAINERS_OCI_AUTH']) {
 		// eg: DEVCONTAINERS_OCI_AUTH=domain1|user1|token1,domain2|user2|token2
 		const authContexts = env['DEVCONTAINERS_OCI_AUTH'].split(',');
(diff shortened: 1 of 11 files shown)


src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -217,45 +218,94 @@ export async function getManifest(output: Log, env: NodeJS.ProcessEnv, url: stri
...
-export async function fetchRegistryAuthToken(output: Log, registry: string, ociRepoPath: string, env: NodeJS.ProcessEnv, operationScopes: string): Promise<string | undefined> {
-	if (registry === 'mcr.microsoft.com') {
-		return undefined;
+// Exported Function
+// Will attempt to generate/fetch the correct authorization header for subsequent requests (Bearer or Basic)
+export async function fetchAuthorization(output: Log, registry: string, ociRepoPath: string, env: NodeJS.ProcessEnv, operationScopes: string): Promise<string | undefined> {
+	const basicAuthTokenBase64 = await getBasicAuthCredential(output, registry, env);
+	const scopeToken = await generateScopeTokenCredential(output, registry, ociRepoPath, env, operationScopes, basicAuthTokenBase64);
+
+	// Prefer returned a Bearer token retreived from the /token endpoint.
```
samruddhikhandale, line 233: typo! 
```suggestion
	// Prefer returned a Bearer token retrieved from the /token endpoint.
```

src/spec-utils/httpRequest.ts
```diff
@@ -20,11 +20,17 @@ export function request(options: { type: string; url: string; headers: Record<st
 			headers: options.headers,
 			agent: new ProxyAgent(),
 		};
-		const req = https.request(reqOptions, res => {
+
+		const plainHTTP = parsed.protocol === 'http:' || parsed.hostname === 'localhost';
```
samruddhikhandale, line 24: Looks like this part of code to find request (http/https) is repetitive. Suggestion: Can we add a new function?

src/test/container-features/containerFeaturesOCIPush.test.ts
```diff
@@ -1,20 +1,116 @@
...
+import * as path from 'path';
+import * as fs from 'fs';
+import { writeLocalFile } from '../../spec-utils/pfs';
 
+const pkg = require('../../../package.json');
 
 export const output = makeLog(createPlainLog(text => process.stdout.write(text), () => LogLevel.Trace));
 const testAssetsDir = `${__dirname}/assets`;
 
+describe('Test OCI Push against reference registry', async function () {
```
samruddhikhandale, line 16: yay 🎉

samruddhikhandale (approved):
🎉  This is awesome, thanks! ✨

src/spec-utils/httpRequest.ts
joshspicer: I don't think this is too repetitive?  I do think that we may not need three functions here, but the first function throws and that behavior is expected across the codebase, which is why I didn't want to modify that behavior and had made a new function.

chrmarti: 'localhost' could be https for testing, would it make sense to drop that part of the condition?

chrmarti (approved):
Left a question regarding 'localhost' implying 'http'. LGTM otherwise!

src/spec-utils/httpRequest.ts
joshspicer: That's a good point. The main reason I added this in was that we have a lot of handcrafted URLs with `https://....` across the codebase, and at each of those points I'd need to determine if I wanted http or https.  To avoid a lot of redundant changes in this PR, I just look for `localhost`. 

I'll make a note to clean this up but I think it's fine for now?

PR #311, opened 2022-12-02
Add support for "extends" keyword

greggroth:
An "extends" keyword that allows one devcontainer file to import another devcontainer file as a base configuration was proposed in https://github.com/devcontainers/spec/issues/22. This PR aims to implement support for that feature.


src/spec-node/configContainer.ts
```diff
@@ -71,26 +71,41 @@ async function resolveWithLocalFolder(params: DockerResolverParameters, parsedAu
...
+	return updated;
+}
+
+export async function readDevContainerConfigFile(cliHost: CLIHost, workspace: Workspace | undefined, configFile: URI, mountWorkspaceGitRoot: boolean, output: Log, consistency?: BindMountConsistency, overrideConfigFile?: URI) {
+	const confPath = overrideConfigFile ?? configFile;
+	let content = await readFile(cliHost, confPath) as DevContainerConfig;
+
+	if (content.extends) {
+		const extendsConfPath = path.resolve(path.dirname(confPath.path), content.extends);
+		const referencedContent = await readFile(cliHost, URI.file(extendsConfPath)) as DevContainerConfig;
```
chrmarti, line 95: Should we check if there is an `extends` property on the referenced content?

src/spec-node/configContainer.ts
```diff
@@ -71,26 +71,41 @@ async function resolveWithLocalFolder(params: DockerResolverParameters, parsedAu
...
+		delete content.extends;
+		content = deepmerge.deepmerge(referencedContent, content);
```
chrmarti, line 97: We have merge logic specified for the image metadata, should we reuse the same here? https://github.com/devcontainers/spec/blob/main/proposals/image-metadata.md#merge-logic

Plain deep merge might not always have the expected outcome and could result in invalid configuration more easily.

chrmarti (changes requested):

src/spec-node/configContainer.ts
greggroth: As-written, it would be present in the final configuration, but not resolved and it's contents merged in (it's easy enough to delete the property if you don't want it polluting the final configuration). I'm inclined to avoid support for recursive/multiple-extends unless we know there's a strong use-case for it. Detecting loops and self-references will make things far more complicated. Do you think it's important to support that immediately or can we wait?

src/spec-node/configContainer.ts
greggroth: In [the "extends" proposal](https://github.com/devcontainers/spec/issues/22), we discussed a using a simple deep-merge where the behavior is based on the type and there isn't specific per-key behavior like the image metadata merge algorithm requires, similar to how [docker-compose handles adding and overriding configuration](https://docs.docker.com/compose/extends/#adding-and-overriding-configuration). For example, in the image metadata merge algorithm, `hostRequirements` picks the highest while for "extends" using a naive deep-merge algorithm, it would pick the one in the referenced configuration irrespective of whether it's higher or lower. As a user of devcontainer without intimate knowledge of the code, I find that more intuitive and easy to predict. Has the proposed image metadata merge algorithm been implemented yet?

src/spec-node/configContainer.ts
chrmarti: Waiting to assess the importance sounds good.

src/spec-node/configContainer.ts
chrmarti: It is implemented here: https://github.com/devcontainers/cli/blob/ddd743e7f04bbb5f90793b29506931cec65b0922/src/spec-node/imageMetadata.ts

It covers having a devcontainer.json to build a base image that is then reused by a project's devcontainer.json. This might turn the simplicity argument around because it will be simpler for users to deal with a single merge logic than with multiple ones. What do you think? (We might have to add to the current merge logic, but that's fine if it keeps the UX simple/r.)

greggroth: I don't have strong feelings either way aside from ensuring we meet the technical needs of allowing users to override settings and that we do that in as intuitive and user-friendly way as possible. If the image-metadata merge algorithm is solidified and there's an expectation that users would be familiar with it, then it makes sense to be consistent and expect the same thing when using "extends". @Chuxel, I know you had good intuition and feedback when we were working on the proposal. Do you have any feelings either way about using a docker-compose-style merge versus the nuanced image-metadata merge?

greggroth: One consideration I forgot to mention is that whatever merging behavior we adopt, we will more than likely need to duplicate it in GitHub.com in order to properly read the configuration for Codespaces. Has that already been discussed for image metadata support?

Chuxel: Yeah I think we said that compose was the obvious baseline implementation, but this was back before we had the features and image metadata merge logic. Now that we have a more refined view, I agree consistency would probably make sense.  We doc'd the merge logic in the spec as well.

WillPapper:
The `extends` keyword solves a major problem for us! We have a number of services that share a base dev environment, but then have additional needs for per-service configuration on top of it.

Is there a chance that this will be merged soon? Would prefer to not have to work around it by building a separate Docker container for the base dev environment, since that would break `postCreateCommand` behavior that we rely on in it.

(For what it's worth, Copilot thinks that this keyword was already merged in and is suggesting it to end users, which is how I found my way here)

asanglard:
hi @joshspicer @Chuxel @greggroth 

is there any update here ? I'm also looking for such feature.

this will definitively help a lot to allow developers of a same project to make some slights adjustements of the devcontainer.json

fsw0422:
Hi team, any update on this PR? I think this is one of most-wanted feature in devcontainers

ahrakos:
hi @joshspicer @Chuxel @greggroth

Any updates regarding this? will be very helpful for us.

mannok:
What's going on?   Almost 4 years and you guys are not doing anything of this?

PR #324, opened 2022-12-12
Fix escaping of embedded JSON

chrmarti:


src/spec-node/dockerCompose.ts
```diff
@@ -541,7 +541,7 @@ while sleep 1 & wait $$!; do :; done", "-"${userEntrypoint.map(a => `, ${JSON.st
     security_opt:${securityOpts.map(securityOpt => `
       - ${securityOpt}`).join('')}` : ''}${additionalLabels.length ? `
     labels:${additionalLabels.map(label => `
-      - ${label.replace(/\$/g, '$$$$')}`).join('')}` : ''}${mounts.length ? `
+      - '${label.replace(/\$/g, '$$$$').replace(/'/g, '\'\'')}'`).join('')}` : ''}${mounts.length ? `
```
joshspicer, line 544: This isn't missing a closing quote, right? I see a new `'` at the front, but not obvious to me if there is supposed to be an added end quote

chrmarti: There is an additional end quote, but the diff is hard to read. 👍

PR #332, opened 2022-12-20
use current directory as default workspace-folder

### LINKED ISSUE #29: Why Is `--workspace-folder` Required or Default to $PWD?
I found it odd that this parameter was required when using `devcontainer up` and was wondering why and if that was intentional. Not sure if it is also required in other commands ( just learning here ) but did want to share it felt odd to do this.

```shell
$ devcontainer up --workspace-folder .
```

nikaro:
This should probably fix #29.

### DIFF
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -90,7 +90,7 @@ function provisionOptions(y: Argv) {
 		'docker-compose-path': { type: 'string', description: 'Docker Compose CLI path.' },
 		'container-data-folder': { type: 'string', description: 'Container data folder where user data inside the container will be stored.' },
 		'container-system-data-folder': { type: 'string', description: 'Container system data folder where system data inside the container will be stored.' },
-		'workspace-folder': { type: 'string', description: 'Workspace folder path. The devcontainer.json will be looked up relative to this path.' },
+		'workspace-folder': { type: 'string', default: '.', description: 'Workspace folder path. The devcontainer.json will be looked up relative to this path.' },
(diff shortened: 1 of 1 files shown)


src/spec-node/devContainersSpecCLI.ts
```diff
@@ -90,7 +90,7 @@ function provisionOptions(y: Argv) {
 		'docker-compose-path': { type: 'string', description: 'Docker Compose CLI path.' },
 		'container-data-folder': { type: 'string', description: 'Container data folder where user data inside the container will be stored.' },
 		'container-system-data-folder': { type: 'string', description: 'Container system data folder where system data inside the container will be stored.' },
-		'workspace-folder': { type: 'string', description: 'Workspace folder path. The devcontainer.json will be looked up relative to this path.' },
+		'workspace-folder': { type: 'string', default: '.', description: 'Workspace folder path. The devcontainer.json will be looked up relative to this path.' },
```
chrmarti, line 93: We should treat it the same with the other commands. Would it make sense to search for the first parent folder with a `.devcontainer` folder (or a `.devcontainer.json` file)? Similar to what `git` does.

chrmarti (changes requested):
Thanks for the PR! Left a comment.

src/spec-node/devContainersSpecCLI.ts
nikaro: It would make perfect sense, but this is above my TypeScript knowledge to do that.
Feel free to close this PR to let a futur contributor to may be implement this.

chrmarti: Sounds good, closing for now, thanks!

Issue #321, opened 2022-12-09
Failed to fetch local disk feature on Windows


julioes:
On Windows Dev Containers v0.266.1 local features prevent devcontainer from building.

With this folder structure in E:\Projects\test-local-feature
![image](https://user-images.githubusercontent.com/7113/206591738-ce487d2f-e154-417b-867d-6466c52339b6.png)

devcontainer.json
```json
{
	"name": "test local feature",
	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
	"features": {
		"./local-features/test": {}
	}
}
```
devcontainer-feature.json
```json
{
	"id": "test",
	"name": "test",
	"options": {	}
}
```
install.sh
```bash
echo HERE!
```

Trying to open the container fails with
```
[2022-12-09T00:08:27.167Z] * Processing feature: ./local-features/test
[2022-12-09T00:08:27.167Z] Local disk feature.
[2022-12-09T00:08:27.167Z] * Fetching feature: test_1_file-path
[2022-12-09T00:08:27.168Z] (!) ERR: Failed to fetch feature:  
[2022-12-09T00:08:27.168Z] Error: ENOENT: no such file or directory, lstat 'c:\e:\Projects\test-local-feature\.devcontainer\local-features\test'
```

Looking at the code it looks like the path to the local feature configuration is being passed to processFeatureIdentifier with a leading \, i.e. instead of E:\Projects\test-local-feature\.devcontainer\local-features\test it was \E:\Projects\test-local-feature\.devcontainer\local-features\test. When this path is eventually passed to the ncp function to copy the feature files the drive of vscode's working directory c:\ is prepended resulting in a path to a file that doesn't exist.

The vscode-uri object has an fsPath property which its documentation says should be used for file system paths instead of the path property, using that property on this line seems to fix the issue in my extremely limited attempt at testing it.
https://github.com/devcontainers/cli/blob/93cb111e4276d202fca1244fe034cb6d6d60196c/src/spec-configuration/containerFeaturesConfiguration.ts#L552

PR #177, opened 2022-09-14
matrix tests to run on linux and windows

joshspicer:
Closes #35

PR #333, opened 2022-12-21
Fix "Failed to fetch local disk feature on Windows"

joshspicer:
fixes: https://github.com/devcontainers/cli/issues/321

Also adds in a test to prevent regression via a new Windows-based test workflow. 

NOTE: Refactoring/updating the test suite will need to be done incrementally to get 100% of the test suite running (at that point we can remove the new workflow and matrix running the tests against `ubuntu-latest` and `windows-latest` in a single workflow)

`shelljs` is added a dev dependency to abstract shell commands when running on linux/windows (see this closed PR for an example of how it will be used: https://github.com/devcontainers/cli/pull/177)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -547,9 +556,16 @@ function featuresToArray(config: DevContainerConfig, additionalFeatures: Record<
 
 // Process features contained in devcontainer.json
 // Creates one feature set per feature to aid in support of the previous structure.
-async function processUserFeatures(output: Log, config: DevContainerConfig, workspaceRoot: string, env: NodeJS.ProcessEnv, userFeatures: DevContainerFeature[], featuresConfig: FeaturesConfig, skipFeatureAutoMapping: boolean): Promise<FeaturesConfig> {
+async function processUserFeatures(params: ContainerFeatureInternalParams, config: DevContainerConfig, workspaceRoot: string, userFeatures: DevContainerFeature[], featuresConfig: FeaturesConfig): Promise<FeaturesConfig> {
+	const { platform, output, skipFeatureAutoMapping, env, } = params;
+
+	let configPath = config.configFilePath.path;
```
chrmarti, line 562: `config.configFilePath.fsPath` should give you the local path.

chrmarti (changes requested):
Left a comment for simplifying the code. Thanks.

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer: In codespaces, it gives a `vscode-remote://` URI - is that what we want?

package.json
```diff
@@ -84,6 +85,7 @@
 		"npm-run-all": "^4.1.5",
 		"p-all": "^4.0.0",
 		"rimraf": "^3.0.2",
+		"shelljs": "^0.8.5",
```
chrmarti, line 88: Is that needed? Maybe a leftover?

chrmarti (changes requested):

package.json
joshspicer: I am eventually planning on using this for getting the tests to run cross platform (see PR description), but yea not needed yet so I will remove 👍

chrmarti (approved):
👍

PR #335, opened 2022-12-22
Feature: Adds support for renaming with `legacyIds` property

samruddhikhandale:
Spec PR - https://github.com/devcontainers/spec/pull/163
Spec issue - https://github.com/devcontainers/spec/issues/146

Spec PR to update schema - https://github.com/devcontainers/spec/pull/171

Demo 

- Sample Repo: https://github.com/samruddhikhandale/test-legacyIds
- [Commit](https://github.com/samruddhikhandale/test-legacyIds/commit/f78fc43a242e7d23b888585cc483a5a99fdfba7a) which renames the Feature
- Logs 
image
- new id package: [new-color](https://github.com/samruddhikhandale/test-legacyIds/pkgs/container/test-legacyids%2Fnew-color/60068268?tag=1.0.10)
- legacyId package: [color](https://github.com/samruddhikhandale/test-legacyIds/pkgs/container/test-legacyids%2Fcolor/60068278?tag=1.0.10)

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -47,6 +47,7 @@ export interface Feature {
 	included: boolean; // set programmatically
 	customizations?: VSCodeCustomizations;
 	installsAfter?: string[];
+	legacyIds?: string[];
 }
 
 export type FeatureOption = {
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -48,6 +48,7 @@ export interface Feature {
 	customizations?: VSCodeCustomizations;
 	installsAfter?: string[];
 	legacyIds?: string[];
+	currentId?: string; // set programmatically
 }
 
 export type FeatureOption = {
--- a/src/spec-configuration/containerFeaturesOrder.ts
+++ b/src/spec-configuration/containerFeaturesOrder.ts
@@ -49,12 +49,36 @@ export function computeInstallationOrder(features: FeatureSet[]) {
         feature,
         before: new Set(),
         after: new Set(),
-    })).reduce((map, feature) => map.set(feature.feature.sourceInformation.userFeatureId, feature), new Map<string, FeatureNode>());
+    })).reduce((map, feature) => map.set(feature.feature.sourceInformation.userFeatureId.split(':')[0], feature), new Map<string, FeatureNode>());
 
-    const nodes = [...nodesById.values()];
+    let nodes = [...nodesById.values()];
+
+    // Currently legacyIds only contain an id, hence append `registry/namespace` to it.
+    nodes = nodes.map(node => {
+        if (node.feature.sourceInformation.type === 'oci' && node.feature.features[0].legacyIds && node.feature.features[0].legacyIds.length > 0) {
+            const featureRef = node.feature.sourceInformation.featureRef;
+            if (featureRef) {
+                node.feature.features[0].legacyIds = node.feature.features[0].legacyIds.map(legacyId => `${featureRef.registry}/${featureRef.namespace}/` + legacyId);
+                node.feature.features[0].currentId = `${featureRef.registry}/${featureRef.namespace}/${node.feature.features[0].currentId}`;
+            }
+        }
+        return node;
+    });
+
     for (const later of nodes) {
         for (const firstId of later.feature.features[0].installsAfter || []) {
-            const first = nodesById.get(firstId);
+            let first = nodesById.get(firstId);
+
+            // Check for legacyIds (back compat)
+            if (!first) {
+                first = nodes.find(node => node.feature.features[0].legacyIds?.includes(firstId));
+            }
+
+            // Check for currentId (forward compat)
+            if (!first) {
+                first = nodes.find(node => node.feature.features[0].currentId === firstId);
+            }
+
             // soft dependencies
             if (first) {
                 later.after.add(first);
@@ -100,7 +124,7 @@ export function computeInstallationOrder(features: FeatureSet[]) {
 
     const missing = new Set(nodesById.keys());
     for (const feature of orderedFeatures) {
-        missing.delete(feature.sourceInformation.userFeatureId);
+        missing.delete(feature.sourceInformation.userFeatureId.split(':')[0]);
     }
 
     if (missing.size !== 0) {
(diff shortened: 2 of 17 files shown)


src/spec-node/featuresCLI/publish.ts
```diff
@@ -90,11 +90,45 @@ async function featuresPublish({
             process.exit(1);
         }
 
-        const thisResult = (publishResult?.digest && publishResult?.publishedVersions.length > 0) ? {
+        const ispublished = (publishResult?.digest && publishResult?.publishedVersions.length > 0);
```
joshspicer, line 93:
```suggestion
        const isPublished = (publishResult?.digest && publishResult?.publishedVersions.length > 0);
```

src/spec-node/featuresCLI/publish.ts
```diff
@@ -90,11 +90,45 @@ async function featuresPublish({
...
+        let thisResult = ispublished ? {
             ...publishResult,
             version: f.version,
```
joshspicer, line 96: I think I would rather have `publishedLegacyIds` always a part of the returned object (an empty array to start, which the loop below can append to if necessary).

That way, we either always get:
1. An object with all properties (if something actually happened, ie: We published tags)
2. A completely empty object (if nothing happened)

src/spec-node/featuresCLI/publish.ts
```diff
@@ -90,11 +90,45 @@ async function featuresPublish({
...
         } : {};
 
+        if (ispublished && f.legacyIds) {
+            output.write(`Processing legacyIds for '${f.id}'...`, LogLevel.Info);
+
+            let publishedLegacyIds: string[] = [];
+            for await (const legacyId of f.legacyIds) {
+                output.write(`Processing feature: ${legacyId}...`, LogLevel.Info);
```
joshspicer, line 104: I think something like this would read better, given the output on line 100.

```suggestion
                output.write(`Processing legacyId '${legacyId}'...`, LogLevel.Info);
```

src/test/container-features/containerFeaturesOCIPush.test.ts
```diff
@@ -18,6 +18,7 @@ interface PublishResult {
 	publishedVersions: string[];
 	digest: string;
 	version: string;
+	publishedLegacyIds?: string[];
```
joshspicer, line 21: This need not be optional if you take my suggestion above.

src/spec-node/collectionCommonUtils/publishCommandImpl.ts
```diff
@@ -39,7 +39,7 @@ export function getSemanticVersions(version: string, publishedVersions: string[]
 	return semanticVersions;
 }
 
-export async function doPublishCommand(version: string, ociRef: OCIRef, outputDir: string, output: Log, collectionType: string) {
+export async function doPublishCommand(version: string, ociRef: OCIRef, outputDir: string, output: Log, collectionType: string, archiveId?: string) {
```
joshspicer, line 42: Maybe we pre-compute the result of `getArchiveName()`, and pass that as an argument to `doPublishCommand`?  Then, we wouldn't need an optional parameter that changes the previous behavior.

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -47,6 +47,7 @@ export interface Feature {
 	included: boolean; // set programmatically
 	customizations?: VSCodeCustomizations;
 	installsAfter?: string[];
+	legacyIds?: string[];
```
joshspicer, line 50: Please be sure to add this to the schema (https://github.com/devcontainers/spec/blob/main/schemas/devContainerFeature.schema.json) when you're ready to!

joshspicer (approved):

src/spec-node/featuresCLI/publish.ts
samruddhikhandale: Renaming a Feature would happen rarely, so I was thinking that we shouldn't always add this object. It could confuse people as there would be very few people who would benefit from it.

Also, this way we could have a clear publishResult, and if something goes wrong, we could always look at the logs 😅

samruddhikhandale: for eg. if you look at a sample 

image

There is already so much info, but this is currently very helpful. However, if every item had `publishedLegacyIds: {}` would it become a bit extra? 🤔 

These were my thoughts when I made this code change, but happy to modify if we think otherwise.

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Yep, I'll start with the spec PR in parallel. Thanks.

src/spec-node/collectionCommonUtils/publishCommandImpl.ts
samruddhikhandale: nice 💡

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: https://github.com/devcontainers/spec/tree/samruddhikhandale/features-add-legacyIds

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -49,12 +49,36 @@ export function computeInstallationOrder(features: FeatureSet[]) {
         feature,
         before: new Set(),
         after: new Set(),
-    })).reduce((map, feature) => map.set(feature.feature.sourceInformation.userFeatureId, feature), new Map<string, FeatureNode>());
+    })).reduce((map, feature) => map.set(feature.feature.sourceInformation.userFeatureId.split(':')[0], feature), new Map<string, FeatureNode>());
```
samruddhikhandale, line 52: The code below for `for (const firstId of later.feature.features[0].installsAfter || []) {` was broken & wasn't having any effect on the computed order.

the installsAfter has id without version, however, this line was setting id with version. Hence, 

> let first = nodesById.get(firstId);

never returned any item. Fixed that.

samruddhikhandale: Also reported in https://github.com/devcontainers/cli/issues/345

chrmarti (approved):
LGTM!

BusHero:
When is it planned for this PR to be released?

samruddhikhandale:
Sometime today, prepping the release --> https://github.com/devcontainers/cli/pull/354

samruddhikhandale:
@BusHero Released with v0.27.0 🚀

Issue #336, opened 2022-12-22
bug: Cannot recognise bash-style arguments in Dockerfile


ShauryaAg:
devcontainer cli is unable to parse bash-style arguments, that is, arguments of type `${arg1:+"something"}`

For example:
When using the below `Dockerfile` 
```Dockerfile
ARG cloud
FROM ${cloud:+"my.ecr.repo/"}myimage:tag
```

The output after parsing should be, if the value `cloud` ARG is set:
```Dockerfile
FROM my.exr.repo/myimage:tag
```
and otherwise should be:
```Dockerfile
FROM myimage:tag
```

However, 
`devcontainer` cli throws an error that it cannot parse the above mentioned `Dockerfile`
```
[11 ms] @devcontainers/cli 0.25.2. Node.js v16.14.2. darwin 21.2.0 arm64.
[469 ms] Parsed version '+' for input 'docker.io/library/1:+' failed validation.
Error: Command failed: docker inspect --type image 1:+
    at Aoe (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:70182:9)
    at uT (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:70175:11)
    at async Poe (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:70543:34)
    at async Zf (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:70597:9)
    at async aue (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:72760:13)
    at async oue (/Users/shaurya/.vscode/extensions/ms-vscode-remote.remote-containers-0.266.1/dist/spec-node/devContainersSpecCLI.js:72747:9)
{"outcome":"error","message":"Command failed: docker inspect --type image 1:+","description":"An error occurred setting up the container."}
```

In the above run, the value of `cloud` argument was set to `1`, hence it outputted `1:+`, which is not the expected result.

devcontainer cli version: 0.25.2

Reference:
- https://docs.docker.com/engine/reference/builder/#environment-replacement

ShauryaAg:
[...]

PR #337, opened 2022-12-22
fix: handle parsing variable expression in dockerfile

ShauryaAg:
fixes: #336 

- Add support for parsing variable expressions in `Dockerfile` as mentioned in Dockerfile references: https://docs.docker.com/engine/reference/builder/#environment-replacement

### DIFF
--- a/src/spec-node/dockerfileUtils.ts
+++ b/src/spec-node/dockerfileUtils.ts
@@ -17,7 +17,7 @@ const fromStatement = /^\s*FROM\s+(?<platform>--platform=\S+\s+)?"?(?<image>[^\s
 const argEnvUserStatements = /^\s*(?<instruction>ARG|ENV|USER)\s+(?<name>[^\s=]+)([ =]+("(?<value1>\S+)"|(?<value2>\S+)))?/gm;
 const directives = /^\s*#\s*(?<name>\S+)\s*=\s*(?<value>.+)/;
 
-const argumentExpression = /\$\{(?<variable>[^:\}]+)(?<isVarExp>:(?<option>-|\+)(?<word>[^\}]+))?(\}|\}[^\$])/g;
+const argumentExpression = /\$\{?(?<variable>[^:\}]+)(?<isVarExp>:(?<option>-|\+)(?<word>[^\}]+))?\}?/g;
 
 export interface Dockerfile {
 	preamble: {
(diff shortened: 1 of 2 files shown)


ShauryaAg:

ShauryaAg:

src/spec-node/dockerfileUtils.ts
```diff
@@ -138,7 +141,40 @@ function extractInstructions(stageStr: string) {
...
+	}).join('');
+}
+
 function replaceVariables(dockerfile: Dockerfile, buildArgs: Record<string, string>, baseImageEnv: Record<string, string>, str: string, stage: { from?: From; instructions: Instruction[] }, beforeInstructionIndex: number) {
+	// Handle replacing variable expressions first (${var:+word}) if they exist
+	if (positiveVariableExpression.test(str) || negativeVariableExpression.test(str)) {
+		const image = handleExpressionVariables(dockerfile, buildArgs, baseImageEnv, str, stage, beforeInstructionIndex);
+		return image;
+	}
+
```
chrmarti, line 177: Would it make sense to handle all 3 cases (regular, negative and positive) below? I think this could be done in a single regex?

src/spec-node/dockerfileUtils.ts
```diff
@@ -138,7 +141,40 @@ function extractInstructions(stageStr: string) {
 		});
 }
 
+function handleExpressionVariables(dockerfile: Dockerfile, buildArgs: Record<string, string>, baseImageEnv: Record<string, string>, str: string, stage: { from?: From; instructions: Instruction[] }, beforeInstructionIndex: number) {
```
chrmarti, line 144: Please add a few tests in dockerFileUtils.test.ts.

chrmarti (changes requested):

src/spec-node/dockerfileUtils.ts
ShauryaAg: Fixed it! Modified it to use only one regex for all 3 cases.

src/spec-node/dockerfileUtils.ts
ShauryaAg: Added tests.

ShauryaAg:
Known issues:
The current implementation does not support _quoted_ expression variables, that is `${arg:+"word"}` will not be evaluated correctly in `FROM` statements as the `fromStatement` regex does not parse such `FROM` lines correctly. 

cc: @chrmarti

chrmarti (approved):
LGTM, thanks! (Getting a second reviewer now before we can merge.)

joshspicer (approved):

PR #340, opened 2022-12-22
clean up arguments in Features code

joshspicer:
Combines arguments into a single `param` object that we can pass around instead.


chrmarti (approved):
LGTM!

Issue #322, opened 2022-12-09
Publishing features to other registries than ghcr.io 


aacebedo:
Hi 

I tried to publish a feature to a private repo on another package registry (gitlab) and it failed. 
Seems that identification token is only handled for GH registries.

Is it confirmed? Is there some way to do this?

samruddhikhandale:
Hi @aacebedo 👋 

Does `gitlab` use an [OCI registry](https://containers.dev/implementors/features-distribution/#oci-registry)? I believe the CLI currently supports only OCI.

@joshspicer could you help correct my statement if I'm missing something.

Clockwork-Muse:
There's at least one issue [asking them to support it](https://gitlab.com/gitlab-org/gitlab/-/issues/385277), so I don't think they implement the distribution spec.

That said, I'm currently not able to publish to a private azure registry, [although it appears that it should be possible to](https://github.com/opencontainers/oci-conformance/blob/main/distribution-spec/live/azurecontainerregistry/badges.md) (I get the same error as for gitlab, so maybe not).

Clockwork-Muse:
More concretely, what's failing is attempts to get login tokens to check for versions before publish (but after build/test/package).  As near as I'm aware, there's no documentation around what's required to be able to upload to private/secured registries - and that you might need different credentials to do so (for instance if you have 2FA enabled, you likely need a credential helper to manage a token).

The cli is trying to get a token from `{registry}/token`:
 - Azure has a (beta/preview) token endpoint at `{registry}/oauth2/token`
 - gitlab has nothing.

joshspicer:
@aacebedo - Yes, the registry needs to support the [distribution spec](https://github.com/opencontainers/distribution-spec) (some registries are listed here: https://oras.land/implementors).  Note that we only auto-detect the GITHUB_TOKEN for auth at the moment (work to

I recently merged a PR that exercises the CLI against the reference `registry` implementation (https://github.com/devcontainers/cli/pull/318). **It just went in so it's not released to npm, but you can build from source to play with it.**

We have an undocumented mechanism to auth against other registries with an environment variable https://github.com/devcontainers/cli/blob/fd4de3fc50fdf299c1d9dce64a6d0c4f29dfe935/src/spec-configuration/containerCollectionsOCI.ts#L262-L270

We have an open work item to read in the local `.docker/config.json` for auth with other registries, but that work is still in flight.

joshspicer:
[...]

PR #341, opened 2022-12-22
Follow Docker Token Authentication Specification

joshspicer:
ref: https://github.com/devcontainers/cli/issues/322

Implements the [Docker Registry v2 authentication via central service specification](https://docs.docker.com/registry/spec/auth/token/) to improve compatibility with registry services.  While not officially part of the distribution specification, this specification is widely used by various container registries.  This change has been tested with Azure Container Registry (acr) and GitHub Container Registry (ghcr). 

As outlined in the spec, requests will initially be made to a registry without any authentication.  If authentication is required by the server, a `401` response will be set and the `WWW-Authenticate` header will be set with information on how to exchange for a bearer token for the given resource, as well as which scopes to request.  

```
realm="https://ghcr.io/token",service="ghcr.io",scope="repository:devcontainers/features:pull,push"
```

Registries like GHCR and ACR require this token exchange even for anonymous access (pulling a public artifact).

This PR wraps all pulling and pushing HTTP operations in `requestEnsureAuthenticated(...)`, which will read the `WWW-Authenticate` header on `401` and attempt to negotiate a token with the server.  The last token to be successfully used will be cached and first attempted on subsequent requests.

This implementation supports exchanging for a Bearer token (as described in the linked specification from Docker), as well as setting the authentication header with Basic auth credentials (as expected by the `registry/registry` reference implementation

Additionally, this PR will now attempt to read the `$HOME/.docker/config.json` file on disk (if one exists) and will share these with the registry server to get a more privileged Bearer token.  By following [the 'Authenticate using token' steps on the Azure Container Registry Docs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions#authenticate-using-token), one will be able to both publish and pull from an ACR that requires authentication. The existing `DEVCONTAINERS_OCI_AUTH` environment variable is still supported, and is preferred over reading the home folder config.json.  For `GHCR`, the environment's `GITHUB_TOKEN` is always preferred.

A test has been added to validate that anonymously pulling a Feature from ACR works end to end.  That Feature was pushed to ACR via the changes in this PR.

image

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -6,12 +6,12 @@ import * as jsonc from 'jsonc-parser';
 import { request } from '../spec-utils/httpRequest';
 import { Log, LogLevel } from '../spec-utils/log';
 import { isLocalFile, mkdirpLocal, readLocalFile, writeLocalFile } from '../spec-utils/pfs';
+import { HEADERS } from './httpOCIRegistry';
 
 export const DEVCONTAINER_MANIFEST_MEDIATYPE = 'application/vnd.devcontainers';
 export const DEVCONTAINER_TAR_LAYER_MEDIATYPE = 'application/vnd.devcontainers.layer.v1+tar';
 export const DEVCONTAINER_COLLECTION_LAYER_MEDIATYPE = 'application/vnd.devcontainers.collection.layer.v1+json';
 
-export type HEADERS = { 'authorization'?: string; 'user-agent': string; 'content-type'?: string; 'accept'?: string; 'content-length'?: string };
 
 export interface CommonParams {
 	env: NodeJS.ProcessEnv;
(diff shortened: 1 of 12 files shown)


joshspicer:
cc/ @brendandburns

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -3,19 +3,19 @@ import * as semver from 'semver';
...
 export const DEVCONTAINER_MANIFEST_MEDIATYPE = 'application/vnd.devcontainers';
 export const DEVCONTAINER_TAR_LAYER_MEDIATYPE = 'application/vnd.devcontainers.layer.v1+tar';
 export const DEVCONTAINER_COLLECTION_LAYER_MEDIATYPE = 'application/vnd.devcontainers.collection.layer.v1+json';
 
-export type HEADERS = { 'authorization'?: string; 'user-agent': string; 'content-type'?: string; 'accept'?: string; 'content-length'?: string };
 
 export interface CommonParams {
 	env: NodeJS.ProcessEnv;
 	output: Log;
+	cachedAuthHeader?: string;
```
chrmarti, line 18: Should this be a map from registries to auth headers?

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -0,0 +1,233 @@
...
+			return Buffer.from(userToken)
+				.toString('base64');
+		}
+	} else {
+		const homeDir = os.homedir();
+		if (homeDir) {
+			const dockerConfigPath = path.join(homeDir, '.docker', 'config.json');
+			if (await isLocalFile(dockerConfigPath)) {
+				try {
+					const dockerConfig: DockerConfigFile = JSON.parse((await readLocalFile(dockerConfigPath)).toString());
```
chrmarti, line 148: Could use `jsonc.parse` for additional robustness. (Though Docker might actually be strict from what I have seen.)

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -0,0 +1,233 @@
...
+
+					if (dockerConfig.auths && dockerConfig.auths[registry] && dockerConfig.auths[registry].auth) {
```
chrmarti, line 150: We should also support Docker credential helpers, these are more common in local setups. I recently figured out how to use them for the Dev Containers extension. The code will need to be adapted, it implements a credential helper proxy: https://github.com/microsoft/vscode-remote-containers/blob/main/src/common/dockerCredentialHelper.ts

Ok to do in a separate PR.

chrmarti (changes requested):
Great! Left a few comments.

src/spec-configuration/containerCollectionsOCI.ts
joshspicer: Good point. For the push operations I think this would be fine (we push to one registry at a time), but for pulling perhaps this `params` object is reused? Nonetheless makes sense to make it a map to be sure we never send an auth header to the wrong server.

src/spec-configuration/httpOCIRegistry.ts
joshspicer: Thanks for the pointer.  I have a TODO comment inline here to add credential helper support, and will tackle that in a subsequent PR.

joshspicer: https://github.com/devcontainers/cli/issues/349

alexdima (approved):

PR #344, opened 2022-12-23
Support Azure Container Registry anonymous pulls (and oauth2 pulls in general)

brendandburns:
cc @joshspicer

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -305,8 +305,8 @@ async function generateScopeTokenCredential(output: Log, registry: string, ociRe
 	const authServer = registry === 'docker.io' ? 'auth.docker.io' : registry;
 	const registryServer = registry === 'docker.io' ? 'registry.docker.io' : registry;
 
-	// TODO: Probably add other registries here, but I'm not sure which ones.
-	const isOauth2 = registry.endsWith('azurecr.io');
+	// TODO: Add other registries here as needed.
+	const isOauth2 = registry.endsWith('azurecr.io') || registry.endsWith('azurecr.cn') || registry.endsWith('azurecr.us');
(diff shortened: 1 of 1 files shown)


src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -304,7 +304,10 @@ async function generateScopeTokenCredential(output: Log, registry: string, ociRe
 
 	const authServer = registry === 'docker.io' ? 'auth.docker.io' : registry;
 	const registryServer = registry === 'docker.io' ? 'registry.docker.io' : registry;
-	const url = `https://${authServer}/token?scope=repository:${ociRepoPath}:${operationScopes}&service=${registryServer}`;
+
+	// TODO: Probably add other registries here, but I'm not sure which ones.
+	const isOauth2 = registry.endsWith('azurecr.io');
```
northtyphoon, line 309: For Azure Container Registry, azurecr.io (public) / azurecr.cn (China) / azurecr.us (US Gov)

brendandburns: Done.

joshspicer:
Thanks for taking the time!

This looks great to patch/support Azure, although it's not generalized to support any registry that implements the docker spec (see issue discussion: https://github.com/devcontainers/cli/issues/322)

I started working on https://github.com/devcontainers/cli/pull/341 before I left for the holidays, which will implement the docker authorization spec. 

@brendandburns do you know if Azure CR is implementing the docker spec (will send a 401 with an authorization endpoint to use in the `WWW-Authenticate` Header)?  If so, it might make more sense to take my change instead as it will be more generic.  If Azure is a special case then this makes a lot of sense to add in

joshspicer:
It appears ACR does implement the docker auth flow! 
image

joshspicer:
I've merged https://github.com/devcontainers/cli/pull/341, which should complete the goal of this PR and more.  

Please feel free to create an issue and tag me in the issue if anything is missed that this PR would provide.

PR #346, opened 2022-12-27
Features: Adds a new 'deprecated' property

samruddhikhandale:
Spec PR - https://github.com/devcontainers/spec/pull/163
Spec issue - https://github.com/devcontainers/spec/issues/146

PR for a similar change to the Feature Schema - https://github.com/devcontainers/spec/pull/170


schemas/devcontainer-features.schema.json
```diff
@@ -139,6 +139,10 @@
                             "description": "Machine specific settings that should be copied into the container."
                         }
                     }
+                },
+                "deprecated": {
```
samruddhikhandale, line 143: Probably this file should be deleted ; if it's in use anywhere then should be replaced with 👇  
https://github.com/devcontainers/spec/blob/main/schemas/devContainerFeature.schema.json

chrmarti: @joshspicer Do we still need this file here? I think this might just be your first version of it and I copied it over when doing the CLI.

joshspicer: correct, we don't need this older file (`devcontainer-features.json` is not a thing anymore).

joshspicer: `devcontainer-feature.json` is the file that I just created that new schema definition file for.

samruddhikhandale: Good to know. Removed 👍

joshspicer (approved):

PR #362, opened 2023-01-04
Add dotfiles functionallity

edgonmsft:
Dotfiles are commonly used as ann extra customization step in devcontainers.

This pr adds 3 more parameters to the `up` command.
If the values are not present then dotfiles are not executed.

- `dotfiles-repository`: Git URL to clone a dotfiles repository from.
- `dotfiles-install-command`: Command to install the dotfiles with. If none is given a list of script names (install.sh, install, bootstrap.sh, bootstrap, setup.sh and setup) are checked for in the checked out dotfiles repository and if none is found all top-level dotfiles are symlinked from the container\'s home folder.
- `dotfiles-target-path`: Folder path to clone the dotfiles repository to.


src/spec-common/dotfiles.ts
```diff
@@ -0,0 +1,112 @@
...
+		break
+	fi
+done
+if [ -z "$installCommand" ]
+then
+	dotfiles=$(ls -d ${targetPath}/.* 2>/dev/null | grep -v -E '/(.|..|.git)$')
+	if [ ! -z "$dotfiles" ]
+	then
+		echo Linking dotfiles: $dotfiles
+		ln -sf $dotfiles ~ 2>/dev/null
```
samruddhikhandale, line 75:
```suggestion
		ln -snf $dotfiles ~ 2>/dev/null
```

src/test/cli.up.test.ts
```diff
@@ -33,6 +33,15 @@ describe('Dev Containers CLI', function () {
 			await shellExec(`docker rm -f ${containerId}`);
 		});
 
+		it('should execute successfully with valid config and dotfiles', async () => {
+			const res = await shellExec(`${cli} up --workspace-folder ${__dirname}/configs/image --dotfiles-repository https://github.com/devcontainers/dotfiles`);
```
samruddhikhandale, line 37: This dotfiles repo is internal, I wonder if we should reference it here? Maybe you could create a public repo in https://github.com/codspace, our tests use that.

src/test/cli.up.test.ts
```diff
@@ -33,6 +33,15 @@ describe('Dev Containers CLI', function () {
...
+			const response = JSON.parse(res.stdout);
+			assert.equal(response.outcome, 'success');
```
samruddhikhandale, line 39: Can we add asserts to validate if the dotfiles changes are reflected?

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -117,6 +117,9 @@ function provisionOptions(y: Argv) {
 		'skip-feature-auto-mapping': { type: 'boolean', default: false, hidden: true, description: 'Temporary option for testing.' },
 		'skip-post-attach': { type: 'boolean', default: false, description: 'Do not run postAttachCommand.' },
 		'experimental-image-metadata': { type: 'boolean', default: experimentalImageMetadataDefault, hidden: true, description: 'Temporary option for testing.' },
+		'dotfiles-repository': { type: 'string', description: 'Git URL to clone a dotfiles repository from.' },
```
samruddhikhandale, line 120: Suggestion: Should we use the wordings from https://code.visualstudio.com/docs/devcontainers/containers#_personalizing-with-dotfile-repositories ? 

```suggestion
		'dotfiles-repository': { type: 'string', description: 'URL of a dotfiles Git repository (e.g., https://github.com/owner/repository.git) or owner/repository of a GitHub repository. },
```

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -117,6 +117,9 @@ function provisionOptions(y: Argv) {
...
+		'dotfiles-install-command': { type: 'string', implies: 'dotfiles-repository', description: 'Command to install the dotfiles with. If none is given a list of script names (install.sh, install, bootstrap.sh, bootstrap, setup.sh and setup) are checked for in the checked out dotfiles repository and if none is found all top-level dotfiles are symlinked from the container\'s home folder.' },
```
samruddhikhandale, line 121:
```suggestion
		'dotfiles-install-command': { type: 'string', implies: 'dotfiles-repository', description: 'The command to run after cloning the dotfiles repository. Defaults to run the first file of `install.sh`, `install`, `bootstrap.sh`, `bootstrap`, `setup.sh` and `setup` found in the dotfiles repository's root folder.' },
```

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -117,6 +117,9 @@ function provisionOptions(y: Argv) {
...
+		'dotfiles-target-path': { type: 'string', implies: 'dotfiles-repository', description: 'Folder path to clone the dotfiles repository to.' },
```
samruddhikhandale, line 122:
```suggestion
		'dotfiles-target-path': { type: 'string', implies: 'dotfiles-repository', description: 'The path to clone the dotfiles repository to. Defaults to `~/dotfiles`.' },
```

samruddhikhandale, line 122:
```suggestion
		'dotfiles-target-path': { type: 'string', implies: 'dotfiles-repository', default: '~/dotfiles', description: 'Folder path to clone the dotfiles repository to.' },
```

src/spec-node/devContainers.ts
```diff
@@ -133,6 +138,11 @@ export async function createDockerParams(options: ProvisionOptions, disposables:
 		skipPostAttach: options.skipPostAttach,
 		experimentalImageMetadata: options.experimentalImageMetadata,
 		skipPersistingCustomizationsFromFeatures: options.skipPersistingCustomizationsFromFeatures,
+		dotfilesConfiguration: {
+			repository: options.dotfiles.repository,
+			installCommand: options.dotfiles.installCommand,
+			targetPath: options.dotfiles.targetPath || '~/dotfiles',
```
samruddhikhandale, line 144: If you apply the suggestion below (which adds the default property ), then can avoid this.

```suggestion
			targetPath: options.dotfiles.targetPath,
```

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -519,6 +531,9 @@ function runUserCommandsOptions(y: Argv) {
 		'skip-feature-auto-mapping': { type: 'boolean', default: false, hidden: true, description: 'Temporary option for testing.' },
 		'skip-post-attach': { type: 'boolean', default: false, description: 'Do not run postAttachCommand.' },
 		'experimental-image-metadata': { type: 'boolean', default: experimentalImageMetadataDefault, hidden: true, description: 'Temporary option for testing.' },
+		'dotfiles-repository': { type: 'string', description: 'Git URL to clone a dotfiles repository from.' },
```
samruddhikhandale, line 534: Same description comments as above!

samruddhikhandale (commented):
Left some comments!

src/spec-common/dotfiles.ts
edgonmsft: It has always been like this so not sure if we should change it.

src/test/cli.up.test.ts
edgonmsft: The idea was to make that repo public after merging this pr.

edgonmsft: Moved to a different repo in that org.

src/test/cli.up.test.ts
edgonmsft: Added

src/test/cli.up.test.ts
```diff
@@ -33,6 +33,17 @@ describe('Dev Containers CLI', function () {
 			await shellExec(`docker rm -f ${containerId}`);
 		});
 
+		it('should execute successfully with valid config and dotfiles', async () => {
+			const res = await shellExec(`${cli} up --workspace-folder ${__dirname}/configs/image-with-git-feature --dotfiles-repository https://github.com/codspace/test-dotfiles`);
+			const response = JSON.parse(res.stdout);
+			assert.equal(response.outcome, 'success');
+			const containerId: string = response.containerId;
+			assert.ok(containerId, 'Container id not found.');
+			const dotfiles = await pathExists(cli, `${__dirname}/configs/image-with-git-feature`, `/tmp/.dotfilesMarker`);
```
chrmarti, line 42: Why wouldn't this be `~/.devcontainer/.dotfilesMarker`? 🤔

src/spec-common/dotfiles.ts
```diff
@@ -0,0 +1,112 @@
...
+		break
+	fi
+done
+if [ -z "$installCommand" ]
+then
+	dotfiles=$(ls -d ${targetPath}/.* 2>/dev/null | grep -v -E '/(.|..|.git)$')
+	if [ ! -z "$dotfiles" ]
+	then
+		echo Linking dotfiles: $dotfiles
+		ln -snf $dotfiles ~ 2>/dev/null
```
chrmarti, line 75: Noticed you added `-n`, does that fix a particular case?

chrmarti (changes requested):

src/test/cli.up.test.ts
edgonmsft: The dotfiles repo used does a `touch` on that file similar to the other tests we have.

src/spec-common/dotfiles.ts
```diff
@@ -0,0 +1,112 @@
+/*---------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
edgonmsft: Added

src/spec-common/injectHeadless.ts
```diff
@@ -307,6 +315,9 @@ export async function setupInContainer(params: ResolverParameters, containerProp
 	if (params.postCreate.enabled) {
 		await runPostCreateCommands(params, containerProperties, updatedConfig, remoteEnv, false);
 	}
+	if (params.dotfilesConfiguration) {
```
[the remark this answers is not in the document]
edgonmsft: Changed the order, I found confusing that its before the personalization check instead of after?

src/spec-common/dotfiles.ts
```diff
@@ -0,0 +1,112 @@
...
+		});
+	}
+}
+
+function quoteValue(value: string | undefined) {
+	return `'${(value || '').replace(/'+/g, '\'"$&"\'')}'`;
+}
+
+function getDotfilesMarkerFile(properties: ContainerProperties) {
+	const machineLocation = path.posix.join(typeof properties === 'string' ? properties : properties.userDataFolder, 'data/Machine');
```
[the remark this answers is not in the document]
edgonmsft: Changed, we can handle the back-compat internally.

src/spec-common/dotfiles.ts
edgonmsft: It was a previous suggestion but nothing in particular, i removed it to be consistent.

src/spec-common/injectHeadless.ts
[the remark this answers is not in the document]
chrmarti: A `~/.gitconfig` from dotfiles would win over a `~/.gitconfig` from the local machine in the Dev Containers extension (which skips copying the local one when there already is one in the container). Maybe it could argued that it should be the other way around, this is just the current order. Ideally we can keep it this way to allow for adoption without breaking existing setups.

chrmarti (approved):
👍

PR #366, opened 2023-01-04
Features preamble: Add warnings for Feature renames & deprecation

samruddhikhandale:
Deprecation msg
image

Renaming msg
image

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -270,7 +270,7 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 		? ` Look at the documentation at ${documentation} for help troubleshooting this error.`
 		: '';
 
-	return `#!/bin/bash
+	return `#!/bin/sh
 set -e
 
 on_exit () {
@@ -281,9 +281,11 @@ on_exit () {
 trap on_exit EXIT
 
 echo ===========================================================================
-if [[ "${warningHeader}" != '' ]]; then
+if [ "${warningHeader}" != '' ]
+then
 	echo '${warningHeader}'
 fi
+
 echo 'Feature       : ${name}'
 echo 'Description   : ${description}'
 echo 'Id            : ${id}'
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -262,7 +262,7 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 	}
 
 	if (feature?.legacyIds && feature.legacyIds.length > 0 && feature.id !== feature.currentId) {
-		warningHeader += `(!) WARNING: Using the old Feature ID, update the reference to "${feature.currentId}".`;
+		warningHeader += `(!) WARNING: This feature has been renamed. Please update the reference in devcontainer.json to "${feature.currentId}".`;
 	}
 
 	const errorMessage = `ERROR: Feature "${name}" (${id}) failed to install!`;
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -256,6 +256,15 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 	const documentation = escapeQuotesForShell(feature.documentationURL ?? '');
 	const optionsIndented = escapeQuotesForShell(options.map(x => `    ${x}`).join('\n'));
 
+	let warningHeader = '';
+	if (feature.deprecated) {
+		warningHeader += `(!) WARNING: Using the deprecated Feature "${feature.id}". This Feature will no longer receive any further updates/support.\n`;
+	}
+
+	if (feature?.legacyIds && feature.legacyIds.length > 0 && feature.id !== feature.currentId) {
+		warningHeader += `(!) WARNING: Using the old Feature ID, update the reference to "${feature.currentId}".`;
+	}
+
 	const errorMessage = `ERROR: Feature "${name}" (${id}) failed to install!`;
 	const troubleshootingMessage = documentation
 		? ` Look at the documentation at ${documentation} for help troubleshooting this error.`
@@ -272,6 +281,7 @@ on_exit () {
 trap on_exit EXIT
 
 echo ===========================================================================
+echo '${warningHeader}'
 echo 'Feature       : ${name}'
 echo 'Description   : ${description}'
 echo 'Id            : ${id}'
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -258,11 +258,11 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 
 	let warningHeader = '';
 	if (feature.deprecated) {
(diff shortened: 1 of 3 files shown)


src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -256,6 +256,15 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 	const documentation = escapeQuotesForShell(feature.documentationURL ?? '');
 	const optionsIndented = escapeQuotesForShell(options.map(x => `    ${x}`).join('\n'));
 
+	let warningHeader = '';
+	if (feature.deprecated) {
+		warningHeader += `(!) WARNING: Using the deprecated Feature "${feature.id}". This Feature will no longer receive any further updates/support.\n`;
+	}
+
+	if (feature?.legacyIds && feature.legacyIds.length > 0 && feature.id !== feature.currentId) {
+		warningHeader += `(!) WARNING: Using the old Feature ID, update the reference to "${feature.currentId}".`;
```
jkeech, line 265:
```suggestion
		warningHeader += `(!) WARNING: This feature has been renamed. Please update the reference in devcontainer.json to "${feature.currentId}".`;
```

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -256,6 +256,15 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 	const documentation = escapeQuotesForShell(feature.documentationURL ?? '');
 	const optionsIndented = escapeQuotesForShell(options.map(x => `    ${x}`).join('\n'));
 
+	let warningHeader = '';
+	if (feature.deprecated) {
+		warningHeader += `(!) WARNING: Using the deprecated Feature "${feature.id}". This Feature will no longer receive any further updates/support.\n`;
```
jkeech, line 261: For all of these strings that come from user-provided metadata, we need to escape any potential quotes as done above with `escapeQuotesForShell`. I'm not sure if `feature.id` or `feature.currentId` can have any quote characters inside, but it wouldn't hurt to add the escaping.

joshspicer (dismissed):
Nice polish!

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -272,6 +281,7 @@ on_exit () {
 trap on_exit EXIT
 
 echo ===========================================================================
+echo '${warningHeader}'
```
jkeech, line 284: Can we only add this if there's a header defined? Otherwise it will add a blank line to each output when there's no warning.

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer, line 265:
```suggestion
		warningHeader += `(!) WARNING: devcontainer.json references a legacy ID. Update the reference to '${feature.currentId}'.`;
```

samruddhikhandale: Umm just merged John's suggestion. Let me know what you think of that.

> `(!) WARNING: This feature has been renamed. Please update the reference in devcontainer.json to "${feature.currentId}".`;

joshspicer: Oh I missed his comment. That looks good to me.

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Makes sense. Updated

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: 👍  Updated

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -256,12 +256,21 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
...
+		warningHeader += `(!) WARNING: This feature has been renamed. Please update the reference in devcontainer.json to "${escapeQuotesForShell(feature.currentId)}".`;
+	}
+
 	const errorMessage = `ERROR: Feature "${name}" (${id}) failed to install!`;
 	const troubleshootingMessage = documentation
 		? ` Look at the documentation at ${documentation} for help troubleshooting this error.`
 		: '';
 
-	return `#!/bin/sh
+	return `#!/bin/bash
```
jkeech, line 273: Can we keep this as `sh`? I had originally used `sh` here because it's available in more containers and unless we really need bash support, it would be good to use the most broadly-available shell. Some alpine containers don't have bash by default as an example and it would be good to support those as much as possible.

One way to do this would be to avoid the conditional in the script itself and put the conditional in the string concatenation that prints out the script.

jkeech (approved):
Thanks for applying the changes. I left one more comment about continuing to use sh instead of bash, but otherwise looks good.

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: 👍  Used an sh condition statement instead. https://github.com/devcontainers/cli/actions/runs/3849044939/jobs/6557564151#step:7:216

image

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -272,6 +281,11 @@ on_exit () {
 trap on_exit EXIT
 
 echo ===========================================================================
+if [ "${warningHeader}" != '' ]
```
jkeech, line 284: What do you think about doing something like the following:
```
const echoWarning = `echo '${warningHeader}'`;
...
return `#!/bin/sh
...
echo ===========================================================================
${echoWarning}
echo 'Feature       : ${name}'
...
```

That way we don't need to add a condition in the script that checks an inlined string. In the case where there's a warning to print, we generate the script with the echo added. In the case where there's no warning to print, we omit the echo entirely and the script output is roughly the same as it was before except there's a blank line in the script itself (not in the script output when the script is executed).

jkeech (dismissed):
LGTM. Left a minor suggestion to simplify the generated script a bit, particularly in the "no warning" case.

src/spec-configuration/containerFeaturesConfiguration.ts
samruddhikhandale: Sweet! Modified.

jkeech (approved):

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -256,6 +256,16 @@ export function getFeatureInstallWrapperScript(feature: Feature, featureSet: Fea
 	const documentation = escapeQuotesForShell(feature.documentationURL ?? '');
 	const optionsIndented = escapeQuotesForShell(options.map(x => `    ${x}`).join('\n'));
 
+	let warningHeader = '';
+	if (feature.deprecated) {
+		warningHeader += `(!) WARNING: Using the deprecated Feature "${escapeQuotesForShell(feature.id)}". This Feature will no longer receive any further updates/support.\n`;
+	}
+
+	if (feature?.legacyIds && feature.legacyIds.length > 0 && feature.currentId && feature.id !== feature.currentId) {
+		warningHeader += `(!) WARNING: This feature has been renamed. Please update the reference in devcontainer.json to "${escapeQuotesForShell(feature.currentId)}".`;
```
eitsupi, line 265: Wouldn't it be better to end with `\n` same as line 261?

PR #373, opened 2023-01-13
feat: add git utils methods

ShauryaAg:
- Added git utils methods to clone a repository, which is heavily based on docker's gitutils: https://github.com/moby/moby/blob/master/builder/remotecontext/git/gitutils.go

### DIFF
--- a/src/spec-utils/git.ts
+++ b/src/spec-utils/git.ts
@@ -1,0 +1,174 @@
+import fs, { PathLike } from 'fs';
+import child_process from 'child_process';
+import path from 'path';
+
+import { URI } from 'vscode-uri';
+
+interface GitRepo {
+  remote: string;
+  ref: string;
+  subdir?: string;
+
+  isolateConfig?: boolean;
+}
+
+export function cloneGitRepo(remoteUrl: string): string {
+  const repo = parseRemoteUrl(remoteUrl);
+
+  return _cloneGitRepo(repo);
+}
+
+function fetchArgs(remoteUrl: string, ref: string): Array<string> {
+  let args = ['fetch'];
+
+  if (supportsShallowClone(remoteUrl)) {
+    args.push('--depth', '1');
+  }
+
+  args.push('origin', '--', ref);
+  return args;
+}
+
+function _cloneGitRepo(repo: GitRepo): string {
+  const fetch = fetchArgs(repo.remote, repo.ref);
+
+  const root = fs.mkdtempSync('vscode-dev-containers');
+  let [stdout, stderr] = execGitWithinDir(repo, root, 'init');
+  if (stderr) {
+    throw new Error(`failed to init git repo in ${root}: ${stderr}`);
+  }
+
+  [stdout, stderr] = execGitWithinDir(repo, root, 'remote', 'add', 'origin', repo.remote); // prettier-ignore
+  if (stderr) {
+    throw new Error(
+      `failed to add remote ${repo.remote} to ${root}: ${stderr}`
+    );
+  }
+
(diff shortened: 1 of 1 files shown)


src/spec-utils/git.ts
```diff
@@ -0,0 +1,201 @@
...
+import child_process from 'child_process';
+import path from 'path';
+
+import { URI } from 'vscode-uri';
+
+interface GitRepo {
+  remote: string;
+  ref: string;
+  subdir?: string;
+
```
samruddhikhandale, line 12: nit: remove blank line?

samruddhikhandale (changes requested):
Thanks, this is awesome. Can we add tests to validate this?

ShauryaAg:
> Thanks, this is awesome. Can we add tests to validate this?

@samruddhikhandale I am looking to write tests for it, but a little unsure as to how we can do that. Since the context for the service is changed by the override file and there's no method to get the final context, it would require a significant refactor to write tests for this. Please feel free to correct/guide me.

samruddhikhandale:
@ShauryaAg I have few ideas on how you could add tests -
-  Can we simply add a docker-compose config with `context: https://**/**.git` and add a similar test like https://github.com/devcontainers/cli/blob/main/src/test/cli.build.test.ts#L36-L43 - Currently, such a config fails to successfully build a devcontainer, hence, we could simply check if the build pass. Also, we could add few asserts to validate if the devcontainer was built correctly, (eg - presence of a file/tooling). 
- Validate functions by adding asserts for functions like `isGitUrl()` , `_cloneGitRepo()` etc.

ShauryaAg:
@samruddhikhandale I have added one test as you mentioned in the previous comment, but I would still like to write tests for all the scenarios - git URL with SSH, parsing git ref, cloning a subdirectory, ensuring shallow clone works as expected. 

Writing tests for covering all the scenarios can be a little difficult and will use a lot of time (and network bandwidth) if done using the way mentioned by you (which is something I am trying to avoid for running tests). 

What would you suggest is the best way to work around that?

src/test/configs/compose-git-without-features/.devcontainer/docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
+version: '3'
+services:
+  app:
+    build: 
+      context: https://github.com/ShauryaAg/productapi-go.git
```
samruddhikhandale, line 5: Can we try not to use a personal repo? 
Feel free to use any repo in the https://github.com/devcontainers org.

samruddhikhandale (changes requested):
Left a comment.

samruddhikhandale:
Thanks for adding test coverage.

> Writing tests for covering all the scenarios can be a little difficult and will use a lot of time (and network bandwidth) if done using the way mentioned by you (which is something I am trying to avoid for running tests).

I believe writing a unit test to validate a function is quite helpful and improves code reliability. We already have multiple tests which validates individual functions & different features/sub commands of the CLI.

Also, given the time needed for tests execution, we have divided all tests into multiple jobs. Hence, parallelism helps with faster test runs.

src/test/configs/compose-git-without-features/.devcontainer/docker-compose.yaml
ShauryaAg: Changed to devcontainers/devcontainers.github.io repo.

ShauryaAg:
@samruddhikhandale I have a few more tests for verifying `cloneGitRepo` and `isGitUrl` functions.

ShauryaAg:
@chrmarti Can you please review?

PR #374, opened 2023-01-13
Cache user env for performance improvement

eljog:
We've seen that user environment prob command often take multiple seconds to run, which slows down the commands

With this PR, the caller of the `up` command or `run-user-commands` command can optionally pass the cache directory name if they would like to use the cache.

If a cache directory name is provided via `--container-session-data-folder` option, user envs will be cached in that directory, and the subsequent invocations will use the cache instead of probing user envs.

### DIFF
--- a/src/spec-common/injectHeadless.ts
+++ b/src/spec-common/injectHeadless.ts
@@ -669,8 +669,13 @@ async function patchEtcProfile(params: ResolverParameters, containerProperties: 
 }
 
 async function probeUserEnv(params: { defaultUserEnvProbe: UserEnvProbe; allowSystemConfigChange: boolean; output: Log; containerSessionDataFolder?: string }, containerProperties: { shell: string; remoteExec: ExecFunction; installFolder?: string; env?: NodeJS.ProcessEnv; shellServer?: ShellServer; launchRootShellServer?: (() => Promise<ShellServer>); user?: string }, config?: CommonMergedDevContainerConfig) {
-	let env = await readUserEnvFromCache(params.containerSessionDataFolder, containerProperties.shellServer);
-	if (env && Object.keys(env).length) {
+	let userEnvProbe = getUserEnvProb(config, params);
+	if (!userEnvProbe || userEnvProbe === 'none') {
+		return {};
+	}
+
+	let env = await readUserEnvFromCache(userEnvProbe, params, containerProperties.shellServer);
+	if (env) {
 		return env;
 	}
 
@@ -682,18 +687,18 @@ async function probeUserEnv(params: { defaultUserEnvProbe: UserEnvProbe; allowSy
 	}
 
 	if (env) {
-		await updateUserEnvCache(env, params.containerSessionDataFolder, containerProperties.shellServer);
+		await updateUserEnvCache(env, userEnvProbe, params, containerProperties.shellServer);
 	}
 
 	return env || {};
 }
 
-async function readUserEnvFromCache(cacheDir?: string, shellServer?: ShellServer) {
-	if (!shellServer || !cacheDir) {
+async function readUserEnvFromCache(userEnvProbe: UserEnvProbe, params: { output: Log; containerSessionDataFolder?: string }, shellServer?: ShellServer) {
+	if (!shellServer || !params.containerSessionDataFolder) {
 		return undefined;
 	}
 
-	const cacheFile = path.posix.join(cacheDir, 'env.json');
+	const cacheFile = path.posix.join(params.containerSessionDataFolder, `${userEnvProbe}.json`);
 	try {
 		if (await isFile(shellServer, cacheFile)) {
 			const { stdout } = await shellServer.exec(`cat '${cacheFile}'`);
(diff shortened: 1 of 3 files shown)


src/spec-common/injectHeadless.ts
```diff
@@ -667,14 +668,60 @@ async function patchEtcProfile(params: ResolverParameters, containerProperties:
...
+
+	const cacheFile = path.posix.join(cacheDir, 'env.json');
+	try {
+		if (await isFile(shellServer, cacheFile)) {
+			const { stdout } = await shellServer.exec(`cat '${cacheFile}'`);
+			return JSON.parse(stdout);
+		}
+	}
+	catch (e) {
+
```
chrmarti, line 704: Should we log this?

src/spec-common/injectHeadless.ts
```diff
@@ -667,14 +668,60 @@ async function patchEtcProfile(params: ResolverParameters, containerProperties:
...
+	const cacheFile = path.posix.join(cacheDir, 'env.json');
+	try {
+		await shellServer.exec(`mkdir -p '${path.posix.dirname(cacheFile)}' && cat > '${cacheFile}' << 'envJSON'
+${JSON.stringify(env, null, '\t')}
+envJSON
+	`);
+	}
+	catch (e) {
+
```
chrmarti, line 723: Should we log this?

src/spec-common/injectHeadless.ts
```diff
@@ -667,14 +668,60 @@ async function patchEtcProfile(params: ResolverParameters, containerProperties:
 	}
 }
 
-async function probeUserEnv(params: { defaultUserEnvProbe: UserEnvProbe; allowSystemConfigChange: boolean; output: Log }, containerProperties: { shell: string; remoteExec: ExecFunction; installFolder?: string; env?: NodeJS.ProcessEnv; shellServer?: ShellServer; launchRootShellServer?: (() => Promise<ShellServer>); user?: string }, config?: CommonMergedDevContainerConfig) {
-	const env = await runUserEnvProbe(params, containerProperties, config, 'cat /proc/self/environ', '\0');
-	if (env) {
+async function probeUserEnv(params: { defaultUserEnvProbe: UserEnvProbe; allowSystemConfigChange: boolean; output: Log; containerSessionDataFolder?: string }, containerProperties: { shell: string; remoteExec: ExecFunction; installFolder?: string; env?: NodeJS.ProcessEnv; shellServer?: ShellServer; launchRootShellServer?: (() => Promise<ShellServer>); user?: string }, config?: CommonMergedDevContainerConfig) {
+	let env = await readUserEnvFromCache(params.containerSessionDataFolder, containerProperties.shellServer);
+	if (env && Object.keys(env).length) {
```
chrmarti, line 673: Is an empty environment not valid?

chrmarti (changes requested):
Looks good, left comments on details I previously missed.

src/spec-common/injectHeadless.ts
eljog: My understanding is that there will be some env variable set in every OS, like `SHELL`?
Here the issue is that, if the user first runs the command with `---default-user-env-probe' none`, the cache file will be saved with empty object `{}`. Next time onwards regardless of the `--default-user-env-probe'` specified, system will by default use the file with empty env.

I think a better solution will be to save the cache to different files depending on the `--default-user-env-probe` passed. Any thoughts?

src/spec-common/injectHeadless.ts
eljog: I think that is a good idea. Th `output` object was not available in this method to emit outputs. But I can certainly check what is needed to get it there.

chrmarti: Sounds good!

src/spec-common/injectHeadless.ts
chrmarti: For `none` it would always be empty? Maybe return early for `none` without writing a file at all?

I would expect the `"userEnvProbe"` value (which can be overwritten in the devcontainer.json) to stay the same over several calls to the CLI during the same startup/session. When the user reloads/reconnects, it would make sense to use a new session folder. (I'm not much opposed to have multiple files, it just doesn't seem to address the issue at its core.) What do you think?

Note that we pass the environment variables to the VS Code Server on startup, so any changes will only take effect when the server is restarted.

eljog: Yes `userEnvProb` will likely stay the same. I was just trying to address it with different files because practically a user of CLI can invoke it with different values each time - but of course not a concern for Codespaces

> When the user reloads/reconnects, it would make sense to use a new session folder. 

Won't discarding the cache on every reload, defeat the purpose? - as what we are trying to make use of the cache to improve the performance on each CLI call.
As you mentioned, for the case of VS Code Server, we only pass the variables on the server startup (not each time we connect to an already running server). So I guess a good point in time to discard the cache will be when we explicitly stop a running server session.

But for this pass, I would leave the cache discarding responsibility to the caller (they could clean the cache directory on their own discretion)

So to summarize, below are the changes I am planning to do now as per the suggestions:
- Return early for `none` without writing a file at all
- Treat empty cache object as valid
- Try to log failures

Let me know if that's okay for this pass.

src/spec-common/injectHeadless.ts
eljog: done

src/spec-common/injectHeadless.ts
eljog: done

src/spec-common/injectHeadless.ts
```diff
@@ -826,6 +827,15 @@ Merged:    ${typeof env.PATH === 'string' ? `'${env.PATH}'` : 'None'}` : ''}`);
 	}
 }
 
+function getUserEnvProb(config: CommonMergedDevContainerConfig | undefined, params: { defaultUserEnvProbe: UserEnvProbe; allowSystemConfigChange: boolean; output: Log; }) {
+	let userEnvProbe = config?.userEnvProbe;
+	params.output.write(`userEnvProbe: ${userEnvProbe || params.defaultUserEnvProbe}${userEnvProbe ? '' : ' (default)'}`);
```
chrmarti, line 832: This would be logged twice now? Maybe pass the value into `runUserEnvProbe`?

src/spec-common/injectHeadless.ts
```diff
@@ -682,54 +687,50 @@ async function probeUserEnv(params: { defaultUserEnvProbe: UserEnvProbe; allowSy
...
 
-async function readUserEnvFromCache(cacheDir?: string, shellServer?: ShellServer) {
-	if (!shellServer || !cacheDir) {
+async function readUserEnvFromCache(userEnvProbe: UserEnvProbe, params: { output: Log; containerSessionDataFolder?: string }, shellServer?: ShellServer) {
+	if (!shellServer || !params.containerSessionDataFolder) {
 		return undefined;
 	}
 
-	const cacheFile = path.posix.join(cacheDir, 'env.json');
+	const cacheFile = path.posix.join(params.containerSessionDataFolder, `${userEnvProbe}.json`);
```
chrmarti, line 701: `env-${userEnvProbe}.json` to make it clearer what it is about?

chrmarti (changes requested):
Looks good. Two nits. Otherwise ready for merging.

src/spec-common/injectHeadless.ts
eljog: done

src/spec-common/injectHeadless.ts
eljog: makes sense. done

chrmarti (approved):
Thanks!

samruddhikhandale (approved):
Thanks, LGTM

Issue #291, opened 2022-11-20
Feature layers should be fully independent


kytans:
Currently features are all put in a single image that is then added in a layer, which separate layers per feature only for running install.sh.

This means that changes to any feature or adding a feature cause a full rebuild of the devcontainer, and also means that creating a variant of a devcontainer with an extra feature will rebuild all the features.

Instead, each feature should be added in a separate layer, which would mean that only features that are listed later than the changed features need to be reran while earlier features are cached.

This would speed up some use cases considerably and is much cleaner.

Issue #313, opened 2022-12-05
`devcontainer build` does not use docker build cache


cwharris:
`devcontainer build` chooses to use a different `--build-context` on each build, forcing `Docker` to build from scratch each time, even if only a single feature was modified.
```
$ devcontainer build ~/dev/project/.devcontainer/ --workspace-folder ~/dev/project/
[10 ms] @devcontainers/cli 0.25.2. Node.js v17.0.1. linux 5.10.0-1050-oem x64.
[12164 ms] Start: Run: docker buildx build
    --load \
    --build-context dev_containers_feature_content_source=/tmp/devcontainercli-charris/container-features/0.25.2-1670206334608 \
    --build-arg _DEV_CONTAINERS_BASE_IMAGE=ubuntu:focal \
    --build-arg _DEV_CONTAINERS_IMAGE_USER=root \
    --build-arg _DEV_CONTAINERS_FEATURE_CONTENT_SOURCE=dev_container_feature_content_temp \
    --target dev_containers_target_stage
    -t vsc-project-de9db83a4d84fadc3ffbe532c095a7c7-features
    -f /tmp/devcontainercli-charris/container-features/0.25.2-1670206334608/Dockerfile.extended
    /tmp/devcontainercli-charris/empty-folder
```

I discovered this while working on a custom feature that depended on multiple time-consuming-to-build features, wasting several hours waiting on builds trying to figure out why no steps of the build were being cached.

This makes devcontainers impractical for me, since I need to develop quite a few features to support the work I'm doing. If each single character change to an install script causes an uncached build of the entire container, developing features becomes infeasible.

https://github.com/devcontainers/cli/blob/0a3e2dc47d1bf2821eec3c55d618cdd36d3c2f58/src/spec-node/utils.ts#L432

What happens if we just remove `Date().now` from tmpFolder name?

chrmarti:
Not sure the new folder plays a role here. We are using a single layer for building the features which is tracked in https://github.com/devcontainers/cli/issues/291. Continuing there. Thanks.

PR #382, opened 2023-01-21
Incrementally copy features near the layer they're installed

trxcllnt:
This PR incrementally copies each feature's contents near the layer they're installed, so rebuilds of a devcontainer take full advantage of the build cache.

To illustrate the behavior I built a devcontainer, edited the source of the last feature in the image, then rebuilt it again. With the public `devcontainer` CLI, the image rebuilds all features even though only the last one changed. With this PR, the second build used the layer cache and started from the layer of the changed feature.

* [build-1-old.log](https://gist.github.com/trxcllnt/8b2ff12f29a434e4d4ff2b45cb2fa823#file-build-1-old-log) (`1m31.803s`)
* [build-2-old.log](https://gist.github.com/trxcllnt/8b2ff12f29a434e4d4ff2b45cb2fa823#file-build-2-old-log) (`1m29.621s`)
* [build-1-new.log](https://gist.github.com/trxcllnt/8b2ff12f29a434e4d4ff2b45cb2fa823#file-build-1-new-log) (`1m27.219s`)
* [build-2-new.log](https://gist.github.com/trxcllnt/8b2ff12f29a434e4d4ff2b45cb2fa823#file-build-2-new-log) (`0m32.794s` :fire: :rocket: :fire: :rocket: :fire: :rocket:)

Fixes https://github.com/devcontainers/cli/issues/291 and https://github.com/devcontainers/cli/issues/313

### DIFF
--- a/src/spec-configuration/containerFeaturesConfiguration.ts
+++ b/src/spec-configuration/containerFeaturesConfiguration.ts
@@ -218,22 +218,29 @@ export function getSourceInfoString(srcInfo: SourceInformation): string {
 }
 
 // TODO: Move to node layer.
-export function getContainerFeaturesBaseDockerFile() {
+export function getContainerFeaturesBaseDockerFile(useBuildKitBuildContexts = false) {
 	return `
 #{featureBuildStages}
 
 #{nonBuildKitFeatureContentFallback}
-
+${(() => {
+	return useBuildKitBuildContexts ? '' : `
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
 USER root
 COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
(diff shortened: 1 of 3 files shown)


trxcllnt:
cc: @chrmarti

jkeech:
@trxcllnt thanks for submitting this PR! This looks like a great improvement. I'd like @chrmarti to review, but just wanted to drop a comment saying it's on our radar and we should be getting to this soon. Thanks for your patience! 🙏

theonlyfoxy:
It seems after changing dockerfile, Features installation is not using cache anymore. does this PR also fix that ?

joshspicer:
Very awesome! Just validated this continues to work as expected when rebased against main.

Validated that changing the contents of the final Feature (in this case a local Feature) preserves all previously cached layers, and the change is picked up 👍 

image

joshspicer (approved):

PR #384, opened 2023-01-23
dereference symlinks when building features and images

trxcllnt:
This PR updates the CLI to dereference symlinks when building features and images.

This makes it easier to build features that share common scripts, .e.g:
```shell
- common
  - init.sh
- src
  - f_1
    - common # link to ../../common
    - install.sh # calls `./common/init.sh`
  - f_2
    - common # link to ../../common
    - install.sh # calls `./common/init.sh`
```


trxcllnt:
An unfortunate side-effect of this PR is symlinks within the feature's local scope are also dereferenced, leading to duplication in the following scenario:

```
- src
  - feature
    - script.sh
    - install.sh # calls `./script`
    - script # soft link to `./script.sh` on the host, but is a copy of `./script.sh` in the image
```

This also breaks things if a user expects a symlink (that may be broken) on the host to be copied into the container (where it may be valid), or otherwise expects soft links to reference container paths:

```
- src
  - feature
    - cuda # links to /usr/local/cuda
    - install.sh
```

With this PR, the above would cause devcontainers to dereference the `cuda` symlink on the host when building the context. 

This seems to be a security risk that could be exploited by a malicious feature, so I'll close this and explore other options.

PR #385, opened 2023-01-25
Update workflows to package and release standalone binaries

trxcllnt:
Opening as a draft so it isn't accidentally merged before #343.

This PR is a follow-up to #343 that updates the GHA workflows to build, test, and release the standalone binaries created by `@vercel/pkg`.

#### Modifications to `dev-containers.yml`:
* Fails fast (can remove this)
* Only run `Test` job when relevant files are modified
* Cache `node_modules` dirs between jobs/across CI runs
* Update `CLI` job to build the standalone binaries on ubuntu-20.04, `macos-11`, and `windows-2019`
* Update `Test` job to run a subset of the tests on the standalone binaries (though can expand this to all tests if preferred!)

[See here](https://github.com/trxcllnt/devcontainers-cli/actions/runs/3992778110/jobs/6848915427) for an example of a successful run.

#### Modifications to `publish-dev-containers.yml`:
* Doesn't publish the npm package if run from a repo other than @devcontainers/cli
* Allow running workflow manually
  * Uses the most recent tag to the current commit if releasing manually
  * Allows maintainers test changes to the `publish-dev-containers.yml` on branches
  * Uploads artifacts to workflow run if `Make GitHub Release` and `Make GitHub Draft Release` aren't selected
* Creates a GitHub release w/ changelog and npm package + standalone binaries (supports draft/pre-releases)

[See here](https://github.com/trxcllnt/devcontainers-cli/actions/runs/3992873007/jobs/6849096101) for an example of a successful release on tag push, and [here](https://github.com/trxcllnt/devcontainers-cli/releases/tag/v0.29.0-trxcllnt) for the release it generated.

#### Notes on cross-compiling
* This PR builds the standalone packages on each target OS
  * My attempts to cross-compile Windows and signed (via `ldid`) MacOS binaries from Ubuntu were unsuccessful
  * [The docs say](https://github.com/vercel/pkg#targets) this should work w/ `--no-bytecode`, but the binaries failed tests on the target OS
  * I can keep investigating this if it's a blocker
* Currently only publishes x86 binaries due to the above, and
  * The GHA runners are x86
  * Windows/MacOS cross-compiling [x86 from arm64 possible, but not arm64 from x86](https://github.com/vercel/pkg#targets)
  * Could solve this with QEMU if we can cross-compile Windows/MacOS binaries from Linux


trxcllnt:
Merging into https://github.com/trxcllnt/devcontainers-cli/tree/fea/pkg (https://github.com/devcontainers/cli/pull/343)

PR #390, opened 2023-01-31
Features contribute lifecycle hooks

joshspicer:
This change implements:
https://github.com/devcontainers/spec/blob/b1d19a5375f667a1bb91e86311bc37c36312996b/proposals/features-contribute-lifecycle-scripts.md

The new contents of the imageMetadata (generated from the [`lifecycle-hooks-advanced`](https://github.com/devcontainers/cli/tree/joshspicer/features-lifecycles-hooks/src/test/container-features/configs/lifecycle-hooks-advanced)) test can be seen below:

image


src/spec-common/injectHeadless.ts
```diff
@@ -412,25 +421,34 @@ async function updateMarkerFile(shellServer: ShellServer, location: string, cont
...
-	return Promise.all((config[`${postCommandName}s`] || []).map(config => runPostCommand(params, containerProperties, config, postCommandName, remoteEnv, doRun)));
+
+async function runLifecycleCommands(params: ResolverParameters, lifecycleCommandOriginMap: LifecycleHooksInstallMap, containerProperties: ContainerProperties, lifecycleHookName: 'onCreateCommand' | 'updateContentCommand' | 'postCreateCommand' | 'postStartCommand' | 'postAttachCommand', remoteEnv: Promise<Record<string, string>>, doRun: boolean) {
+	const commandsForHook = lifecycleCommandOriginMap[lifecycleHookName];
+	if (commandsForHook.length === 0) {
+		return;
+	}
+
+	for (const { command, origin } of commandsForHook) {
+		const displayOrigin = origin ? (origin === 'devcontainer.json' ? origin : `Feature '${origin}'`) : '???'; /// '???' should never happen.
```
samruddhikhandale, line 436: Could origin be `.devcontainer.json`?

samruddhikhandale: Ah! Makes sense --> https://github.com/devcontainers/cli/pull/390/files#diff-fe757ed747d00188273214482dfeb49fe54d578827ab64f39741e886564ad81bR135

src/test/container-features/configs/lifecycle-hooks-advanced/.devcontainer/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu
```
samruddhikhandale, line 1: Curious, why not use an `image` property instead?

src/test/container-features/configs/lifecycle-hooks-advanced/.devcontainer/otter/devcontainer-feature.json
```diff
@@ -0,0 +1,22 @@
+{
+    "id": "otter",
+    "version": "1.2.3",
+    "options": {},
+    "updateContentCommand": "${featureRoot}/helper_script.sh updateContentCommand",
```
samruddhikhandale, line 5: Should we add the `featureRoot` property to the Feature interface?

samruddhikhandale (commented):
Nice, left some comments!

src/test/container-features/configs/lifecycle-hooks-advanced/.devcontainer/Dockerfile
joshspicer: There is no reason. I think I borrowed the skeleton of this test config from an existing one that had a dockerfile.

src/spec-common/injectHeadless.ts
joshspicer: Yea, this is just for logging purposes

src/test/container-features/configs/lifecycle-hooks-advanced/.devcontainer/otter/devcontainer-feature.json
joshspicer: What do you mean by this? It is a part of the spec (https://github.com/devcontainers/spec/pull/196) and will make its way to the documentation, which interface?

src/spec-node/imageMetadata.ts
```diff
@@ -102,14 +113,37 @@ const replaceProperties = [
...
-	postAttachCommands?: (string | string[])[];
+	onCreateCommands?: LifecycleCommand[];
+	updateContentCommands?: LifecycleCommand[];
+	postCreateCommands?: LifecycleCommand[];
+	postStartCommands?: LifecycleCommand[];
+	postAttachCommands?: LifecycleCommand[];
 	shutdownAction?: 'none' | 'stopContainer' | 'stopCompose';
 }
 
+export function lifecycleCommandOriginMapFromMetadata(metadata: ImageMetadataEntry[]): LifecycleHooksInstallMap {
```
chrmarti, line 124: One alternative would be to change MergedDevContainerConfig in `imageMetadata.ts` to remember the origin (the merged config is already passed around).

chrmarti (changes requested):
Left some comments, looks good overall!

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -226,14 +247,14 @@ export function getContainerFeaturesBaseDockerFile() {
 
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
 USER root
-COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /tmp/build-features/
-RUN chmod -R 0700 /tmp/build-features
+COPY --from=dev_containers_feature_content_source {contentSourceRootPath} /opt/build-features/
+RUN chmod -R 0777 /opt/build-features
```
chrmarti, line 251: Should we make it `0755` to avoid making it writable by everyone?

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -23,34 +23,55 @@ export const V1_DEVCONTAINER_FEATURES_FILE_NAME = 'devcontainer-features.json';
 // v2
 export const DEVCONTAINER_FEATURE_FILE_NAME = 'devcontainer-feature.json';
 
-export interface Feature {
+export type Feature = SchemaFeatureBaseProperties & SchemaFeatureLifecycleHooks & DeprecatedSchemaFeatureProperties & InternalFeatureProperties;
+
+export const CONTAINER_BUILD_FEATURES_DIR = '/opt/build-features';
```
chrmarti, line 28: Would `/usr/share/devcontainer/features` be a better match? (Looking at https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard.)

chrmarti (changes requested):

src/spec-configuration/containerFeaturesConfiguration.ts
joshspicer: Hm ok that sounds good.  I also looked at that list but ended up choosing `opt` as it was short and somewhat related - but reading /usr/share that makes more sense 👍

src/test/container-features/configs/lifecycle-hooks-advanced/.devcontainer/otter/devcontainer-feature.json
samruddhikhandale: I think I misunderstood the use of this variable in first instance.

Clarifying question, are we expecting the Feature authors to use `${featureRootFolder}` in `devcontainer-feature.json` file? I think that's not the case from the [spec](https://github.com/devcontainers/spec/pull/196/files#diff-872faa37e7738b0b1885f9c60d4e5bad1462119460f6b4ebbdd95064143ec254R38-R48), but double checking

joshspicer: If it's necessary then yes - https://github.com/devcontainers/spec/pull/196/files#diff-872faa37e7738b0b1885f9c60d4e5bad1462119460f6b4ebbdd95064143ec254R60

It's not required if the lifecycle command doesn't need to "reach into" the Feature (eg: the command is something on the PATH or relative to the workspace folder)

src/spec-node/imageMetadata.ts
```diff
@@ -247,7 +281,20 @@ function collectOrUndefined<T, K extends keyof T>(entries: T[], property: K): No
...
 
-	const raw = featuresConfig?.featureSets.map(featureSet => featureSet.features.map(feature => ({
+	// Variable substitution
+	// eslint-disable-next-line code-no-unused-expressions
+	featuresConfig?.featureSets.forEach(featureSet =>
+		featureSet.features.forEach(f => {
+			pickFeatureLifecycleHookProperties.forEach(hook => {
+				const buildPath = `/opt/build-features/${f.consecutiveId}`;
+				if (f[hook]) {
+					f[hook] = substituteFeatureRoot(f[hook], buildPath);
```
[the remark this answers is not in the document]
joshspicer: After exploring this a bit, wouldn't it make more sense to continue to edit `raw`? The image metadata label stores the `raw`, 

```
[imageMetadataLabel]: JSON.stringify(getDevcontainerMetadata(imageBuildInfo.metadata, config, undefined).raw),
```

Which means that we lose the context of the where on disk the `featureRootFolder` is:

image

${featureRootFolder} is not just the name of the Feature, but something we calculate (`consecutiveId`) - this was done to avoid name collisions.  

Do you still think it makes sense to not embed the path in the metadata.  That seems the most simple, but I could also store the path as a separate element in the metadata?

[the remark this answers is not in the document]
joshspicer: @chrmarti - I've now made a change to not substitute the `raw` metadata, and instead cache an additional property `featureContainerDestFolder` that keeps track of the path within the container that the Feature is cached to.

I don't think there's currently a way around caching the path on the metadata, unless we rearchitect how the paths are crafted (without using the `consecutiveId`.  That change is a bit far-reaching and I think it would be better to leave that to further patch.  What do you think?

image

joshspicer:
An example of `read-configuration` correctly outputting the substituted `mergedConfiguration`

image

src/spec-node/imageMetadata.ts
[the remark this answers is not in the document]
chrmarti: You're right. The updated version is even better. 👍

src/spec-node/imageMetadata.ts
```diff
@@ -102,14 +115,49 @@ const replaceProperties = [
...
+			const substitutedCommand = substituteFeatureRoot(command, entry.featureRootFolder);
+			if (substitutedCommand) {
+				map[hook].push({ origin, command: substitutedCommand });
+			}
+		}
+	}
+	return map;
+}
+
+function mergeLifecycleHooks(metadata: ImageMetadataEntry[], hook: (keyof SchemaFeatureLifecycleHooks)): LifecycleCommand[] | undefined {
```
chrmarti, line 149: Should this include the origin as above? That would change the JSON of the mergedConfiguration for `read-configuration`, so, e.g., Codespaces could use the origin in the UI when running the postAttachCommands.

chrmarti (changes requested):
LGTM, left one comment.

src/spec-node/imageMetadata.ts
joshspicer: I've raised https://github.com/devcontainers/spec/pull/206 to change the devcontainer.json schema to be usable as the result of `mergedConfiguration`.  

After that, I think adding in a generic means to figure out the origin of a piece of the merged config would be useful (not just for lifecycle hooks, but for all the pieces.  That might be worth its own PR, so i'll leave it out here.

joshspicer: And then if there's a generic way we cache the origin of config, we could remove the `lifecycleCommandOriginMapFromMetadata()` function above.

chrmarti (approved):
We should agree on how we want to include the origin of the lifecycle hooks as part of https://github.com/devcontainers/spec/pull/206. IMO we can still merge the work here since it only affects the `mergedConfiguration` which does not need to adhere to a schema in the specification yet. 👍

Issue #381, opened 2023-01-20
Metadata: ContainerEnv does not follow the merge logic when image is built with `devcontainer build`


samruddhikhandale:
Reported in https://github.com/devcontainers/images/issues/368

If a `devcontainer.json`  & `devcontainer-feature.json`  both adds the same `containerEnv`. I would except that the final container has the value set by the `devcontainer.json`. Also, that is how the [spec](https://github.com/devcontainers/spec/blob/2f4062c9669aef8f7ef6bc7950347a0e881a5c36/proposals/image-metadata.md?plain=1#L64) mentions it.

However, for the [java](https://github.com/devcontainers/images/tree/main/src/java)  image (which is built with `devcontainer build` command), the value of `JAVA_HOME`  is set by the [devcontainer-feature.json](https://github.com/devcontainers/features/blob/main/src/java/devcontainer-feature.json#L91) . Instead it should be `/usr/lib/jvm/msopenjdk-current` which is set by the [devcontainer.json](https://github.com/devcontainers/images/blob/main/src/java/.devcontainer/devcontainer.json#L17)

Repro steps: 

1. docker run -it -d mcr.microsoft.com/vscode/devcontainers/java:0.205.19-11
2. docker exec -it ${CONTAINER_ID} bash
3. echo $JAVA_HOME

![image](https://user-images.githubusercontent.com/24955913/213807664-9672bf85-f932-4a08-9a4d-be7b731620ac.png)

ℹ️  Merge logic is followed by the `devcontainer up` command

samruddhikhandale:
@chrmarti Looking at the CLI implementation of why the metadata is available with `up` command but not with `buiild`.

In both cases, the logic of `build` is exactly the same, however, in `up` scenario we have [spawnDevContainer](https://github.com/devcontainers/cli/blob/a53630c47d1a2b9d71dcbf9fd51597fe4bf3eb51/src/spec-node/singleContainer.ts#L324-L411) function which does the extra work of making sure that the dev container has correct metadata. 

[...]

PR #392, opened 2023-02-01
Bug fix: Merge metadata logic for containerEnv - devcontainer build

samruddhikhandale:
Fixes - https://github.com/devcontainers/cli/issues/381

### DIFF
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -15,7 +15,7 @@ import { readLocalFile } from '../spec-utils/pfs';
 import { includeAllConfiguredFeatures } from '../spec-utils/product';
 import { createFeaturesTempFolder, DockerResolverParameters, getCacheFolder, getFolderImageName, getEmptyContextFolder, SubstitutedConfig } from './utils';
 import { isEarlierVersion, parseVersion } from '../spec-common/commonUtils';
-import { getDevcontainerMetadata, getDevcontainerMetadataLabel, getImageBuildInfoFromImage, ImageBuildInfo, ImageMetadataEntry, imageMetadataLabel, MergedDevContainerConfig } from './imageMetadata';
+import { getContainerEnvMetadata, getDevcontainerMetadata, getDevcontainerMetadataLabel, getImageBuildInfoFromImage, ImageBuildInfo, ImageMetadataEntry, imageMetadataLabel, MergedDevContainerConfig } from './imageMetadata';
 import { supportsBuildContexts } from './dockerfileUtils';
 import { ContainerError } from '../spec-common/errors';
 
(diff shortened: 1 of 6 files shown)


samruddhikhandale:
The following tests are failing https://github.com/devcontainers/cli/actions/runs/4069380934/jobs/7009165123#step:7:4476
since quite a few days and is not related to this PR changes

Fetching a manifest is failing with the following error 👇 

> Did not fetch manifest: {“errors”:[{“code”:“MANIFEST_UNKNOWN”,“message”:“OCI index found, but accept header does not support OCI indexes”}]}

Looking at how to fix it!

src/spec-node/containerFeatures.ts
```diff
@@ -294,6 +294,14 @@ async function getFeaturesBuildOptions(params: DockerResolverParameters, devCont
 		.replace('#{copyFeatureBuildStages}', getCopyFeatureBuildStages(featuresConfig, buildStageScripts))
 		.replace('#{devcontainerMetadata}', getDevcontainerMetadataLabel(imageMetadata, common.experimentalImageMetadata))
 		;
+	
+	const rawContainerEnv = imageMetadata.raw.find(entry => entry.containerEnv)?.containerEnv;
```
chrmarti, line 298: Use `imageMetadata.config`, that has variables like `${remoteWorkspaceFolder}` replaced. (`imageMetadata.raw` comes without these variables replaced.)

src/spec-node/containerFeatures.ts
```diff
@@ -294,6 +294,14 @@ async function getFeaturesBuildOptions(params: DockerResolverParameters, devCont
...
+	if (rawContainerEnv !== undefined) {
+		let keys = Object.keys(rawContainerEnv);
+		const concatenatedEnv = keys.map(k => `ENV ${k}=${rawContainerEnv![k]}`).join('\n');
+		dockerfile += concatenatedEnv;
```
chrmarti, line 302: Maybe add a placeholder in `getContainerFeaturesBaseDockerFile()` and add a `replace` line above instead to use the same pattern as above?

src/spec-node/containerFeatures.ts
chrmarti: `imageMetadata` actually includes the features for which we already apply `containerEnv`. Can you try using `devContainerConfig.config.containerEnv`?

chrmarti (changes requested):
Makes sense. Should we remove `containerEnv` from `up` then?

chrmarti:
I just now see: We still need `containerEnv` in `up` for the cases when we can avoid building a custom image (`getExtendImageBuildInfo` then skips `getFeaturesBuildOptions`).

src/spec-node/containerFeatures.ts
samruddhikhandale: > Can you try using devContainerConfig.config.containerEnv?

Works 👍

src/spec-node/containerFeatures.ts
samruddhikhandale: Updated

chrmarti (approved):
LGTM!

PR #393, opened 2023-02-02
[oci] Support querying registries that Accept `application/vnd.oci.image.index.v1+json`

joshspicer:
This fixes two failing tests `inspect image in docker.io` and `inspect image in mcr.microsoft.com` that began failing on `main`.

The fix is to implement parsing the image index media type,
https://github.com/opencontainers/image-spec/blob/main/image-index.md.  Before this change, something had changed on the dockerhub/mcr side to reject the Accept header we had been sending and expecting as a response:

image

Note that Features/Templates are uneffected since we have our own mediaTypes and expected format.


chrmarti (approved):
Great, thanks!

PR #397, opened 2023-02-03
Add label for config file

chrmarti:
This is to distinguish containers built from the same workspace folder, but with different devcontainer.json files.

Codespaces and the Dev Containers extension always pass in the label to be used, this change does not change their behavior. This change is to replicate the Dev Containers extension's labels here, so they match when the OSS CLI is used standalone.

Codespaces has only one dev container per Docker daemon (the scope of labels) and should not be affected by these considerations.

### DIFF
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -9,12 +9,12 @@ import yargs, { Argv } from 'yargs';
 import * as jsonc from 'jsonc-parser';
 
 import { createDockerParams, createLog, experimentalImageMetadataDefault, launch, ProvisionOptions } from './devContainers';
-import { SubstitutedConfig, createContainerProperties, createFeaturesTempFolder, envListToObj, inspectDockerImage, isDockerFileConfig, SubstituteConfig, addSubstitution } from './utils';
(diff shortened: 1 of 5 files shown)


src/spec-node/devContainersSpecCLI.ts
```diff
@@ -193,7 +193,7 @@ async function provision({
 	const addRemoteEnvs = addRemoteEnv ? (Array.isArray(addRemoteEnv) ? addRemoteEnv as string[] : [addRemoteEnv]) : [];
 	const addCacheFroms = addCacheFrom ? (Array.isArray(addCacheFrom) ? addCacheFrom as string[] : [addCacheFrom]) : [];
 	const additionalFeatures = additionalFeaturesJson ? jsonc.parse(additionalFeaturesJson) as Record<string, string | boolean | Record<string, string | boolean>> : {};
-	const idLabels = idLabel ? (Array.isArray(idLabel) ? idLabel as string[] : [idLabel]) : getDefaultIdLabels(workspaceFolder!);
+	const providedIdLabels = idLabel ? Array.isArray(idLabel) ? idLabel as string[] : [idLabel] : undefined;
```
samruddhikhandale, line 196: Current description of `id-label` -
> If no --id-label is given, one will be inferred from the --workspace-folder path.

In what scenario would `providedIdLabels` be undefined? Won't it conflict with the description (that there will be some inference and something will be set for id-label). Or does inferring refer to a possibility of being `undefined`?

samruddhikhandale (approved):
Looks good!

PR #411, opened 2023-02-10
add pre-commit formatting hook

joshspicer:
Adds a `pre-commit` hook with `husky` to enforce code style/linting.

When create a commit, the `.husky/pre-commit` script will run in `--fix` mode.  If something cannot be auto-fixed, the commit will fail.

image

Since `yarn lint` is required to pass by CI, no commit should ever be pushed without passing this check.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -14,8 +14,7 @@
 		// Keep command history across instances
 		"source=dev-containers-cli-bashhistory,target=/home/node/commandhistory"
 	],
-
-	"postCreateCommand": "yarn install",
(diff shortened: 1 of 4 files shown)


samruddhikhandale (commented):
This is super awesome! 💡  However, I wasn't able to take advantage of this hook unless I executed `git add .husky/pre-commit` in the terminal.

image

Not sure if contributors will know that they need to add `git add .husky/pre-commit`, hence, can we add this to dev config? 🤔

joshspicer:
Ah, looks like I missed a step.  I've changed it up a bit - let me know what you think.

samruddhikhandale (approved):
🚀

Issue #413, opened 2023-02-15
devcontainer should return non-zero exit code if it fails


akhildevelops:
Hi,

`devcontainer features test -f color; echo $?` returns `0` though `/workspaces/devcontainer-features/src/color` is not present. It displays the error but returns zero.

It's difficult to catch issues in [CI Pipeline](https://github.com/akhildevelops/devcontainer-features/actions/runs/4183819226/jobs/7248666682). Ran against https://github.com/akhildevelops/devcontainer-features/tree/28e76f09b7ca3e78e18a47dd82ffaa965f4c7f03

samruddhikhandale:
Hi 👋 

Thanks for reporting, that's definitely a bug. I'll look into fixing it.

samruddhikhandale:
Closing as fix is merged https://github.com/devcontainers/cli/pull/418 and will be included in the next release!

PR #418, opened 2023-02-17
Fix -`devcontainer feature test` cmd should fail with non-zero exit code if Feature's sub-folder does not exist

samruddhikhandale:
Addresses - https://github.com/devcontainers/cli/issues/413

image


joshspicer (approved):
Makes sense to me!

PR #424, opened 2023-02-27
Make --workspace-folder optional, default to ".".  Fixes #29

### LINKED ISSUE #29: Why Is `--workspace-folder` Required or Default to $PWD?

alexec:
Fixes #29

I don't know how to test this, and ironically I created this PR in Codespaces, but I cannot run tests:

image

Please advise.


MunsMan:
Isn't that similar to #332?

Should we check if a `.devcontainer` file exists in the current directory?

alexec:
Exactly the same, to the byte. I don't think the check needs to be smart.

PR #427, opened 2023-02-28
fix: environment variables with spaces are not supported

dhhyi:
When using environment variables with spaces in `containerEnv`, the rendered Dockerfile will not be valid, because the spaces are not escaped.

i.e. using
```json
    "containerEnv": {
        "ENV_WITH_SPACES": "Environment variable with spaces"
    }
```

fails with:
```
Syntax error - can't find = in "variable". Must be of the form: name=value
```

because the following is inserted in the Dockerfile:
```Dockerfile
ENV ENV_WITH_SPACES=Environment variable with spaces
```

but it should be:
```Dockerfile
ENV ENV_WITH_SPACES=Environment\ variable\ with\ spaces
```

### DIFF
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -496,7 +496,7 @@ export function getContainerEnvMetadata(containerEnv: Record<string, string> | u
 		return '';
 	}
 
-	const keys = Object.keys(containerEnv);
-	const concatenatedEnv = keys.map(k => `ENV ${k}=${containerEnv![k]}`).join('\n');
-	return concatenatedEnv;
+	return Object.entries(containerEnv)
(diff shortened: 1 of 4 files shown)


dhhyi:

dhhyi:
@joshspicer PR #455 replaces this, correct? Please close if true. 
I can't really figure out the CI error...

dhhyi:
> Hey @dhhyi thanks for the PR! Yea, this looks like an alternative approach to my PR.  Not sure which one is superior (if either..).

@joshspicer let's go with double quotes. It looks more stable.

But please have a look at the test cases I added. I liked the one with the `eval` on an environment variable 😅

joshspicer:
Thank you, I will graciously steal those tests for my PR :)

dhhyi:
Sure, go ahead 😉👍

Issue #59, opened 2022-06-23
Would like to be able to execute interactive command


hpe-ykoehler:
In my env, developer will likely want to run devcontainer exec bash and then be able to navigate the container and do stuff within using command-line interface.

As such, I think the option to exec the command in interactive mode is important.

hpe-ykoehler:
When I use the devcontainer exec of today to run "bash" I cannot type anything, the output shows the prompt and the container is not accepting inputs.

That is normally because when using docker, to run bash we must use -ti to make the docker knows we require a TTY and tha that the session will be interactive so that it connects the stdin to the container (or someting along those line).

So the test is "devcontainer exec bash" and see that you get a functional terminal, which in my setup I do not get.

Nking92:
+1 to this request. There are devs I work with who don't use VSCode as their editor but would still find devcontainers useful in their workflow. An interactive shell is essential for them.

In the meantime, it's possible to open a terminal session by using `docker container ls` to get the running container ID, then run bash with `docker run -it <id> /bin/bash`


```
nk@DESKTOP-LIH0TIU:~/code/vscode-remote-try-rust$ docker container ls
CONTAINER ID   IMAGE                                                             COMMAND                  CREATED         STATUS             PORTS     NAMES
d5b67fea7ce8   vsc-vscode-remote-try-rust-d007786a17c62e0d1a94f4da56714892-uid   "/bin/sh -c 'echo Co…"   5 minutes ago   Up 5 minutes                 kind_brahmagupta

nk@DESKTOP-LIH0TIU:~/code/vscode-remote-try-rust$ docker exec -it d5b67fea7ce8 /bin/bash
root ➜ / $
```

zflat:
This works OK from within a workspace directory:

```bash
docker exec -it -u $(id -u) $(devcontainer up --workspace-folder . | jq -r .containerId) bash
```
It will get the ID of a container that was previously started from the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) VSCode extension.

mscottford:
> This works OK from within a workspace directory:
> 
> ```shell
> docker exec -it -u $(id -u) $(devcontainer up --workspace-folder . | jq -r .containerId) bash
> ```
> 
> It will get the ID of a container that was previously started from the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) VSCode extension.

[...]

PR #429, opened 2023-03-02
Connect stdin to execed process

chrmarti:
#59


src/spec-common/injectHeadless.ts
```diff
@@ -548,38 +552,86 @@ export function createFileCommand(location: string) {
...
+			if (stdin.isTTY) {
+				stdin.setRawMode(true);
+			}
+			stdin.on('data', listener);
+			sub = { dispose: () => stdin.off('data', listener) };
 		}
-		if (options.resolveOn && options.resolveOn.exec(cmdOutput)) {
-			doResolveEarly();
+	} else {
+		const p = pp = await remoteExec({
```
joshspicer, line 590: why do you need to assign this twice?

joshspicer (approved):

src/spec-common/injectHeadless.ts
chrmarti: `pp` has a wider type than `p`. `p` is used inside the else with specific properties while `pp` is also assigned in the "then" block and used outside the if-then-else with less specific properties (the `exit` property).

PR #436, opened 2023-03-06
Doc: add shell highlighting to markdown

jarrodcolburn:
Just added a single word 'shell' to markdown to enable syntax highlighting. For comparison...

With ("shell" in markdown):
```shell
#!/bin/bash

set -e

# Import test library for `check` command
source dev-container-features-test-lib

check "Oryx version" oryx --version
check "Dotnet is not removed if it is not installed by the Oryx Feature" dotnet --version
...
```

Without:
```
#!/bin/bash

set -e

# Import test library for `check` command
source dev-container-features-test-lib

check "Oryx version" oryx --version
check "Dotnet is not removed if it is not installed by the Oryx Feature" dotnet --version
...
```

Note: If you're not noticing the difference, it's more apparent on dark themes


joshspicer (approved):
Thanks!

Issue #434, opened 2023-03-03
Cannot pull features from private azure container registry


Clockwork-Muse:
I'm currently unable to pull features from a private azure container registry when building containers.

I believe the accounts used should already have the proper permissions; at minimum, base images can be successfully pulled and used for the devcontainer.

The devcontainer is trivial
```json
{
    "name": "Sample",
    "image": "private.azurecr.io/alpine",
    "features": {
        "private.azurecr.io/devcontainers/features/hello:1": {}
    },
}
```

```log
$devcontainer build /tmp/somedir/ --log-level trace
devcontainer build /tmp/somedir/ --log-level trace
(node:274018) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `code --trace-deprecation ...` to show where the warning was created)
[11 ms] @devcontainers/cli 0.31.0. Node.js v16.14.2. linux 5.19.0-32-generic x64.
[11 ms] Start: Run: docker buildx version
[73 ms] Stop (62 ms): Run: docker buildx version
[74 ms] github.com/docker/buildx v0.10.2 00ed17d
[74 ms] 
[76 ms] Start: Run: docker inspect --type image private.azurecr.io/alpine
[114 ms] Stop (38 ms): Run: docker inspect --type image private.azurecr.io/alpine
[114 ms] Start: Run: docker inspect --type image private.azurecr.io/alpine
[152 ms] Stop (38 ms): Run: docker inspect --type image private.azurecr.io/alpine
[152 ms] local container features stored at: /home/simhoff/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/node_modules/vscode-dev-containers/container-features
[153 ms] Start: Run: tar --no-same-owner -x -f -
[167 ms] Stop (14 ms): Run: tar --no-same-owner -x -f -
[167 ms] workspace root: /tmp/somedir
[171 ms] Loaded /home/simhoff/.vscode/extensions/ms-vscode-remote.remote-containers-0.282.0/dist/node_modules/vscode-dev-containers/container-features/devcontainer-features.json, which declares 25 features and does not contain explicit source info.
[171 ms] --- Processing User Features ----
[171 ms] configPath: /tmp/somedir/.devcontainer/devcontainer.json
[171 ms] * Processing feature: private.azurecr.io/devcontainers/features/hello:1
[...]

PR #444, opened 2023-03-14
[oci registry impl] Support `az acr login -n <REGISTRY>`

joshspicer:
ref: https://github.com/devcontainers/cli/issues/434

Use the `identitytoken` added to the standard `~/.docker/config.json` by some tools 

Notably, the result of `az acr login -n <REGISTRY>` command writes the `identitytoken` property, which is expected to be used as a refresh_token.  See the following capture from how oras performs this exchange.

image

This may be more useful generally, but the only registry we've encountered that expects this is the Azure Container Registry.

Below demonstrates fetching a Feature's manifest by using the auth outputted by `az acr login`.

image

In an example project:
image


samruddhikhandale (approved):
Looks great!

chrmarti (approved):

Issue #409, opened 2023-02-10
[feature] remove test containers after every run


akhildevelops:
I :heart: the tool thanks for sharing it.

I observed that test containers don't stop and removed after every completion of tests.

After a successful run:
```
🏃 Starting test(s)...

🧪 Starting 'android-cli' tests...
🧪 Executing scenarios for feature 'android-cli'...
⚠️ No scenario file found at '/workspaces/devcontainer-features/test/android-cli/scenarios.json'. Skipping...



  ================== TEST REPORT ==================
✅ Passed:      'android-cli'
```
I observe the test containers still run:
```
node ➜ /workspaces/devcontainer-features/test/color (main) $ docker ps
CONTAINER ID   IMAGE                                                         COMMAND                  CREATED         STATUS         PORTS     NAMES
32aa248c7dc0   vsc-1676027268210-864360691c93743c8b0c629ee643756c-features   "/bin/sh -c 'echo Co…"   9 seconds ago   Up 7 seconds             flamboyant_moore
```

PR #450, opened 2023-03-15
[features test cmd] Automatically clean up test containers

joshspicer:
ref: https://github.com/devcontainers/cli/issues/409


src/spec-node/featuresCLI/test.ts
```diff
@@ -20,6 +20,7 @@ export function featuresTestOptions(y: Argv) {
 			'remote-user': { type: 'string', alias: 'u', describe: 'Remote user.  Not used for scenarios.', },  // TODO: Optionally replace 'scenario' configs with this value?
 			'log-level': { choices: ['info' as 'info', 'debug' as 'debug', 'trace' as 'trace'], default: 'info' as 'info', description: 'Log level.' },
 			'skip-image-metadata': { type: 'boolean', default: false, description: 'Skip applying image metadata to the resultant test image.  Image metadata is potentially appended to base images built with the dev container CLI' },
+			'preserve-test-containers': { type: 'boolean', default: false, description: 'Do not remove test containers after running tests.' },
```
samruddhikhandale, line 23: Can we validate this flag with a test?

samruddhikhandale, line 23: Also, can we check in either of https://github.com/devcontainers/cli/blob/e3d12b3157cc2ec74805d6764f67ad461d49ecc1/src/test/container-features/featuresCLICommands.test.ts#L25 tests, that the containers are getting cleaned up as expected.

src/spec-node/featuresCLI/testCommandImpl.ts
```diff
@@ -66,10 +67,27 @@ export async function doFeaturesTestCommand(args: FeaturesTestCommandInput): Pro
...
+		await cleanup(cliHost);
+	}
+
 	// Pretty-print test results and exit with 0 or 1 exit code.
 	return analyzeTestResults(testResults);
 }
 
+async function cleanup(cliHost: CLIHost) {
+	// Delete any containers that have the 'devcontainer.is_test_run=true' label set.
+	const filterForContainerIdArgs = ['ps', '-a', '--filter', 'label=devcontainer.is_test_run=true', '--format', '{{.ID}}'];
```
samruddhikhandale, line 81: Let's say, I run tests with `preserve-test-containers` flag, then for the second time I run tests without it.

Won't this command also clean up containers from first run? 
I think this is okay for now, but thought of pointing it out.

chrmarti (approved):

src/spec-node/featuresCLI/testCommandImpl.ts
joshspicer: Yep, it will clean everything up.  I don't think that will cause any problems, it might actually be nice that subsequent runs do the clean up for you unless you keep on preserving.  I imagine this flag being used as you're manually iterating on a single test.

Issue #448, opened 2023-03-15
[question] How can I pass external args to tests?


eddumelendez:
My feature depends on a secret. For that reason, I can not put it in plain text. Is there currently a way to pass an external args? In GHA, would be easy to set using `devcontainer features test -f secret-feature --args ${{ secrets.MY_TOKEN }}`

joshspicer:
Are you looking to be able to add remoteEnv variables to a container via the command line when testing? (ie: Should these be environment variables in the running container?)

That's not something we have wired up in the test command currently, but seems like a useful feature

joshspicer:
I could imagine a `remoteEnv` section here https://github.com/devcontainers/cli/blob/main/src/spec-node/featuresCLI/testCommandImpl.ts#L322

eddumelendez:
yes, please that would be awesome! and probably related to this other question https://github.com/devcontainers/feature-starter/issues/41

PR #451, opened 2023-03-15
[features test cmd] Add --remote-env flag

joshspicer:
ref: https://github.com/devcontainers/cli/issues/448

Provide a way for users to pass environment variables when invoking the test command (using the standardized `--remote-env` flag).  This is particularly useful now that Features can contribute lifecycle scripts, so things like secrets can be feasibly utilized by Features.

See the regression test for how this would be used.


chrmarti (changes requested):
For secrets we might want to follow what we come up with for `up` and other commands. (E.g., this could be `--secret-env-file`.) Are there other use cases for `--remote-env`?

joshspicer:
That's fair - perhaps we put this on the backburner until we have a solution for passing secrets to `up`.

PR #452, opened 2023-03-15
Update docs on running tests

joshspicer:


package.json
```diff
@@ -40,7 +40,7 @@
 		"clean": "npm-run-all clean-dist clean-built",
 		"clean-dist": "rimraf dist",
 		"clean-built": "rimraf built",
-		"test": "env TS_NODE_PROJECT=src/test/tsconfig.json mocha -r ts-node/register --exit src/test/*.test.ts",
+		"test": "env TS_NODE_PROJECT=src/test/tsconfig.json mocha -r ts-node/register --exit src/test/**/*.test.ts",
```
samruddhikhandale, line 43: Why do we need to change this? Aren't the generic tests in https://github.com/devcontainers/cli/tree/main/src/test?

src/test/**/*.test.ts --> would refer to [src/test/container-features](https://github.com/devcontainers/cli/tree/main/src/test/container-features) & [src/test/container-templates](https://github.com/devcontainers/cli/tree/main/src/test/container-templates), don't they already have separate commands (test-container-features & test-container-templates)?

PR #453, opened 2023-03-16
Ability to skip storing remoteEnv from devcontainer config on container metadata label

eljog:
Ability to skip storing remoteEnv from devcontainer config on container metadata label, by using the new option `skip-persisting-remote-env-from-config` for the `up` command.

This will help users to avoid putting sensitive variables/secrets on the image label.

### DIFF
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -131,7 +131,7 @@ export async function getExtendImageBuildInfo(params: DockerResolverParameters, 
 		if (canAddLabelsToContainer && !imageBuildInfo.dockerfile) {
 			return {
 				labels: {
-					[imageMetadataLabel]: JSON.stringify(getDevcontainerMetadata(imageBuildInfo.metadata, config, undefined, [], getOmitDevcontainerPropertyOverride(params.common.skipPersistingRemoteEnvFromConfig)).raw),
+					[imageMetadataLabel]: JSON.stringify(getDevcontainerMetadata(imageBuildInfo.metadata, config, undefined, [], getOmitDevcontainerPropertyOverride(params.common)).raw),
 				}
 			};
 		}
@@ -224,7 +224,7 @@ function getImageBuildOptions(params: DockerResolverParameters, config: Substitu
 		dstFolder,
 		dockerfileContent: `
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_target_stage
-${getDevcontainerMetadataLabel(getDevcontainerMetadata(imageBuildInfo.metadata, config, { featureSets: [] }, [], getOmitDevcontainerPropertyOverride(params.common.skipPersistingRemoteEnvFromConfig)), params.common.experimentalImageMetadata)}
+${getDevcontainerMetadataLabel(getDevcontainerMetadata(imageBuildInfo.metadata, config, { featureSets: [] }, [], getOmitDevcontainerPropertyOverride(params.common)), params.common.experimentalImageMetadata)}
 `,
 		overrideTarget: 'dev_containers_target_stage',
 		dockerfilePrefixContent: `${syntax ? `# syntax=${syntax}` : ''}
@@ -237,8 +237,8 @@ ${getDevcontainerMetadataLabel(getDevcontainerMetadata(imageBuildInfo.metadata, 
 	};
 }
 
-function getOmitDevcontainerPropertyOverride(shouldOmit: boolean | undefined): (keyof DevContainerConfig & keyof ImageMetadataEntry)[] {
-	if (shouldOmit) {
+function getOmitDevcontainerPropertyOverride(resolverParams: { omitConfigRemotEnvFromMetadata?: boolean }): (keyof DevContainerConfig & keyof ImageMetadataEntry)[] {
+	if (resolverParams.omitConfigRemotEnvFromMetadata) {
 		return ['remoteEnv'];
 	}
 
@@ -281,7 +281,7 @@ async function getFeaturesBuildOptions(params: DockerResolverParameters, devCont
 	const buildContentImageName = 'dev_container_feature_content_temp';
 
 	const omitPropertyOverride = params.common.skipPersistingCustomizationsFromFeatures ? ['customizations'] : [];
-	const imageMetadata = getDevcontainerMetadata(imageBuildInfo.metadata, devContainerConfig, featuresConfig, omitPropertyOverride, getOmitDevcontainerPropertyOverride(params.common.skipPersistingRemoteEnvFromConfig));
+	const imageMetadata = getDevcontainerMetadata(imageBuildInfo.metadata, devContainerConfig, featuresConfig, omitPropertyOverride, getOmitDevcontainerPropertyOverride(params.common));
 	const { containerUser, remoteUser } = findContainerUsers(imageMetadata, composeServiceUser, imageBuildInfo.user);
 	const builtinVariables = [
 		`_CONTAINER_USER=${containerUser}`,
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -129,10 +129,9 @@ export async function getExtendImageBuildInfo(params: DockerResolverParameters, 
 	const featuresConfig = await generateFeaturesConfig({ ...params.common, platform }, dstFolder, config.config, getContainerFeaturesFolder, additionalFeatures);
 	if (!featuresConfig) {
 		if (canAddLabelsToContainer && !imageBuildInfo.dockerfile) {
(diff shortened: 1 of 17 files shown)


src/spec-node/devContainersSpecCLI.ts
```diff
@@ -122,6 +122,7 @@ function provisionOptions(y: Argv) {
 		'dotfiles-install-command': { type: 'string', description: 'The command to run after cloning the dotfiles repository. Defaults to run the first file of `install.sh`, `install`, `bootstrap.sh`, `bootstrap`, `setup.sh` and `setup` found in the dotfiles repository`s root folder.' },
 		'dotfiles-target-path': { type: 'string', default: '~/dotfiles', description: 'The path to clone the dotfiles repository to. Defaults to `~/dotfiles`.' },
 		'container-session-data-folder': { type: 'string', description: 'Folder to cache CLI data, for example userEnvProbe results' },
+		'skip-persisting-remote-env-from-config': { type: 'boolean', default: false, description: 'Do not persist remoteEnv from devcontainer.json in container metatdata label' },
```
jkeech, line 125: Typo
```suggestion
		'skip-persisting-remote-env-from-config': { type: 'boolean', default: false, description: 'Do not persist remoteEnv from devcontainer.json in container metadata label' },
```

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -545,6 +549,7 @@ async function doBuild({
 			skipPostAttach: true,
 			experimentalImageMetadata,
 			skipPersistingCustomizationsFromFeatures: skipPersistingCustomizationsFromFeatures,
+			skipPersistingRemoteEnvFromConfig: false,
```
jkeech, line 552: Why hardcode this here and in other places instead of using `skipPersistingRemoteEnvFromConfig` from the command line arg? It seems like if you build an image separately, you'd still want this flag honored.

jkeech (approved):
Overall looks pretty good, but I left a couple suggestions. It would be good for @joshspicer and @chrmarti to also review.

src/spec-node/devContainersSpecCLI.ts
eljog: From the Codespaces perspective we only create containers with `up` command, so I have added the commandline option only for the `up` command. But the downstream code is shared by most commands and the downstream code expects this param.
A simple alternative would have been to make this flag optional `skipPersistingRemoteEnvFromConfig? : boolean` in the downstream, so that we don't have to define them with `false` value everywhere else. I was just following the pattern @joshspicer established for `skipPersistingCustomizationsFromFeatures`.

eljog: On a second thought, I felt it is better to cleanup the code and avoid the need for setting default value in other commands.
Done

src/spec-node/containerFeatures.ts
```diff
@@ -129,9 +129,10 @@ export async function getExtendImageBuildInfo(params: DockerResolverParameters,
 	const featuresConfig = await generateFeaturesConfig({ ...params.common, platform }, dstFolder, config.config, getContainerFeaturesFolder, additionalFeatures);
 	if (!featuresConfig) {
 		if (canAddLabelsToContainer && !imageBuildInfo.dockerfile) {
+			const omitDevcontainerPropertyOverride: (keyof DevContainerConfig & keyof ImageMetadataEntry)[] = params.common.skipPersistingRemoteEnvFromConfig ? ['remoteEnv'] : [];
 			return {
 				labels: {
-					[imageMetadataLabel]: JSON.stringify(getDevcontainerMetadata(imageBuildInfo.metadata, config, undefined).raw),
+					[imageMetadataLabel]: JSON.stringify(getDevcontainerMetadata(imageBuildInfo.metadata, config, undefined, [], omitDevcontainerPropertyOverride).raw),
```
chrmarti, line 135: Two more cases of calls to `getDevcontainerMetadata()` that need the same treatment:
- No features, with Dockerfile: https://github.com/devcontainers/cli/blob/e8493e11c6a920d68fd9b339dbcb7ac77fbf3abf/src/spec-node/containerFeatures.ts#L228
- With features (with or without Dockerfile): https://github.com/devcontainers/cli/blob/e8493e11c6a920d68fd9b339dbcb7ac77fbf3abf/src/spec-node/containerFeatures.ts#L293

Also add tests for these.

I found that following references to `imageMetadataLabel` (from `imageMetadata.ts`) is a good way to verify completeness when reviewing.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -122,6 +122,7 @@ function provisionOptions(y: Argv) {
 		'dotfiles-install-command': { type: 'string', description: 'The command to run after cloning the dotfiles repository. Defaults to run the first file of `install.sh`, `install`, `bootstrap.sh`, `bootstrap`, `setup.sh` and `setup` found in the dotfiles repository`s root folder.' },
 		'dotfiles-target-path': { type: 'string', default: '~/dotfiles', description: 'The path to clone the dotfiles repository to. Defaults to `~/dotfiles`.' },
 		'container-session-data-folder': { type: 'string', description: 'Folder to cache CLI data, for example userEnvProbe results' },
+		'skip-persisting-remote-env-from-config': { type: 'boolean', default: false, description: 'Do not persist remoteEnv from devcontainer.json in container metadata label' },
```
chrmarti, line 125: Add `hidden: true`, so we don't propagate the use of this workaround.

`persisting` is maybe to general (we still write the env cache in the container - is that an issue?). `omit-config-remote-env-from-metadata` may be clearer?

src/spec-node/imageMetadata.ts
```diff
@@ -299,7 +300,7 @@ export function getDevcontainerMetadata(baseImageMetadata: SubstitutedConfig<Ima
 	const raw = [
 		...baseImageMetadata.raw,
 		...featureRaw,
-		pick(devContainerConfig.raw, pickConfigProperties),
+		pick(devContainerConfig.raw, effectivePickDevcontainerProperties),
```
chrmarti, line 303: Also use `effectivePickDevcontainerProperties` below for `config` for consistency. (It doesn't change the behavior in the cases we are looking at.)

chrmarti (changes requested):

src/spec-node/devContainersSpecCLI.ts
joshspicer: I prefer 'omit' as well

src/spec-node/containerFeatures.ts
eljog: Great catch. Done

src/spec-node/containerFeatures.ts
```diff
@@ -129,9 +129,10 @@ export async function getExtendImageBuildInfo(params: DockerResolverParameters,
 	const featuresConfig = await generateFeaturesConfig({ ...params.common, platform }, dstFolder, config.config, getContainerFeaturesFolder, additionalFeatures);
 	if (!featuresConfig) {
 		if (canAddLabelsToContainer && !imageBuildInfo.dockerfile) {
+			const omitDevcontainerPropertyOverride: (keyof DevContainerConfig & keyof ImageMetadataEntry)[] = params.common.skipPersistingRemoteEnvFromConfig ? ['remoteEnv'] : [];
```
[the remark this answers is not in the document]
eljog: Yes. 
The pattern I see is that we use object destructuring generally when the inner properties are accessed more than once

src/spec-node/devContainersSpecCLI.ts
eljog: Sounds good. Updated

src/spec-node/imageMetadata.ts
eljog: Changes made and the behavior seems the same :+1:
With the change, the method will omit `remoteEnv` in both `raw` and `metadata` properties in the response
Unit test has been updated [here (line 261)](https://github.com/devcontainers/cli/pull/453/files/bc78df5bed179fcbba1136933dddaffdf5a3c1b3..93c235bf636e37468094ad75aa203e4aab12edfd#diff-50eea78b7e48891b271d364f69b5c80f216c1cccf1d77affa79499af4d5bc1d4)

src/test/configs/dockerfile-without-features/.devcontainer.json
```diff
@@ -0,0 +1,20 @@
+{
+	"build": {
+		"dockerfile": "Dockerfile",
+		"args": {
+			"VARIANT": "16-bullseye"
+		}
+	},
+	// "postCreateCommand": [
```
joshspicer, line 8: Remove commented config.

joshspicer (dismissed):

src/test/configs/dockerfile-without-features/.devcontainer.json
eljog: oops, fixed

joshspicer (approved):

chrmarti (approved):

Issue #454, opened 2023-03-16
Syntax error - can't find = in "-a." with features enabled


xec-abailey:
Howdy!

This morning I logged in as usual and rebuilt my dev container but received the following error: 

```
ERROR: failed to solve: rpc error: code = Unknown desc = 
failed to solve with frontend dockerfile.v0: failed to solve with 
frontend gateway.v0: rpc error: code= Unknown desc = 
Syntax error - can't find = in "-a". Must be of the form: name=value
```

After a bit of messing around with it (trying again using a local dev container to make sure that it wasn't caused by codespaces for some reason) I removed "features" from my devcontainer and attempted to rebuild and it worked fine

See [this thread](https://github.com/orgs/community/discussions/50251) for others that are experiencing this issue

joshspicer:
Given the report above, it seems to be a case of not escaping containerEnv properly.  It was likely introduced in https://github.com/devcontainers/cli/pull/392/files#diff-fe757ed747d00188273214482dfeb49fe54d578827ab64f39741e886564ad81bR453

jkeech:
The change in #392 needs to be updated to properly quote and escape env var values as described at https://docs.docker.com/engine/reference/builder/#env. It looks like it's not quoting and escaping today, and spaces in the ENV values is now causing it to break.

PR #455, opened 2023-03-16
Escape and enclose containerEnv in quotes when writing to Dockerfile

joshspicer:
ref: https://github.com/devcontainers/cli/issues/454


samruddhikhandale (dismissed):
Nice!

src/spec-node/imageMetadata.ts
```diff
@@ -494,8 +494,11 @@ export function getContainerEnvMetadata(containerEnv: Record<string, string> | u
 	if (!containerEnv) {
 		return '';
 	}
-
 	const keys = Object.keys(containerEnv);
-	const concatenatedEnv = keys.map(k => `ENV ${k}=${containerEnv![k]}`).join('\n');
+	// https://docs.docker.com/engine/reference/builder/#envs
+	const concatenatedEnv = keys.map(k => `ENV ${k}="${containerEnv[k]
+		.replace(/(?=["\\$])/g, '\\') // escape " and \
```
jkeech, line 500: What's the purpose of matching the `$` character? Can you add a test to help explain and cover this scenario? Same thing for the backslash literal `\`.

src/spec-node/imageMetadata.ts
```diff
@@ -494,8 +494,11 @@ export function getContainerEnvMetadata(containerEnv: Record<string, string> | u
 	if (!containerEnv) {
 		return '';
 	}
-
 	const keys = Object.keys(containerEnv);
-	const concatenatedEnv = keys.map(k => `ENV ${k}=${containerEnv![k]}`).join('\n');
+	// https://docs.docker.com/engine/reference/builder/#envs
+	const concatenatedEnv = keys.map(k => `ENV ${k}="${containerEnv[k]
+		.replace(/(?=["])/g, '\\') // escape double quotes
```
chrmarti, line 500: Would backslashes also require escaping?

chrmarti (changes requested):

src/spec-node/imageMetadata.ts
jkeech: Yes, I think we need to escape backslashes as well. Good call.

jkeech: Actually I think the original regex was probably correct: `replace(/(?=["\\$])/g, '\\')`. Looks like we need to escape `$` as well.

chrmarti (approved):
👍

PR #460, opened 2023-03-19
Add support for Docker credential helpers

aaronlehmann:
This adds the ability to retrieve registry credentials from a credential helper specified in `credHelpers` or `credsStore` inside ~/.docker/config.json.

### DIFF
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -1,5 +1,4 @@
 import * as os from 'os';
-import * as fs from 'fs';
 import * as path from 'path';
 import * as jsonc from 'jsonc-parser';
 
@@ -221,7 +220,7 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 			let defaultCredHelper = '';
 			// Try platform-specific default credential helper
 			if (process.platform === 'linux') {
-				if (existsInPath('pass')) {
+				if (await existsInPath('pass')) {
 					defaultCredHelper = 'pass';
 				} else {
 					defaultCredHelper = 'secret';
@@ -246,14 +245,14 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 	return;
 }
 
-function existsInPath(filename: string): boolean {
+async function existsInPath(filename: string): Promise<boolean> {
 	if (!process.env.PATH) {
 		return false;
 	}
 	const paths = process.env.PATH.split(':');
 	for (const path of paths) {
 		const fullPath = `${path}/${filename}`;
-		if (fs.existsSync(fullPath)) {
+		if (await isLocalFile(fullPath)) {
 			return true;
 		}
 	}
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -232,7 +232,7 @@ async function getCredentialFromHelper(params: CommonParams, registry: string, c
 			stdin: Buffer.from(registry, 'utf-8'),
 			output,
 		});
-		helperOutput = stdout
+		helperOutput = stdout;
 	} catch (err) {
 		output.write(`[httpOci] Failed to execute credential helper ${credHelperName}"`, LogLevel.Error);
 		return undefined;
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -2,6 +2,7 @@ import * as os from 'os';
 import * as path from 'path';
 import * as jsonc from 'jsonc-parser';
 
+import { runCommandNoPty, plainExec } from '../spec-common/commonUtils';
 import { request, requestResolveHeaders } from '../spec-utils/httpRequest';
 import { LogLevel } from '../spec-utils/log';
 import { isLocalFile, readLocalFile } from '../spec-utils/pfs';
@@ -16,6 +17,15 @@ interface DockerConfigFile {
 			identitytoken?: string; // Used by Azure Container Registry
 		};
 	};
+	credHelpers: {
+		[registry: string]: string
+	};
+	credsStore: string;
+}
+
+interface CredentialHelperResult {
+	Username: string;
+	Secret: string;
 }
 
 // WWW-Authenticate Regex
@@ -137,8 +147,6 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 	const { output, env } = params;
 	const { registry } = ociRef;
 
-	// TODO: Ask docker credential helper for credentials.
-
 	if (!!env['GITHUB_TOKEN'] && registry === 'ghcr.io') {
 		output.write('[httpOci] Using environment GITHUB_TOKEN for auth', LogLevel.Trace);
 		const userToken = `USERNAME:${env['GITHUB_TOKEN']}`;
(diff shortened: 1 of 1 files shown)


aaronlehmann:
@microsoft-github-policy-service agree company="Netflix"

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -186,6 +194,20 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 							refreshToken: undefined,
 						};
 					}
+					if (dockerConfig.credHelpers && dockerConfig.credHelpers[registry]) {
+						const credHelper = dockerConfig.credHelpers[registry];
+						output.write(`[httpOci] Found credential helper '${credHelper}' in config.json for registry '${registry}'`, LogLevel.Trace);
```
samruddhikhandale, line 199:
```suggestion
						output.write(`[httpOci] Found credential helper '${credHelper}' in '${dockerConfigPath}' for registry '${registry}'`, LogLevel.Trace);
```

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -186,6 +194,20 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
...
+						const auth = await getCredentialFromHelper(params, registry, credHelper);
+						if (auth) {
+							return auth;
+						}
+					} else if (dockerConfig.credsStore) {
```
samruddhikhandale, line 204: Let's say the user has credHelper but the CLI wasn't able to get auth from it, then shouldn't we also check for credsStore?
Hence, should this be an `if` condition as well?

samruddhikhandale (commented):
Nice, thanks for putting up this PR! Left some minor comments

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -16,6 +17,15 @@ interface DockerConfigFile {
 			identitytoken?: string; // Used by Azure Container Registry
 		};
 	};
+	credHelpers: {
+		[registry: string]: string
```
samruddhikhandale, line 21:
```suggestion
		[registry: string]: string;
```

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -198,6 +220,34 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
...
+	}
+
+	const creds:CredentialHelperResult = jsonc.parse(stdout.toString());
+	if (creds.Username === '<token>') {
+		return {
+			refreshToken: creds.Secret,
+			base64EncodedCredential: undefined,
+		};
+	}
+	const userToken = creds.Username+':'+creds.Secret
```
samruddhikhandale, line 244:
```suggestion
	const userToken = `${creds.Username}:${creds.Secret}`;
```

src/spec-configuration/httpOCIRegistry.ts
aaronlehmann: The Docker docs say:

> The default credential store (`credsStore` or the config file itself) will not be used for operations concerning credentials of the specified registries.

https://docs.docker.com/engine/reference/commandline/login/

I think this means that if a registry has a `credHelpers` entry, `credsStore` should not be used for that registry.

aaronlehmann:
Thanks for the review! I made the suggested changes.

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -198,6 +220,34 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 	return;
 }
 
+async function getCredentialFromHelper(params: CommonParams, registry: string, credHelperName: string) : Promise<{ base64EncodedCredential: string | undefined; refreshToken: string | undefined } | undefined>{
+	const { output } = params;
+
+	const { stdout } = await runCommandNoPty({
```
chrmarti, line 226: This would throw an error if the exit code is not zero. Should we handle that by returning undefined here?

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -198,6 +220,34 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
...
+		cmd: 'docker-credential-'+credHelperName,
+		args: ['get'],
+		stdin: Buffer.from(registry, 'utf-8'),
+		output,
+	});
+	if (stdout.length === 0) {
+		return undefined;
+	}
+
+	const creds:CredentialHelperResult = jsonc.parse(stdout.toString());
```
chrmarti, line 237: `jsonc.parse` throws an error if the string cannot be parsed. Should we catch that and return undefined here?

chrmarti (changes requested):
Thanks, left a few comments.

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -186,6 +194,20 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 							refreshToken: undefined,
 						};
 					}
+					if (dockerConfig.credHelpers && dockerConfig.credHelpers[registry]) {
```
chrmarti, line 197: Is there documentation on whether auths or credHelpers should take precedence? I also tried finding Docker's implementation, but somehow couldn't find it.

chrmarti (changes requested):

aaronlehmann:
Thanks! I've made these changes.

chrmarti (changes requested):
Great, thanks! Left another question and ran the linter.

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -168,8 +176,22 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
 				if (await isLocalFile(dockerConfigPath)) {
 					const dockerConfig: DockerConfigFile = jsonc.parse((await readLocalFile(dockerConfigPath)).toString());
 
+					if (dockerConfig.credHelpers && dockerConfig.credHelpers[registry]) {
```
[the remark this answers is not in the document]
aaronlehmann: I've gone ahead and added equivalent behavior here.

joshspicer (commented):
This is awesome, thanks for taking the time (triggered the existing unit tests to run on this PR).

Do you have any suggestions on how to craft unit tests for this?  Most of our unit tests run in linux, and we do have a subset of them running in windows.  Theoretically we could run against macOS too (if we wanted to...)

Currently familiarizing myself with [how this works](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol)

aaronlehmann:
> Do you have any suggestions on how to craft unit tests for this?

Great question. The credential helper "protocol" is very simple and it would be easy to mock a credential helper. But to ensure the returned creds are actually used, we'd need an authenticated registry to connect to, and I'm not sure what kind of test infra exists for that.

joshspicer (approved):
+1 to some of the earlier feedback (which you've addressed).  I don't want to block on the tests - I can make it an action item for myself if you don't have the time. There's one minor linting error in CI.

Validated this works well on MacOS with AzureCR creds (this branch is `./devcontainer.js` and the current release is on my PATH)

image

src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -213,13 +216,50 @@ async function getCredential(params: CommonParams, ociRef: OCIRef | OCICollectio
...
 }
 
+function existsInPath(filename: string): boolean {
+	if (!process.env.PATH) {
+		return false;
+	}
+	const paths = process.env.PATH.split(':');
+	for (const path of paths) {
+		const fullPath = `${path}/${filename}`;
+		if (fs.existsSync(fullPath)) {
```
chrmarti, line 256: Could we use an async call to avoid blocking the JavaScript thread?

joshspicer: you may consider taking a look at https://github.com/devcontainers/cli/blob/main/src/spec-utils/pfs.ts#L21-L31

aaronlehmann: Updated to use `isLocalFile` and make this function async - thanks

joshspicer (approved):

chrmarti (approved):
Thanks for your contribution!

aaronlehmann:
Thanks so much for the help geting this through! It's been a great experience.

Issue #479, opened 2023-04-04
Support fetching Features|Templates by SHA digest


joshspicer:
Just as docker image can be pulled [by digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) instead of tags, Features and Templates should be able to be pinned to a SHA256 digest in the user `devcontainer.json`.  

This change requires no changes to published Features|Templates, only a change to the CLI to respect a provided SHA.

-----

Digests can be found on the GHCR package page for a published Feature: https://github.com/devcontainers/features/pkgs/container/features%2Fgit-lfs/82847789?tag=1

<img width="777" alt="image" src="https://user-images.githubusercontent.com/23246594/229860375-07ded780-7710-42a1-8c0c-f3312b4ae6ac.png">

It's also available as the [JSON response to the `devcontainer feature publish` command](https://github.com/devcontainers/features/actions/runs/4608512953/jobs/8144373610#step:3:412)
```jsonc
{
   // ...
   "git-lfs": {
      "publishedVersions" :["1","1.1","1.1.0","latest"],
      "digest":"9e0b6df853ec932203945b76b9cfa7470ebd1f8bd765484027608581f9cf8a77",
      "version":"1.1.0"
    }
}
```

jkeech:
@joshspicer with this change, what would be the syntax in devcontainer.json? Would it be the following?

```
{
  "features": {
    "ghcr.io/devcontainers/features/git-lfs@sha256:9e0b6df853ec932203945b76b9cfa7470ebd1f8bd765484027608581f9cf8a77": {}
  }
}
```

joshspicer:
Yes, I don't see any reason to not align with the pattern used elsewhere in the OCI/docker specs for container images.

The one caveat is that we currently detect the "v1" Features (that we hosted via GitHub Release) with the `@` symbol, but I believe we can get around that while also maintaining the back compat.

joshspicer:
Version [0.39.0](https://github.com/devcontainers/cli/blob/main/CHANGELOG.md#0390) of the dev container CLI now supports pinning Features by digest.

PR #480, opened 2023-04-05
Fetch Features/Templates by digest

joshspicer:
ref: https://github.com/devcontainers/cli/issues/479

Previously, all Features and Templates resolved a digest and associated manifest by tag.  This change also allow resolving the Feature or Template by sha256 digest directly. 

Eg: 
```json
{
  "features": {
    "ghcr.io/devcontainers/features/git-lfs@sha256:9e0b6df853ec932203945b76b9cfa7470ebd1f8bd765484027608581f9cf8a77": {}
  }
}
```

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -20,6 +20,7 @@ export interface CommonParams {
 
 // Represents the unique OCI identifier for a Feature or Template.
 // eg:  ghcr.io/devcontainers/features/go:1.0.0
+// eg:  ghcr.io/devcontainers/features/go@sha256:fe73f123927bd9ed1abda190d3009c4d51d0e17499154423c5913cf344af15a3
 // Constructed by 'getRef()'
 export interface OCIRef {
 	registry: string; 		// 'ghcr.io'
@@ -28,7 +29,10 @@ export interface OCIRef {
 	path: string;			// 'devcontainers/features/go'
 	resource: string;		// 'ghcr.io/devcontainers/features/go'
 	id: string;				// 'go'
-	version?: string;		// '1.0.0'
+
+	version: string;		// (Either the contents of 'tag' or 'digest')
+	tag?: string;			// '1.0.0'
+	digest?: string; 		// 'sha256:fe73f123927bd9ed1abda190d3009c4d51d0e17499154423c5913cf344af15a3'
 }
 
 // Represents the unique OCI identifier for a Collection's Metadata artifact.
@@ -38,6 +42,7 @@ export interface OCICollectionRef {
 	registry: string;		// 'ghcr.io'
 	path: string;			// 'devcontainers/features'
 	resource: string;		// 'ghcr.io/devcontainers/features'
+	tag: 'latest';			// 'latest' (always)
 	version: 'latest';		// 'latest' (always)
 }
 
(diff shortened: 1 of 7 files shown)


samruddhikhandale (dismissed):
Awesome 🔥

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -124,20 +131,49 @@ export function getRef(output: Log, input: string): OCIRef | undefined {
...
+		resource = input.substring(0, indexOfLastAtCharacter);
+		const digestWithHashingAlgorithm = input.substring(indexOfLastAtCharacter + 1);
+		const splitOnColon = digestWithHashingAlgorithm.split(':');
+		if (splitOnColon.length !== 2) {
+			output.write(`Failed to parse digest '${digestWithHashingAlgorithm}'.   Expected format: 'sha256:abcdefghijk'`, LogLevel.Error);
+			return undefined;
+		}
+
+		if (splitOnColon[0] !== 'sha256') {
+			output.write(`Digest algorithm for input '${input}' failed validation.  Expected hashing algorithm to be 'sha256'.`, LogLevel.Error);
```
chrmarti, line 152: Do we rely on it being sha256?

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -124,20 +131,49 @@ export function getRef(output: Log, input: string): OCIRef | undefined {
...
+		if (!regexForVersionOrDigest.test(splitOnColon[1])) {
+			output.write(`Digest for input '${input}' failed validation.  Expected digest to match regex '${regexForVersionOrDigest}'.`, LogLevel.Error);
+		}
+
+		digest = digestWithHashingAlgorithm;
 	} else {
-		resource = input.substring(0, indexOfLastColon);
-		version = input.substring(indexOfLastColon + 1);
+		// In both cases, assume 'latest' tag.
+		if (indexOfLastColon === -1 || indexOfLastColon < input.indexOf('/')) {
```
chrmarti, line 163: Would `input.lastIndexOf('/')` be more precise?

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -149,36 +185,48 @@ export function getRef(output: Log, input: string): OCIRef | undefined {
...
-	// Validate results of parse.
+	// -- Validate results of parse.
 
 	if (!regexForPath.exec(path)) {
-		output.write(`Parsed path '${path}' for input '${input}' failed validation.`, LogLevel.Error);
+		output.write(`Path '${path}' for input '${input}' failed validation.  Expected path to match regex '${regexForPath}'.`, LogLevel.Error);
+		return undefined;
+	}
+
+	if (digest && tag) {
```
chrmarti, line 210: This shouldn't ever be true given the above code.

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -149,36 +185,48 @@ export function getRef(output: Log, input: string): OCIRef | undefined {
...
+		output.write(`Parsed both a digest ('${digest}') and a tag ('${tag}') for input '${input}'. Resource can only have one.`, LogLevel.Error);
 		return undefined;
 	}
 
-	if (!regexForReference.test(version)) {
-		output.write(`Parsed version '${version}' for input '${input}' failed validation.`, LogLevel.Error);
+	if (tag && !regexForVersionOrDigest.test(tag)) {
```
chrmarti, line 215: Move up to where `tag` is assigned for symmetry with the digest code path?

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -28,7 +29,10 @@ export interface OCIRef {
 	path: string;			// 'devcontainers/features/go'
 	resource: string;		// 'ghcr.io/devcontainers/features/go'
 	id: string;				// 'go'
-	version?: string;		// '1.0.0'
+
+	version: string;		// (Either the contents of 'tag' or 'digest')
```
chrmarti, line 33: Would it clarify the code if you omit the `version` property? I guess tags and digests cannot always be used the same way and having (only) separate properties makes it less likely to overlook the other case when writing and reviewing code?

chrmarti (changes requested):
A few nits, otherwise looks great!

src/spec-configuration/containerCollectionsOCI.ts
joshspicer: Yes, there is one section where we compute a sha256 of a manifest to generate the digest.  That's currently hardcoded.  I don't believe changing the hashing algorithm changing is anything we need to worry about for a while - I think it's likely appended here in case sha256 is found to be broken in the future and a new hashing algortithm needs to be swapped out for it.

src/spec-configuration/containerCollectionsOCI.ts
joshspicer: Actually the registry api handles both tag and digest the same way, so it's cleaner in a lot of the codebase to have a single 'version' that is interpolated into the request paths.  I added both `tag` and `digest` in case there are instances in the codespace where it's interesting to know which is which.

chrmarti (approved):

joshspicer:
+ using this new functionality in an E2E test.

samruddhikhandale (approved):

Issue #473, opened 2023-03-31
Pulling features from ghcr.io breaks when GHE token is set


gpiper14:
There is an issue where pulling a feature from `ghcr.io` will fail with a 403 if you have a `GITHUB_TOKEN` env var set for a GitHub Enterprise instance.  The issue appears to stem from [here](https://github.com/devcontainers/cli/blob/d02379c581b55233946f407a09a8ee1eeeff5fb4/src/spec-configuration/httpOCIRegistry.ts#L131) as it assumes any `GITHUB_TOKEN` env var is a `github.com` token.

## Possible solution

A possible solution could be to check for the existence of a `GITHUB_HOST` env var, and if it is set, only use the `GITHUB_TOKEN` for `ghcr.io` requests if it is set to `github.com`

joshspicer:
@gpiper14, in your opinion (from the GitHub Enterprise scenario) is it ok behavior to assume that if no `GITHUB_HOST` is set that the `GITHUB_TOKEN` is for the public github.com instance?  

I'm thinking to bump GITHUB_TOKEN to have the least precedence (looking at docker config/cred helpers first), and for back compat continuing to fall back to use `GITHUB_TOKEN` unless `GITHUB_HOST` is set and not `github.com` (see: https://github.com/devcontainers/cli/pull/482/files#diff-00403d33ea4c9f9520b612ddfc2028d85401eb237f1d42edf13a6b258c421fadR176)

Is `GITHUB_HOST` documented somewhere that you could share with me, or is this env var convention of some sort?

gpiper14:
That seems like a decent assumption to me, but would likely want to make the behavior clear in some sort of documentation.

Im not sure if there is any "official" guidance from github for `GITHUB_HOST`.  The closest thing I have is the `hub` cli produced by github, which is why i have that variable set in the first place.  See docs [here](https://hub.github.com/hub.1.html)

Thanks for the quick reply on this!

joshspicer:
Thanks for putting this on our radar!

brcarp:
[...]

Issue #468, opened 2023-03-27
[oci implementation] Test all authentication mechanisms


joshspicer:
https://github.com/devcontainers/cli/pull/460 is adding in support for [docker credential helpers](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol).  There's existing test coverage on the "simpler" auth mechanisms but not on all.

This likely means utlizing runners for macOS, linux, and windows mocking out the credential helper api.  This would also mean needing to embed credentials in the test runner to fetch private Features based on the credentials. (I personally use `privatedevcontainercli.azurecr.io/features/a`, but those secrets aren't available to CI yet).

cc/ @aaronlehmann

PR #482, opened 2023-04-05
Add regression testing and reorganize auth precedence levels `httpOCIRegistry`

joshspicer:
related: https://github.com/devcontainers/cli/issues/473, #468 

Changes here are primarily around re-structuring the different auth strategies to rely more on the standard docker auth strategies (docker config file and docker credential helper), and less on the registry-specific strategies that previously held higher precedence.

The precedence order is now (encoded in [`getCredential(...)`](https://github.com/devcontainers/cli/blob/4fde394ac16df1061b731d2d2f226850277cbce2/src/spec-configuration/httpOCIRegistry.ts#L147):

   - parsed out of a special `DEVCONTAINERS_OCI_AUTH` environment variable
   - Read from a docker credential helper indicated in config
   - Read from a docker cred store indicated in config (https://docs.docker.com/engine/reference/commandline/login/#credentials-store)
   - Read from a docker config file (flat file with base64 encoded credentials)
   - Read from the platform's default credential helper
   - Crafted from the GITHUB_TOKEN environment variable

Two new minor changes have also been added:

-  If `GITHUB_HOST` is set to something other than `github.com`, `GITHUB_TOKEN` is not used for ghcr.io (useful for GitHub Enterprise instances, among other cases). Reported in #473 
- Aligning with the docker cli, `DOCKER_CONFIG` can be set to change the path that the docker config file is looked for (https://docs.docker.com/engine/reference/commandline/cli/#change-the-docker-directory).  (This is immediately useful in the tests outlined below).

---

This change also introduces significant changes to the `registryCompatibilityOCI.test.ts` test, introducing a way to quickly add "testPlans" (new registries) with different forms of provided auth.  This PR introduces four "testPlans", testing both AzureCR and GHCR both anonymous and with some form of auth.

Since we want folks to be able to fork this repo and run all the tests without failure, the tests will automatically skip if there is a set `useAuthStrategy` and no accompanying `authStrategyKey` (environment variable, containing a credential if needed for the test and parsed account according to the `useAuthStrategy` enum).  This lets us add in the necessary secrets to our GitHub Action CI tests without introducing flakey tests to other contributors who may have forks of this repo.  Mocha has a way to mark tests as "skipped", which is what we use here.

```ts
const registryCompatibilityTestPlan: TestPlan[] = [
	{
		name: 'Anonymous access of Azure Container Registry',
		configName: 'azure-anonymous',
		testFeatureId: 'devcontainercli.azurecr.io/features/color',
	},
	{
		name: 'Anonymous access of GHCR',
		configName: 'github-anonymous',
		testFeatureId: 'ghcr.io/devcontainers/feature-starter/color',
	},
	// https://learn.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions
	{
		name: 'Authenticated access of Azure Container Registry with registry scoped token',
		configName: 'azure-registry-scoped',
		testFeatureId: 'privatedevcontainercli.azurecr.io/features/rabbit',
		useAuthStrategy: AuthStrategy.DockerConfigAuthFile,
		authStrategyKey: 'FEATURES_TEST__AZURE_REGISTRY_SCOPED_CREDENTIAL',
		testCommand: 'rabbit',
		testCommandResult: /rabbit-is-the-best-animal/,
	},
	// Via GHCR visibility settings, this repo's GitHub Actions CI should be able to access this Feature via its GITHUB_TOKEN.
	{
		name: 'Private access of GHCR via an environment GITHUB_TOKEN',
		configName: 'github-private',
		testFeatureId: 'ghcr.io/devcontainers/private-feature-set-for-tests/color',
		useAuthStrategy: AuthStrategy.GitHubToken,
		authStrategyKey: 'RUNNING_IN_DEVCONTAINERS_CLI_REPO_CI'
	}
];
```

In CI on this repository the above `authStrategyKey` variables are set, so no tests should be skipped (as seen [here](https://github.com/devcontainers/cli/actions/runs/4660764835/jobs/8249229694?pr=482#step:7:1786))

### DIFF
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -193,16 +193,12 @@ async function existsInPath(filename: string): Promise<boolean> {
 	if (!process.env.PATH) {
 		return false;
 	}
-	try {
-		const paths = process.env.PATH.split(':');
-		for (const path of paths) {
-			const fullPath = `${path}/${filename}`;
-			if (await isLocalFile(fullPath)) {
-				return true;
-			}
+	const paths = process.env.PATH.split(':');
+	for (const path of paths) {
+		const fullPath = `${path}/${filename}`;
+		if (await isLocalFile(fullPath)) {
+			return true;
 		}
-	} catch (err) {
-		return false;
 	}
 	return false;
 }
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -136,8 +136,8 @@ export async function requestEnsureAuthenticated(params: CommonParams, httpOptio
 // This credential is used to offer the registry in exchange for a Bearer token.
 // These may be:
 //   - parsed out of a special DEVCONTAINERS_OCI_AUTH environment variable
-//   - Read from a docker config file
 //   - Read from a docker credential helper (https://docs.docker.com/engine/reference/commandline/login/#credentials-store)
+//   - Read from a docker config file
 //   - Crafted from the GITHUB_TOKEN environment variable
 //  Returns:
 //   - undefined: No credential was found.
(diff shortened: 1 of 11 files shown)


src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -249,16 +193,105 @@ async function existsInPath(filename: string): Promise<boolean> {
...
-			return true;
+	try {
+		const paths = process.env.PATH.split(':');
+		for (const path of paths) {
+			const fullPath = `${path}/${filename}`;
+			if (await isLocalFile(fullPath)) {
+				return true;
+			}
 		}
+	} catch (err) {
```
chrmarti, line 204: When is an error thrown? (Wondering if we might hide other issues with the catch-all.)

chrmarti (dismissed):

src/spec-configuration/httpOCIRegistry.ts
joshspicer: Looking again, I don't have a great reason in mind to keep this. Removing.

samruddhikhandale (approved):

PR #490, opened 2023-04-13
Surface `docker-content-digest` on manifest pulls, manifest annotations per Feature/Template, and `info` command clean up

joshspicer:
This patch introduces two small additions to the registry/OCI implementation.  Both additions are  generally useful to have, and have been extracted from my [experimental implementation of  'Feature Dependencies' with a `dependsOn` attribute](https://github.com/devcontainers/cli/tree/joshspicer/dependsOn-experiment).  This patch, however, does not depend on any dependency-related spec and can be included regardless.

Ref: https://github.com/github/codespaces/issues/11243, https://github.com/devcontainers/spec/issues/109

Changes:

- When fetching a manifest, also read the `docker-content-digest` header and pass that along to the caller.  This will be useful:
   - In combination with https://github.com/devcontainers/cli/pull/480
   - As a possible way to assist in deduping Feature dependencies
   - Provides the ability to use the hash as an integrity check
- The open containers image spec includes a [map of arbitrary annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md). Supporting the ability to attach annotations to an individual Feature/Template at publish time is a generally useful mechanism to have. 

   -  One example where this could be used is attaching pointers to Features that a given Feature depends on.  This allows the CLI to not need to download and extract an entire Feature archive to get to this metadata.

Possible Example (`dev.containers.experimental.dependsOn`) :
```
{
    "contentDigest": "sha256:b4153e1557157ebb610ca86353fa62602a759110a6a3bc6730599c86d75fd0cc",
    "manifestObj": {
        "schemaVersion": 2,
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "config": {
            "mediaType": "application/vnd.devcontainers",
            "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
            "size": 0
        },
        "layers": [
            {
                "mediaType": "application/vnd.devcontainers.layer.v1+tar",
                "digest": "sha256:204ab620e4f4ad0e9db79de73d8941b0de849ec3cc5dbacae693e87eeaca8d88",
                "size": 3584,
                "annotations": {
                    "org.opencontainers.image.title": "devcontainer-feature-C.tgz"
                }
            }
        ],
        "annotations": {
            "dev.containers.experimental.dependsOn": "{\"ghcr.io/joshspicer/dependsOnExperiment/A\":{\"magicNumber\":\"40\"},\"ghcr.io/joshspicer/dependsOnExperiment/E\":{\"magicNumber\":\"50\"}}",
            "com.github.package.type": "devcontainer_feature"
        }
    }
```
- This patch also refactors the `info` command to consolidate code and allows for a 'verbose' mode that dumps all the info this command can query.

image

image


src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -227,28 +236,59 @@ export async function fetchOCIManifestIfExists(params: CommonParams, ref: OCIRef
...
+	const { body, headers } = res;
+	const manifestStringified = JSON.stringify(body);
+
+	// Per the specification:
+	// https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#pulling-manifests
+	// The registry server SHOULD return the canonical content digest in a header, but it's not required to.
+	// That is useful to have, so if the server doesn't provide it, recalculate it outselves.
+	let contentDigest = headers['docker-content-digest'] || headers['Docker-Content-Digest'];
+	if (!contentDigest) {
+		contentDigest = `sha256:${crypto.createHash('sha256').update(manifestStringified).digest('hex')}`;
```
chrmarti, line 270: Do we rely on this being the same digest as computed on the response body? If so, we should compute it from the original response body (the Buffer object, as opposed to the parsed and then stringified one).

src/spec-configuration/containerCollectionsOCI.ts
```diff
@@ -227,28 +236,59 @@ export async function fetchOCIManifestIfExists(params: CommonParams, ref: OCIRef
...
+	}
+
+	const { body, headers } = res;
+	const manifestStringified = JSON.stringify(body);
+
+	// Per the specification:
+	// https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#pulling-manifests
+	// The registry server SHOULD return the canonical content digest in a header, but it's not required to.
+	// That is useful to have, so if the server doesn't provide it, recalculate it outselves.
+	let contentDigest = headers['docker-content-digest'] || headers['Docker-Content-Digest'];
```
chrmarti, line 268: Node lowercases the incoming headers.

src/spec-configuration/containerTemplatesOCI.ts
```diff
@@ -36,8 +36,13 @@ export async function fetchTemplate(params: CommonParams, selectedTemplate: Sele
 		output.write(`Failed to fetch template manifest for ${userSelectedId}`, LogLevel.Error);
 		return;
 	}
+	const blobDigest = ociManifest?.manifestObj?.layers[0].digest;
```
chrmarti, line 39: Can layers be empty?

chrmarti (changes requested):
Looks good, a few comments.

src/spec-configuration/containerCollectionsOCI.ts
joshspicer: Good to know!

src/spec-configuration/containerTemplatesOCI.ts
joshspicer: For the custom dev container media types we have out in the wild, there should always be exactly one layer (the tgz or single json collection file).  That said it is good here to validate that.

src/spec-configuration/containerCollectionsOCI.ts
joshspicer: Yes this definitely should be the same digest that everyone else computes (and that we originally computed)

Comparing what this returns to what I see on GHCR, the results hashes are equal.  

image

This is also how I generate the manifest at publish time (https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerCollectionsOCIPush.ts#L408), so if any changes are made, it should be made in both spots.

chrmarti: It will be more robust to compute the hash on a Buffer with the originally received bits. `manifestStringified` is derived from that like `JSON.stringify(JSON.parse(buffer.toString()))` and the hash function will itself first convert the string back to a Buffer. These conversions are not guaranteed to get you back the original bits.

E.g.: `JSON.stringify(JSON.parse('{ "p": "\\u0022" }')) === '{"p":"\\""}'` (whitespace and escaping changes)

chrmarti (dismissed):
Looks good, thanks!

chrmarti (dismissed):

samruddhikhandale (approved):

PR #503, opened 2023-04-21
Consistent usage of digest

joshspicer:
This change updates the output of the `devcontainer features publish` command to include the hashing algorithm (`sha256:`) at the beginning of the digest in the command output.  This simply makes it more consistent with how a user would pin to digest.

### DIFF
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -302,7 +302,7 @@ async function getCredentialFromHelper(params: CommonParams, registry: string, c
 		});
 		helperOutput = stdout;
 	} catch (err) {
-		output.write(`[httpOci] Failed to execute credential helper ${credHelperName}`, LogLevel.Error);
+		output.write(`[httpOci] Failed to query for '${registry}' credential from 'docker-credential-${credHelperName}': ${err}`, LogLevel.Trace);
 		return undefined;
 	}
 	if (helperOutput.length === 0) {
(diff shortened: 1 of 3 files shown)


src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -302,7 +302,7 @@ async function getCredentialFromHelper(params: CommonParams, registry: string, c
 		});
 		helperOutput = stdout;
 	} catch (err) {
-		output.write(`[httpOci] Failed to execute credential helper ${credHelperName}`, LogLevel.Error);
+		output.write(`[httpOci] Failed to query for '${registry}' credential from 'docker-credential-${credHelperName}': ${err}`, LogLevel.Trace);
```
samruddhikhandale, line 305: Question: Why change the log level to Trace? If that's not an error, then maybe change to `LogLevel.Warning` instead?

samruddhikhandale (approved):
Looks great! ⚡

src/spec-configuration/httpOCIRegistry.ts
joshspicer: The way this code is structured, this `catch` will always fire unless the user has a credential helper with this registry's secret. I thought it would be confusing for users to see this fire when there's no expectation these credential helpers are available.

CRH35: I am going to KILL YOU. Like Osama Bin Laden. - A1  CIA

chrmarti (approved):

PR #509, opened 2023-04-28
Install dependency modules via common js in the dev container.

naturedamends:
Ensure chalk installation in devcontainer to an earlier version, since it now uses es modules.


joshspicer:
Thanks for taking the time to contribute.  Could you please update the description box with an explanation of the change?

naturedamends:
> Thanks for taking the time to contribute. Could you please update the description box with an explanation of the change?

I've just installed this with the `yarn install` and it worked, looks like it imports chalk latest version.

I must have corrupted my package lock. closed.

PR #511, opened 2023-05-01
Experiment: `dependsOn` Feature dependencies

joshspicer:
ref: https://github.com/devcontainers/spec/issues/109
implements proposal: https://github.com/devcontainers/spec/pull/234

This change modifies Feature installation order.  Namely, the effect that the `dependsOn`, `installsAfter`, and `overrideFeatureInstallOrder` properties have on installation order.  The default installation order (with no additional properties) may also change with this change.  

The large majority of the logical changes can be found in the [`containerFeaturesOrder.ts`](https://github.com/devcontainers/cli/blob/joshspicer/dependsOn/src/spec-configuration/containerFeaturesOCI.ts) file, which has been completely rewritten.

The algorithm in this file in invoked as shown below.

https://github.com/devcontainers/cli/blob/592dba5ace8664dcb014cde1d5a9f1cc6d191ff1/src/spec-configuration/containerFeaturesConfiguration.ts#L551-L554

`installationOrder` contains featureSets in the correct order of installation - those are passed forward to the remainder of the Features installation code. Everything else following works exactly as it had before.

https://github.com/devcontainers/cli/blob/joshspicer/dependsOn/src/spec-configuration/containerFeaturesConfiguration.ts#L556-L559

Some notable effects of this change are that:
- Features with a `dependsOn` property will _recursively_ fetch and add its dependencies to the list of Features to install.  If the algorithm encounters a cycle or invalid state, it will fail.
- Due to the definition of `dependsOn`, the same Feature may be installed more than once (provided it has either a different version _or_ different options).  Existing Features may not be expecting this behavior and may need to be updated before being declared as a dependency.


joshspicer:
Closing in favor of https://github.com/devcontainers/cli/pull/530, which rebases changes into more easily digestible commits

PR #512, opened 2023-05-01
Fix multiple tarball repositories

### LINKED ISSUE #506: computeInstallationOrder does not work when specifying features via repository url
When using absolute URLs for features, computeInstallationOrder reduces the feature set to the last entry. This is caused due to https://github.com/devcontainers/cli/commit/dc23b12bafc8fa0eac43fb1b1daccef8ba17ce82 which is now mapping the features to be a key by `feature.feature.sourceInformation.userFeatureId.split(':')[0]` which for the case where you specify a URL would be `https` for example. 

What results is the calculated install order will only contain a single entry if multiple features are using a url link to the tarball

brian-mcnamara:
Fixes #506 

Sorry for the delay on this one. I have tested this internally and can confirm this solves the issue. The expectation here is that features pulled in via tarball are versioned a different way and within a container there should only be one tarball feature sharing a feature id

### DIFF
--- a/src/spec-configuration/containerFeaturesOrder.ts
+++ b/src/spec-configuration/containerFeaturesOrder.ts
@@ -49,7 +49,9 @@ export function computeInstallationOrder(features: FeatureSet[]) {
         feature,
         before: new Set(),
         after: new Set(),
-    })).reduce((map, feature) => map.set(feature.feature.sourceInformation.userFeatureId.split(':')[0], feature), new Map<string, FeatureNode>());
+    })).reduce((map, feature) => {
+        return map.set(getUniqueFeatureId(feature.feature), feature);
+    }, new Map<string, FeatureNode>());
 
     let nodes = [...nodesById.values()];
 
@@ -124,7 +126,7 @@ export function computeInstallationOrder(features: FeatureSet[]) {
 
     const missing = new Set(nodesById.keys());
     for (const feature of orderedFeatures) {
-        missing.delete(feature.sourceInformation.userFeatureId.split(':')[0]);
+        missing.delete(getUniqueFeatureId(feature));
     }
 
     if (missing.size !== 0) {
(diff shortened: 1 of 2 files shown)


joshspicer (commented):
Thank you! That is very likely an oversight, so thanks for taking the time to track it down.  

I enabled CI and there's at least one thing that seems like it needs to be tweaked - would you be able to take a look at that?

brian-mcnamara:
> 

brian-mcnamara:
> Thank you! That is very likely an oversight, so thanks for taking the time to track it down.
> 
> I enabled CI and there's at least one thing that seems like it needs to be tweaked - would you be able to take a look at that?

Thanks, looking into it and should have a fix. Work calling out this will alter the behavior of existing devcontainers which use tarball and reference them by their url in other operations (for the test case, by overrrideFeatureInstallOrder

joshspicer:
Thanks for the ping.  Looking again, I _think_ that we should likely keep the entire URL in as the key for the override property.  That seems like the intention of that test (which it sounds like was not asserting correctly) 🤔

brian-mcnamara:
> Thanks for the ping. Looking again, I _think_ that we should likely keep the entire URL in as the key for the override property. That seems like the intention of that test (which it sounds like was not asserting correctly) 🤔

The other option would be to add branching logic to https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesOrder.ts#L52 such that if its a tarball we do not split by ':'.

I originally considered that, but thought it would make things too complex. The issue would be if a feature is published and a devcontainer specifies the public feature and also a local tarball copy of the same feature (which I know is a corner case). But if we keep the tarball url as the key, this corner case can happen.

Anyway let me know if you have other ideas :)

joshspicer:
> The issue would be if a feature is published and a dev container specifies the public feature and also a local tarball copy of the same feature (which I know is a corner case). But if we keep the tarball url as the key, this corner case can happen.

This can fundamentally happen with Features published to an OCI registry too, right?  I could use the `publish` command to push the exact same packaged Feature to `ghcr.io/foo/a` and `ghcr.io/bar/a`.  The override property would consider these Features as different even though their contents are the same.  I think that it's good to keep the behavior the same in this case, and have the full URL be the key.

Did I understand the concern here correctly?

brian-mcnamara:
@joshspicer sorry for the silence on this, went out on vacation and just getting back to reimplement another solution.

You are correct on the last comment, I have gone back and reimplemented this to use the url as the key and instead allowed the reduce function to use the url for direct-tarball features. Let me know your thoughts when you have some time to look this over

brian-mcnamara:
@joshspicer checking back on this. Is the new approach I have acceptable?

PR #514, opened 2023-05-09
Add object notation support for `initializeCommand`

joshspicer:
refs: https://github.com/microsoft/vscode-remote-release/issues/7765, https://github.com/devcontainers/spec/issues/168, https://github.slack.com/archives/C0431R35H37/p1683554998695899

Aligns the `initializeCommand` lifecycle hook with the other hooks, allow for the 'object notation', running each provided (and named) command in parallel with one another.

For logging, the behavior is mirrored for the other lifecycle hooks, _not_ continuously streaming the output from each parallel hook, but instead holding the output until the end (otherwise, output will be interleaved and hard to follow).  See the demo below for an illustration of this.

https://user-images.githubusercontent.com/23246594/236972162-860377bb-0337-487b-9dd9-67440ebacefb.mov

### DIFF
--- a/src/spec-common/commonUtils.ts
+++ b/src/spec-common/commonUtils.ts
@@ -181,8 +181,10 @@ export async function runCommand(options: {
 	resolveOn?: RegExp;
 	onDidInput?: Event<string>;
 	stdin?: string;
+	print?: 'off' | 'continuous' | 'end';
 }) {
 	const { ptyExec, cmd, args, cwd, env, output, resolveOn, onDidInput, stdin } = options;
+	const print = options.print || 'continuous';
 
 	const p = await ptyExec({
 		cmd,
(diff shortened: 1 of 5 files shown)


chrmarti (approved):
LGTM!

justin-mccann:
Will this PR also fixup VSCode's problem-highlighting on the acceptable types for `initializeCommand`, or are there separate changes necessary for that?

joshspicer:
> Will this PR also fixup VSCode's problem-highlighting on the acceptable types for `initializeCommand`, or are there separate changes necessary for that?

Separate changes are needed for that. Specifically, an update to the schema published here: https://raw.githubusercontent.com/devcontainers/spec/main/schemas/devContainer.base.schema.json

joshspicer:
Manually confirming things look good with the existing syntaxes

image
image

PR #515, opened 2023-05-09
HttpOci: Retry fetching bearer token anonymously if credentials appear expired

joshspicer:
ref: https://github.com/microsoft/vscode-remote-release/issues/8449

When trying to resolve a bearer token, we first attempt with the user credentials available to us for that service.  If the returned response is 401 or 403 (ghcr appears to return 403), then we attempt one more time after clearing the `Authorization` header.

In this example, I set a GitHub PAT via `docker login` and then revoked the PAT.
image

### DIFF
--- a/src/spec-configuration/httpOCIRegistry.ts
+++ b/src/spec-configuration/httpOCIRegistry.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import * as jsonc from 'jsonc-parser';
 
 import { runCommandNoPty, plainExec } from '../spec-common/commonUtils';
-import { request, requestResolveHeaders } from '../spec-utils/httpRequest';
+import { requestResolveHeaders } from '../spec-utils/httpRequest';
 import { LogLevel } from '../spec-utils/log';
 import { isLocalFile, readLocalFile } from '../spec-utils/pfs';
 import { CommonParams, OCICollectionRef, OCIRef } from './containerCollectionsOCI';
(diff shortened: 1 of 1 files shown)


src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -396,30 +396,31 @@ async function fetchRegistryBearerToken(params: CommonParams, ociRef: OCIRef | O
...
-	let authReq: Buffer;
-	try {
-		authReq = await request(httpOptions, output);
-	} catch (e: any) {
-		// This is ok if the registry is trying to speak Basic Auth with us.
-		output.write(`[httpOci] Could not fetch bearer token for '${service}': ${e}`, LogLevel.Error);
-		return;
+	let res = await requestResolveHeaders(httpOptions, output);
+	if (res.statusCode === 401 || res.statusCode === 403) {
+		output.write(`[httpOci] Credentials for '${service}' may be expired: ${res.resBody.toString()?.trimEnd()}`, LogLevel.Warning);
```
chrmarti, line 401: Only log res.statusCode and res.statusMessage at LogLevel.Info to avoid too much verbosity since this is a regular case? (The user reported everything was working normally otherwise.)

chrmarti (changes requested):
Thanks for the quick response! Left a single comment.

src/spec-configuration/httpOCIRegistry.ts
joshspicer: I reduced one LogLevel down, does that look good to you?

joshspicer: Thoughts? 
image

chrmarti: Sounds good. What about including `res.statusCode` and `res.statusMessage` in the log message for completeness? (The body might be empty, I guess.)

chrmarti (approved):

PR #525, opened 2023-05-17
Allow using `${remoteUser}` inside of features

max06:
This PR aims to (partially) solve devcontainers/spec/issues/220

The primary goal is to allow features access to the `remoteUser`-option set in the parent `devcontainer.json`. 

The example container/feature included in this PR (located at `src/test/container-features/configs/image-with-local-feature`) demonstrates this by using `${remoteUser}` in the feature configuration. 

I am far away from being fluent in typescript or this codebase. There's very likely a better way to implement this change - in that case I'd be very happy about a hint how to improve this PR.

A PR containing the changes for the feature specs, the schema and the documentation pages will be opened by me right after this PR got merged and verified. Also I'll be very happy to publish my first feature making use of that change as well.

### DIFF
--- a/src/test/container-features/lifecycleHooks.test.ts
+++ b/src/test/container-features/lifecycleHooks.test.ts
@@ -390,24 +390,23 @@ describe('Feature lifecycle hooks', function () {
 			});
 
 			it('executes lifecycle hooks with variable substitution', async () => {
+				// substitution in feature
 				const res1 = await shellExec(`${cli} exec --workspace-folder ${testFolder} cat /tmp/feature.variable-substitution.testMarker`);
 				assert.strictEqual(res1.error, null);
 
 				const outputOfExecCommand1 = res1.stdout;
 				console.log(outputOfExecCommand1);
 
-				// Executes the command that was installed by the local Feature's 'postCreateCommand'.
 				assert.strictEqual(outputOfExecCommand1, 'vscode\n');
 				assert.match(containerUpStandardError, /Running the postCreateCommand from Feature '.\/test-feature/);
 
-				// substitutuin in main devcontainer.json
+				// substitution in main devcontainer.json
 				const res2 = await shellExec(`${cli} exec --workspace-folder ${testFolder} cat /tmp/container.variable-substitution.testMarker`);
 				assert.strictEqual(res2.error, null);
 
 				const outputOfExecCommand2 = res2.stdout;
 				console.log(outputOfExecCommand2);
 
-				// Executes the command that was installed by the local Feature's 'postCreateCommand'.
 				assert.strictEqual(outputOfExecCommand2, 'vscode\n');
 				assert.match(containerUpStandardError, /Running the postCreateCommand from devcontainer.json/);
 
@@ -415,7 +414,6 @@ describe('Feature lifecycle hooks', function () {
 				const res3 = await shellExec(`docker inspect ${containerId} --format '{{json .Mounts}}'`);
 				assert.strictEqual(res3.error, null);
 
-				// json parse res3
 				const mounts = JSON.parse(res3.stdout);
 				assert.exists(mounts.find((item: { Type: string; Destination: string }) => item.Type === 'volume' && item.Destination === '/home/vscode'));
 			});


max06:

max06:
In case there's no notification for marking a PR as ready: It's ready 🎉

src/test/configs/image-with-local-feature/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,8 @@
+{
```
joshspicer, line 1: Thanks for adding the test. It doesn't look like it's being executed anywhere?

max06: Test added for `image-with-local-feature`. I'm afraid the resulting image doesn't contain any useful details about the mount from the local feature (or if the var replacement happened). I might have to test this using cli up instead if this test is not sufficient.

joshspicer: That would be great - we have plenty of [examples](https://github.com/devcontainers/cli/blob/main/src/test/container-features/e2e.test.ts) of using `up` with Features.  

Perhaps you could use the new `${remoteUser}` variable in a Feature lifecycle hook to write that in a file (and assert it as you expect)?

joshspicer: If you decide to go that route, the tests in [this file](https://github.com/devcontainers/cli/blob/main/src/test/container-features/lifecycleHooks.test.ts) may be more helpful.

max06: Gotcha! I thought about using it in a mount directive, starting the container and checking it with `docker inspect` if the mount is used properly. But your idea is better. I'll adapt it tomorrow.

max06: Test adapted. Feature now writes `${remoteUser}` into `/tmp/variable-substitution.testMarker` and asserts the content.

max06:
Anyone alive? 😄 @chrmarti maybe?

max06:
Folks... I'd really appreciate some activity in here.

Can I do something to speed this up a bit?

src/spec-common/variableSubstitution.ts
```diff
@@ -109,6 +110,9 @@ function replaceWithContext(isWindows: boolean, context: SubstitutionContext, ma
 		case 'containerWorkspaceFolderBasename':
 			return context.containerWorkspaceFolder !== undefined ? path.posix.basename(context.containerWorkspaceFolder) : match;
 
+		case 'remoteUser':
```
joshspicer, line 113: I believe the behavior will now be that `devcontainer.json` configs will also replace `${remoteUser}`.  I think it would be good to add a test/confirm that behavior so we can document accordingly.

joshspicer (commented):
Apologies for the delay in reviewing.  Thank for adding the test, that's what I had in mind!  I've asked @chrmarti to take a quick peek as he's very familiar with the variable substitution code.  Thanks for following up again

src/test/container-features/configs/image-with-local-feature/.devcontainer/test-feature/devcontainer-feature.json
```diff
@@ -0,0 +1,15 @@
+{
+	"$schema": "https://raw.githubusercontent.com/devcontainers/spec/main/schemas/devContainerFeature.schema.json",
+	"id": "test-feature",
+	"name": "test-feature",
+	"version": "0.0.1",
+	"description": "Testing a feature",
+	"mounts": [
+		{
+			"source": "test-${devcontainerId}",
+			"target": "/home/${remoteUser}",
```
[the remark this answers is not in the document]
max06: I'm not really sure what you mean - the target is inside the container, so it depends on the image used. I could do a `stat` on that directory...

src/test/container-features/lifecycleHooks.test.ts
```diff
@@ -371,6 +371,38 @@ describe('Feature lifecycle hooks', function () {
...
+
+			it('executes lifecycle hooks with variable substitution', async () => {
+				const res = await shellExec(`${cli} exec --workspace-folder ${testFolder} cat /tmp/variable-substitution.testMarker`);
+				assert.strictEqual(res.error, null);
+
+				const outputOfExecCommand = res.stdout;
+				console.log(outputOfExecCommand);
+
+				// Executes the command that was installed by the local Feature's 'postCreateCommand'.
+				assert.match(outputOfExecCommand, /vscode/);
```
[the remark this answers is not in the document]
max06: Ah, right! I expected something around that indicating which step caused the message. I'll correct that.

src/spec-common/variableSubstitution.ts
max06: Will do!

src/test/container-features/configs/image-with-local-feature/.devcontainer/test-feature/devcontainer-feature.json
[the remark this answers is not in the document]
joshspicer: I guess i'm curious here to assert that this source mounted properly to the target after the variable replacement.  That said, your other assertion is probably enough that we're confident the replacement is working, so might not be worth the extra work.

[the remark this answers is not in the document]
max06: Gotcha. I'm currently not sure if you can see mounts inside of a container. In this case `/home/vscode` exists in any case, either as the original directory from the image or as the replacement bind mount. I'd have to test this from the outside using `docker inspect`.

max06:
We have 3 tests now:

- Tests if substitution works inside of a features json
- Tests if substitution works inside of the main devcontainer.json
- Tests if a specified mount containing a substitution gets resolved and used properly.

Happy to add more! Thanks for your time!

joshspicer (approved):

joshspicer:
Merging this in we'll also want to update the documentation on variables on the spec (https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainerjson-reference.md#variables-in-devcontainerjson and https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-features.md)  and website.

src/spec-node/configContainer.ts
```diff
@@ -97,6 +97,7 @@ export async function readDevContainerConfigFile(cliHost: CLIHost, workspace: Wo
 		containerWorkspaceFolder: workspaceConfig.workspaceFolder,
 		configFile,
 		env: cliHost.env,
+		remoteUser: updated.remoteUser
```
chrmarti, line 100: `updated.remoteUser` could use other variables like `${localEnv:DEV_CONTAINER_USER}`, would that work correctly? E.g.:
```json
{
	"name": "Demo container",
 	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
 	"features": {
 		"./test-feature": {}
 	},
 	"remoteUser": "${localEnv:DEV_CONTAINER_USER}",
 	"postCreateCommand": "echo ${remoteUser} > /tmp/container.variable-substitution.testMarker"
 } 
```

Maybe there needs to be a second pass for substituting `${remoteUser}` which can get its value from the devcontainer.json?

Another case to consider is when the `remoteUser` is set in the base image's metadata (e.g., `mcr.microsoft.com/devcontainers/base:bullseye`), in that case `updated.remoteUser` would be `undefined`.

What if the `remoteUser` is not set, but the dev container will run with the container user? Should `${remoteUser}` be replaced with the container user? We also do that for the `_REMOTE_USER` variable.

chrmarti (changes requested):
Thanks for the PR! I have left a few questions.

src/spec-node/configContainer.ts
joshspicer: > Another case to consider is when the remoteUser is set in the base image's metadata (e.g., mcr.microsoft.com/devcontainers/base:bullseye), in that case updated.remoteUser would be undefined.

There's a lot of tests here that could help assert the behavior when `remoteUser` is set on as image metadata: https://github.com/devcontainers/cli/blob/main/src/test/imageMetadata.test.ts

(If you're curious what metadata is attached to an image, this is how I generally take a peek at it on the fly:)
```bash
$ docker pull mcr.microsoft.com/devcontainers/base:bullseye
bullseye: Pulling from devcontainers/base
Digest: sha256:cd9bbfbe4d380278d08fe22c39c5801667424a059c59368819a89ad22ff6936f
Status: Image is up to date for mcr.microsoft.com/devcontainers/base:bullseye
mcr.microsoft.com/devcontainers/base:bullseye

$ export IMAGE="mcr.microsoft.com/devcontainers/base:bullseye"

$ docker image inspect "$IMAGE" | jq '.[0].Config.Labels."devcontainer.metadata" | fromjson'
[
  {
    "id": "ghcr.io/devcontainers/features/common-utils:2"
  },
  {
    "id": "ghcr.io/devcontainers/features/git:1"
  },
  {
    "remoteUser": "vscode"
  }
]
```

> What if the remoteUser is not set, but the dev container will run with the container user? Should ${remoteUser} be replaced with the container user? We also do that for the _REMOTE_USER variable.

That's a good point that I did not think of, perhaps it should align with those variable (info here https://containers.dev/implementors/features/#user-env-var)

max06: Hello @chrmarti, hello @joshspicer,

I understood the assignment. Configuration options can come from many different sources. Relying on the devcontainer.json to contain them is not the proper way to go. 

Usually I'm pretty good at reading foreign code and getting the gist of it. At least enough to be able to provide smaller bugfixes and features. This project here... doesn't work very well with my brain. I'm having a hard time following along the different workflows without seeing the red string showing a common structure. Most likely the lack of experience on my side, I'm more ops than dev.

I'll give it another shot next weekend, unless one of you already knows where to use the screwdriver. Feel free to do whatever needs to be done, I definitely won't complain.

PR #530, opened 2023-05-22
Feature Dependencies

joshspicer:
ref: https://github.com/devcontainers/spec/issues/109
implements proposal: https://github.com/devcontainers/spec/pull/234

_NOTE: This is a large diff, although the majority of the edits can be attributed to tests.  The commits are organized logically by changes, and is likely the best way to view this change_.

This change modifies Feature installation order.  Namely, the effect that the `dependsOn`, `installsAfter`, and `overrideFeatureInstallOrder` properties have on installation order.  The default installation order (with no additional properties) may also change with this change.  

The large majority of the logical changes can be found in the [`containerFeaturesOrder.ts`](https://github.com/devcontainers/cli/blob/joshspicer/dependsOn-staging/src/spec-configuration/containerFeaturesOrder.ts) file, which has been completely rewritten.

The algorithm in this file in invoked as shown below.

https://github.com/devcontainers/cli/blob/e635d0486134827f5a4f72c9ba50c6848dc57e03/src/spec-configuration/containerFeaturesConfiguration.ts#L554

`installationOrder` contains featureSets in the correct order of installation - those are passed forward to the remainder of the Features installation code. Everything else following works exactly as it had before.

https://github.com/devcontainers/cli/blob/joshspicer/dependsOn-staging/src/spec-configuration/containerFeaturesConfiguration.ts#L560-L564

Some notable effects of this change are that:
- Features with a `dependsOn` property will _recursively_ fetch and add its dependencies to the list of Features to install.  If the algorithm encounters a cycle or invalid state, it will fail.
- Due to the definition of `dependsOn`, the same Feature may be installed more than once (provided it has either a different version _or_ different options).  Existing Features may not be expecting this behavior and may need to be updated before being declared as a dependency.


chrmarti (changes requested):
Looks great. Will add more questions as I make my way through it!

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+		};
+		const processed = await processFeature(overrideNode);
+		if (!processed) {
+			throw new Error(`Feature '${userFeatureId}' in 'overrideFeatureInstallOrder' could not be processed.`);
+		}
+
+		if (featureSupportsDependencies(processed)) {
+			overrideNode.featureSet = processed;
+			overrideFeatureNodes.push(overrideNode);
+			config.overrideFeatureInstallOrder.splice(j, 1);
```
chrmarti, line 216: Why remove it?

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+	overrideFeatureNodes.reverse();
+	output.write(`[override-feature-nodes]: ${overrideFeatureNodes.map(n => n.userFeatureId).join(', ')}`, LogLevel.Info);
+
+	for (let i = 0; i < overrideFeatureNodes.length - 1; i++) {
+		// Scan the resolvedGraph for a node that matches the next override node.
+		const currentOverrideNode = overrideFeatureNodes[i];
+		const nextOverrideNode = overrideFeatureNodes[i + 1];
+		let countMatches = 0;
+		for (const node of worklist) {
+			if (satisfiesSoftDependency(node, nextOverrideNode)) {
```
chrmarti, line 229: Somehow we're not looking at nodes with `satisfiesSoftDependency(node, overrideFeatureNodes[0])`?

chrmarti: I guess we need another loop looking at `satisfiesSoftDependency(node, currentOverrideNode)` and adding `currentOverrideNode.installsAfter.push(node)`?

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+			&& !satisfiesSoftDependency(node, lastOverrideNode) // Prevent cycles with itself
+			&& overrideFeatureNodes.every(override => !satisfiesSoftDependency(node, override))) // Prevent cycles with other override nodes 
+		{
+			output.write(`[override-feature-nodes-tail]: ${node.userFeatureId} installsAfter ${lastOverrideNode.userFeatureId}`, LogLevel.Info);
+			node.installsAfter.push(lastOverrideNode);
+		}
+	}
+
+	// Return the modified worklist.
+	return worklist;
```
chrmarti, line 258: So this does not include the `overrideFeatureNodes`.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+	}
+
+	// Add the final node to each leaf node's soft-dependency adjacency list (Enforcing override Features are installed first),
+	// Ensuring that such an act does not introduce a cycle.
+	const lastOverrideNode = overrideFeatureNodes[overrideFeatureNodes.length - 1];
+	for (const node of worklist) {
+		if (node.installsAfter.length === 0 //  TODO: necessary?
+			&& node.dependsOn.length === 0 // Leaf node
+			&& !satisfiesSoftDependency(node, lastOverrideNode) // Prevent cycles with itself
+			&& overrideFeatureNodes.every(override => !satisfiesSoftDependency(node, override))) // Prevent cycles with other override nodes
```
chrmarti, line 250: This only detects cycles with 2 nodes involved. We detect cycles later and report them. Should we leave the handling to the user? (A few lines up we don't try to detect them either, but these additons to `installsAfter` could also introduce additional cycles.)

chrmarti (changes requested):

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -542,21 +521,18 @@ function updateFromOldProperties<T extends { features: (Feature & { extensions?:
 export async function generateFeaturesConfig(params: ContainerFeatureInternalParams, dstFolder: string, config: DevContainerConfig, getLocalFeaturesFolder: (d: string) => string, additionalFeatures: Record<string, string | boolean | Record<string, string | boolean>>) {
 	const { output } = params;
 
+	// Use experimental feature dependencies if explicitly enabled via flag, or if running in CI
+	const useExperimentalFeatureDependencies = params.experimentalFeatureDependencies || !!process.env['CI'];
```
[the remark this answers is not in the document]
joshspicer: My intention here was to run all the CI tests without needing to run the entire set twice. We already pass in this variable when running in ci, but I understand this is a bit hacky.

What do you think would be best?

1) Run _all_ the tests twice, once with the FF and one without.
2) Keep as is (running ci tests just against this new code path).
3) Some other third thing?

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: I remove it here so that later on when this object is used for legacyFeatures, we have spliced out all the overrides that have been processed in the dependencies algorithm.

src/spec-configuration/containerFeaturesConfiguration.ts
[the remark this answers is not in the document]
joshspicer: Another option is to remove the FF altogether.  Functionally this change is supposed to be the same as the existing implementation when Features don't declare a `dependsOn` property.

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: The goal of this is to implement the suggestion here https://github.com/devcontainers/spec/pull/234#discussion_r1183470261

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+					const legacyIds = (softDepMetadata.legacyIds || []).concat(softDepMetadata.id);
+					dependency.legacyIdAliases = legacyIds;
+				}
+
+				current.installsAfter.push(dependency);
+			}
+		}
+
+		acc.push(current);
+		await _buildDependencyGraph(params, processFeature, worklist, acc, legacyAcc);
```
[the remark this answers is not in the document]
joshspicer: It appears we don't, that seemed to just be doing extra work.

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: Every Feature indicated in `overrideFeatureNodes` should already be present in the `worklist` (That's the goal of [this invariant check](https://github.com/devcontainers/cli/pull/530/files#diff-5d796ceb58c276912ce5289b37ea9eda883f4abbf9b6cc1252661deccae91e67R237-R240), although I think that check needs some work).  The work done above this return is adding `installsAfter` references to the nodes in the `worklist` to influence install order in the next stage.

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: This has been removed in favor of a priority-based approach evaluated per-round. See https://github.com/devcontainers/cli/pull/530/commits/c8bfd1b40c50603da78d9066c80f10e3669d158c

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: This has been removed in favor of a priority-based approach evaluated per-round. See https://github.com/devcontainers/cli/commit/c8bfd1b40c50603da78d9066c80f10e3669d158c

src/spec-configuration/containerFeaturesConfiguration.ts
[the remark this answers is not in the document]
chrmarti: I was thinking about merging the change in behind the experimental flag. When we do that the env variable shouldn't be a common one like 'CI' because that might well be set in other CIs.

We could keep it as-is and remove it before merging to `main`.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+			if (aResource !== bResource) {
+				return aResource.localeCompare(bResource);
+			}
+
+			const aTag = aSourceInfo.featureRef.tag;
+			const bTag = bSourceInfo.featureRef.tag;
+			// Sort by tags (if both have tags)
+			// Eg: 1.9.9, 2.0.0, 2.0.1, 3, latest
+			if ((aTag && bTag) && (aTag !== bTag)) {
+				return aTag.localeCompare(bTag);
```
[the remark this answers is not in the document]
joshspicer: It will result in a different sort since we're using the unresolved tags.  We could choose here not to compare the tags and instead compare solely on digest. The resolved digest will change unpredictably each release of the Feature. This feels more stable and intuitive to me, but at the end of the day the "stable order" here can be whatever we think is most intuitive.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,538 @@
...
+
+			default:
+				throw new Error(`Feature dependencies are only supported in Features published (2) to an OCI registry, (2) as direct HTTPS tarball, or (3) as a local Feature.  Got type: '${processedFeature.sourceInformation.type}'.`);
+		}
+
+		// Resolve dependencies given the current Feature's metadata.
+		if (metadata) {
+			current.featureSet.features[0] = {
+				...current.featureSet.features[0],
+				...metadata,
```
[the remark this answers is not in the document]
joshspicer: Good catch, working on this.

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: You suggest removing this check entirely, as we won't be able to detect all cycles at this stage anyway?

joshspicer: This code was removed.

src/spec-configuration/containerFeaturesOrder.ts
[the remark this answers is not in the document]
chrmarti: Having the latest version run last might have the advantage of that getting a chance at arranging things last, e.g., when previous versions have a bug.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,631 @@
...
+		worklist.splice(0, worklist.length, ...worklist.filter(node => !round.some(r => equals(params, r, node))));
+
+		// Sort rounds lexicographically by id.
+		round.sort((a, b) => comparesTo(params, a, b));
+
+		output.write(`[round-sort-id] ${round.map(r => r.userFeatureId).join(', ')}`, LogLevel.Trace);
+
+		// Final sort of rounds by priority
+		// Higher priority nodes are installed earlier.
+		round.sort((a, b) => b.roundPriority - a.roundPriority);
```
chrmarti, line 579: I think `round` needs to be filtered down to the nodes with the highest `roundPriority` otherwise `overrideFeatureInstallOrder` doesn't have the global effect it is meant to have.

E.g., if B depends on A and C is independent and `overrideFeatureInstallOrder` is A, B, C, the above would install A and C in the first round and B in the second. If in each round you only install the highest priority nodes, the first round will install A, the second round will install B and only the third round will install C.

src/spec-configuration/containerFeaturesOrder.ts
chrmarti: I would keep the original unchanged to keep it easier to understand in code and in the debugger. Can we modify a copy and pass that around to avoid modifying the original array?

Or better: Remove the legacy algorithm and let the new one handle everything.

chrmarti (changes requested):

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: Ok, I will make that change.  I was hoping to reduce the complexity by not supporting deprecated concepts in the new algorithm.  This was the tradeoff, but perhaps now that i'm at this point it would not be too difficult. Perhaps that can be a later improvement - I would love to delete the previous code if it didn't involve a lot more work :D

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: I see, so the algorithm 

- From `dependsOn

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,631 @@
...
+
+		output.write(`[round-sort-priority] ${round.map(r => `${r.userFeatureId} (${r.roundPriority})`).join(', ')}`, LogLevel.Trace);
+
+		installationOrder.push(...round);
+	}
+
+	// Handle legacy worklist
+	const legacy = computeFeatureInstallationOrder_deprecated(config, legacyWorklist.reverse().map(n => n.featureSet!));
```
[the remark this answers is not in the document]
joshspicer: We also need to define an `equals`, `comparesTo`, and `satisifiesSoftDependency`

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: Ok, I think I understand the distinction.  I agree that overrideFeatureInstallOrder is only evaluated locally per round given my current implementation.

I will capture this example in a test and make the change.

joshspicer: Fixed in https://github.com/devcontainers/cli/pull/530/commits/9b8eb0f111bc44c2466e5f0b4fc1661669fc5a85

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -507,8 +507,11 @@ export async function computeDependsOnInstallationOrder(
 
 	const { output } = params;
 
+	// Make a copy of config to avoid mutating the original.
+	const configCopy = { ...config };
```
chrmarti, line 511: This is a shallow copy, the array you modify will still be shared.

Maybe do the copy in the same function (`buildDependencyGraph` I believe) that modifies the array and also copy the array there and then let that function work on the copy, so the code here does not need to know that the config gets modified at all?

chrmarti (changes requested):

src/spec-configuration/containerFeaturesOrder.ts
joshspicer: The goal here was to mutate only configCopy, so that here it only contains the IDs of 'legacy Features'
https://github.com/devcontainers/cli/blob/61de1857a7945e977e010fb366fa4e3d0c04546f/src/spec-configuration/containerFeaturesOrder.ts#L596

Otherwise, the legacy algorithm will detect a non-legacy Feature and fail.

joshspicer: I am attempting to remove the legacyWorklist and associated algorithm per your suggestion.  If I do that then I won't need to take a copy of the config.

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,626 @@
...
+			// Compare values
+			const aVal = a[aKeys[i]];
+			const bVal = b[bKeys[i]];
+			if (typeof aVal === 'string' && typeof bVal === 'string') {
+				const v = aVal.localeCompare(bVal);
+				if (v !== 0) {
+					return v;
+				}
+			}
+			if (typeof aVal === 'boolean' && typeof bVal === 'boolean') {
```
chrmarti, line 139: Could use `else if` here and below (also a small readability improvement).

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -3,133 +3,626 @@
...
+				const v = aVal === bVal ? 0 : aVal ? 1 : -1;
+				if (v !== 0) {
+					return v;
+				}
+			}
+
+			if (typeof aVal === undefined) {
```
chrmarti, line 146: Edge case: aVal and bVal could be both undefined.

chrmarti (changes requested):
Two minor comments. LGTM otherwise!

src/spec-configuration/containerFeaturesOrder.ts
```diff
@@ -142,13 +142,11 @@ function optionsCompareTo(a: string | boolean | Record<string, string | boolean
...
 				}
 			}
-
-			if (typeof aVal === undefined) {
-				return 1;
-			}
-
-			if (typeof bVal === undefined) {
-				return -1;
+			if (typeof aVal === undefined && typeof bVal === undefined) {
```
chrmarti, line 145:
```suggestion
			if (typeof aVal === 'undefined' || typeof bVal === 'undefined') {
```

chrmarti (changes requested):

src/spec-configuration/containerFeaturesOrder.ts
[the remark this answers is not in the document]
chrmarti: Maybe instead of using `userFeatureId` directly (e.g., in `compareTo`), you could have a function `normalizedFeatureId(feature: FeatureSet)` that returns the current feature id? I guess that only matters for OCI features, in all other cases this function would return `userFeatureId`?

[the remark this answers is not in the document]
joshspicer: Ah I am just seeing this now.  I did something similar where it considers all combination of legacyIds and the current Id.

chrmarti (approved):

Issue #478, opened 2023-04-01
Issue: Unnamed docker volumes


max06:
(This is probably in the wrong repository - I'm not sure where to put this, please move it to the right people)

Specifying a mountpoint without a source-attribute creates a named volume named `undefined`. According to the spec (and to the same mount used in a `devcontainer.json`), it should create an unnamed volume.

Example:
```jsonc
    "mounts": [
        {
            "target": "/home/vscode/.vscode-server",
            "type": "volume"
        }
    ]
```

ends up as:

```text
            {
                "Type": "volume",
                "Name": "undefined",
                "Source": "/var/lib/docker/volumes/undefined/_data",
                "Destination": "/home/vscode/.vscode-server",
                "Driver": "local",
                "Mode": "z",
                "RW": true,
                "Propagation": ""
            }
```

samruddhikhandale:
Hi 👋 

Thanks for reporting. 

According to this [docker documentation](https://docs.docker.com/storage/#:~:text=Anonymous%20volumes%20are%20not%20given,within%20a%20given%20Docker%20host.),

>  Anonymous volumes are not given an explicit name when they are first mounted into a container, so Docker gives them a random name that is guaranteed to be unique within a given Docker host.

Looks like a bug in the CLI - if the `source` is omitted, then I expect a new volume with random unique name should be created for volume mounts. This should definitely not be named as `undefined`

max06:
Hey 👋🏼 

if I'm not mistaken, your schema requires the `source`-parameter for mounts. The bug in the cli might be its result.

max06:
Hey there, @samruddhikhandale 
not trying to pressure anyone, just interested if there's any progress on this topic?

samruddhikhandale:
Thanks for the ping on this issue. 

@alexander-smolyakov can you help prioritize this issue? We need a bug fix for this issue for the dev containers CLI.

max06:
Closing this issue as it was fixed in #537 some time ago. Passed right by me unnoticed. 
I just tested it, I can confirm the solution, and I'm very happy about it! Thank you very much! 🥳

PR #537, opened 2023-05-23
Rework `mount` command string generation

alexander-smolyakov:
**Description**:

This PR addresses the issue when the Docker volume is named as `undefined` if the `src` argument is not specified in the `mount` command.

The issue related to the following code:

https://github.com/devcontainers/cli/blob/7de4011ca1d21781730198f718ff449c1772ebe5/src/spec-node/singleContainer.ts#L359-L364

When the `mount` command string is generated, there is no check for the `src` argument. This results in the `src` attribute being set to `undefined` when cast to a string.

To fix this behavior, we need to apply null-check for the `src` argument and don't pass it to Docker. When the `src` arg is not passed, Docker generates a random name for the volume.

**Repro sample**:

```json
{
	"image": "mcr.microsoft.com/devcontainers/base:latest",
	"mounts": [
        {
            "target": "/home/vscode/.vscode-server",
            "type": "volume",
        }
    ]
}
```

**Examples**:
_Without fix_:

```console
node ➜ /workspaces/devcontainers-cli (main) $ docker inspect -f '{{json .Mounts }}' a85f6e52d478
```

```json
[
	{
		"Type": "bind",
		"Source": "/workspaces/devcontainers-cli",
		"Destination": "/workspaces/devcontainers-cli",
		"Mode": "",
		"RW": true,
		"Propagation": "rprivate"
	},
	{
		"Type": "volume",
		"Name": "undefined",
		"Source": "/var/lib/docker/volumes/undefined/_data",
		"Destination": "/home/vscode/.vscode-server",
		"Driver": "local",
		"Mode": "z",
		"RW": true,
		"Propagation": ""
	}
]
```

_With fix_:

```console
node ➜ /workspaces/devcontainers-cli (users/alexander-smolyakov/add_proper_handling_for_src_arg_in_mount_command) $ docker inspect -f '{{json .Mounts }}' 2e77ff5392dc
```

```json
[
	{
		"Type": "bind",
		"Source": "/workspaces/devcontainers-cli",
		"Destination": "/workspaces/devcontainers-cli",
		"Mode": "",
		"RW": true,
		"Propagation": "rprivate"
	},
	{
		"Type": "volume",
		"Name": "628f29020e5d6a9514614ce0ad90411d2400161b67e07bb8744f39a83a397922",
		"Source": "/var/lib/docker/volumes/628f29020e5d6a9514614ce0ad90411d2400161b67e07bb8744f39a83a397922/_data",
		"Destination": "/home/vscode/.vscode-server",
		"Driver": "local",
		"Mode": "z",
		"RW": true,
		"Propagation": ""
	}
]
```

**Attached related issue**:

- #478

**Changelog**:
- Move the `mount` command generation logic to a separate function;
- Add check if the `src` argument is undefined;

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -710,6 +710,7 @@ function convertMountToVolume(mount: Mount): string {
  * Convert mount command' arguments to volume top-level element
  * @param mount 
  * @returns mount object representation as volumes top-level element
+ * @throws if `source` property is undefined
  */
 function convertMountToVolumeTopLevelElement(mount: Mount): string {
 	let volume: string = `
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -500,7 +500,7 @@ async function generateFeaturesComposeOverrideContent(
 		...mergedConfig.mounts || [],
 		...additionalMounts,
 	].map(m => typeof m === 'string' ? parseMount(m) : m);
-	const namedVolumeMounts = mounts.filter(m => m.type === 'volume' && m.source);
+	const volumeMounts = mounts.filter(m => m.type === 'volume');
 	const customEntrypoints = mergedConfig.entrypoints || [];
 	const composeEntrypoint: string[] | undefined = typeof service.entrypoint === 'string' ? shellQuote.parse(service.entrypoint) : service.entrypoint;
 	const composeCommand: string[] | undefined = typeof service.command === 'string' ? shellQuote.parse(service.command) : service.command;
@@ -543,9 +543,9 @@ while sleep 1 & wait $$!; do :; done", "-"${userEntrypoint.map(a => `, ${JSON.st
     labels:${additionalLabels.map(label => `
       - '${label.replace(/\$/g, '$$$$').replace(/'/g, '\'\'')}'`).join('')}` : ''}${mounts.length ? `
     volumes:${mounts.map(m => `
-      - ${convertMountToVolume(m)}`).join('')}` : ''}${gpuResources}${namedVolumeMounts.length ? `
-volumes:${namedVolumeMounts.map(m => `
-  ${convertMountToVolumeTopLevelElement(m)}`).join('')}` : ''}
+      - ${convertMountToVolume(m)}`).join('')}` : ''}${gpuResources}${volumeMounts.length ? `
+volumes:${volumeMounts.map(m => `
+  ${m.source}:${m.external ? '\n    external: true' : ''}`).join('')}` : ''}
 `;
 }
 
@@ -702,24 +702,6 @@ function convertMountToVolume(mount: Mount): string {
 	}
 
 	volume += mount.target;
-
-	return volume;
-}
-
-/**
- * Convert mount command' arguments to volume top-level element
- * @param mount 
- * @returns mount object representation as volumes top-level element
- * @throws if `source` property is undefined
- */
-function convertMountToVolumeTopLevelElement(mount: Mount): string {
-	let volume: string = `
-  ${mount.source}
-`;
-
-	if (mount.external) {
-		volume += '\n    external: true';
-	}
 
 	return volume;
 }
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -18,7 +18,7 @@ import { getExtendImageBuildInfo, updateRemoteUserUID } from './containerFeature
 import { Mount, parseMount } from '../spec-configuration/containerFeaturesConfiguration';
 import path from 'path';
 import { getDevcontainerMetadata, getImageBuildInfoFromDockerfile, getImageBuildInfoFromImage, getImageMetadataFromContainer, ImageBuildInfo, lifecycleCommandOriginMapFromMetadata, mergeConfiguration, MergedDevContainerConfig } from './imageMetadata';
-import { ensureDockerfileHasFinalStageName } from './dockerfileUtils';
+import { ensureDockerfileHasFinalStageName, convertMountToVolume } from './dockerfileUtils';
 
 const projectLabel = 'com.docker.compose.project';
 const serviceLabel = 'com.docker.compose.service';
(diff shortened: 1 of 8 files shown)


samruddhikhandale (commented):
Thanks, can you help add a unit test to validate this fix?

alexander-smolyakov:
> Thanks, can you help add a unit test to validate this fix?

Added test in https://github.com/devcontainers/cli/pull/537/commits/7fc4270e4416c0725b14e17d0415e7bb754d4efb

joshspicer (dismissed):
Looks good to me - thanks @alexander-smolyakov !

joshspicer:
Here's the spec article on anonymous volumes https://docs.docker.com/storage/volumes/#choose-the--v-or---mount-flag. Not sure why this is particularly useful, I guess so as to not clutter up a container's filesystem between starts/stops?

chrmarti (changes requested):
Thanks for looking into this!

- Could you make `source` optional in the interface: https://github.com/devcontainers/cli/blob/9064c705687bd214c796aaeeacba260bfa7c3ae3/src/spec-configuration/containerFeaturesConfiguration.ts#L98
- Handle the two references to source for Docker Compose: https://github.com/devcontainers/cli/blob/9064c705687bd214c796aaeeacba260bfa7c3ae3/src/spec-node/dockerCompose.ts#L546

We will also need to update the schemas to make source optional: https://github.com/devcontainers/spec/tree/main/schemas

src/spec-node/dockerCompose.ts
```diff
@@ -543,7 +543,7 @@ while sleep 1 & wait $$!; do :; done", "-"${userEntrypoint.map(a => `, ${JSON.st
     labels:${additionalLabels.map(label => `
       - '${label.replace(/\$/g, '$$$$').replace(/'/g, '\'\'')}'`).join('')}` : ''}${mounts.length ? `
     volumes:${mounts.map(m => `
-      - ${m.source}:${m.target}`).join('')}` : ''}${gpuResources}${volumeMounts.length ? `
+      - ${convertMountToVolume(m)}`).join('')}` : ''}${gpuResources}${volumeMounts.length ? `
 volumes:${volumeMounts.map(m => `
   ${m.source}:${m.external ? '\n    external: true' : ''}`).join('')}` : ''}
```
chrmarti, line 548: Another case here. Not sure if anonymous volumes can be external.

chrmarti (changes requested):

src/spec-node/dockerCompose.ts
alexander-smolyakov: > Not sure if anonymous volumes can be external.

This should not work for anonymous volumes since we need to set the path in the `docker-compose.yml` file explicitly.

Related doc: https://docs.docker.com/compose/compose-file/07-volumes/#external

Could you please provide guidance regarding how to handle this case better? 
It seems we can fire an exception if `source` is undefined.

src/spec-node/dockerfileUtils.ts
```diff
@@ -292,3 +292,20 @@ export function generateMountCommand(mount: Mount | string): string[] {
 
 	return [command, args];
 }
+
+/**
+ * Convert mount command' arguments to Docker Compose volume
+ * @param mount 
+ * @returns mount command representation for Docker compose
+ */
+export function convertMountToVolume(mount: Mount): string {
```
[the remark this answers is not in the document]
alexander-smolyakov: Got it, thanks!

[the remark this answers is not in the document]
alexander-smolyakov: Fixed in 17a3f98acf99aa096d737babb56a939544533d43

src/spec-node/dockerCompose.ts
alexander-smolyakov: Fixed in 5a202a14a4f523487b79cf9a3275ac8d748f76d7

samruddhikhandale (dismissed):
Looks great, thank you!

src/spec-node/dockerCompose.ts
```diff
@@ -704,4 +704,22 @@ function convertMountToVolume(mount: Mount): string {
 	volume += mount.target;
 
 	return volume;
-}
\ No newline at end of file
+}
+
+/**
+ * Convert mount command' arguments to volume top-level element
```
chrmarti, line 710: nit: extra '

src/spec-node/dockerCompose.ts
```diff
@@ -704,4 +704,22 @@ function convertMountToVolume(mount: Mount): string {
...
+ * @param mount 
+ * @returns mount object representation as volumes top-level element
+ * @throws if `source` property is undefined
```
chrmarti, line 713: It currently doesn't throw.

src/spec-node/dockerCompose.ts
```diff
@@ -704,4 +704,22 @@ function convertMountToVolume(mount: Mount): string {
...
+ */
+function convertMountToVolumeTopLevelElement(mount: Mount): string {
+	let volume: string = `
+  ${mount.source}
+`;
```
chrmarti, line 718: nit: This seems to include an extra newline compared to before.

src/spec-node/dockerCompose.ts
```diff
@@ -688,3 +688,38 @@ export function dockerComposeCLIConfig(params: Omit<PartialExecParameters, 'cmd'
 		})());
 	};
 }
+
+/**
+ * Convert mount command' arguments to Docker Compose volume
```
chrmarti, line 693: nit: extra '

chrmarti (changes requested):
Looks good. A few polishing suggestions. Thanks.

src/spec-node/dockerCompose.ts
alexander-smolyakov: Fixed in d2510412db8afb196ab6e8f852021de2338d8c30

src/spec-node/dockerCompose.ts
alexander-smolyakov: Fixed in d2510412db8afb196ab6e8f852021de2338d8c30

src/spec-node/dockerCompose.ts
alexander-smolyakov: Fixed in d2510412db8afb196ab6e8f852021de2338d8c30

chrmarti (approved):

PR #541, opened 2023-06-01
Automatically set execute bit when running dotfiles install script

joshspicer:
The bootstrap script will add the execute bit to the 'installCommand' when cloning dotfiles, if it is not already set as executable.  This reduces friction in setting up dotfiles.

https://github.com/codspace/test-dotfiles/commit/0a444c1d694432b6e102e896651c317be67c7d19 removes the execute bit from the existing test.

ref: https://github.com/github/codespaces/issues/13870


src/spec-common/dotfiles.ts
```diff
@@ -45,14 +45,14 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
...
-	echo Executing script ${installCommand}
-	cd ${targetPath} && ${env}${installCommand}
-else
-	echo Error: ${installCommand} not executable
-	exit 126
+	echo Setting '${targetPath}'/'${installCommand}' as executable
+	chmod +x ${installCommand}
 fi
+echo Executing script ${installCommand}
+${env}./${installCommand}
```
samruddhikhandale, line 55: Curious, why did we change `${env}${installCommand}` to `${env}./${installCommand}` ?

samruddhikhandale (dismissed):

src/spec-common/dotfiles.ts
joshspicer: I changed it here to align with the other block.  Functionally it should be no different

chrmarti: This would break cases, e.g., that use `~/targetPath/installScript.sh`. Should we keep the previous variant to minimize the risk of breaking existing configurations? (Below we know that `$installCommand` is a filename without path component.)

chrmarti (changes requested):
Looks good, added to the previous conversation.

chrmarti (approved):

src/spec-common/dotfiles.ts
joshspicer: It appears this flag had been broken unless you provided it an absolute path. 
image

I've made changes to support both just a file name in the root of the dotfiles directory, as well as an absolute path (if the former doesn't resolve as a file).

I added several new tests to confirm.

src/spec-common/dotfiles.ts
```diff
@@ -44,20 +44,34 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
 ${createFileCommand(markerFile)} || (echo dotfiles marker found && exit 1) || exit 0
 command -v git >/dev/null 2>&1 || (echo git not found && exit 1) || exit 0
 [ -e ${targetPath} ] || ${allEnv}git clone ${repository} ${targetPath} || exit $?
-if [ -x ${installCommand} ]
+echo Setting current directory to ${targetPath}
+cd ${targetPath}
+
+if [ -f ${targetPath}/${installCommand} ]
```
chrmarti, line 50: With `cd ${targetPath}` above you can cover the absolute and relative path both with `if [ -f "${installCommand}" ]` and use the `else` to exit with an error message because the file should exist.

chrmarti (changes requested):

src/test/cli.up.test.ts
```diff
@@ -45,6 +45,61 @@ describe('Dev Containers CLI', function () {
 			await shellExec(`docker rm -f ${containerId}`);
 		});
 
+		it('should execute successfully with valid config and dotfiles with custom install path as filename', async () => {
```
chrmarti, line 48: I think we forgot to move the existing tests over: https://github.com/microsoft/vscode-remote-containers/blob/0be791aae106ec5bb5e1c57fd9edf13b62f4d18a/test/core/dotfiles.test.ts#L23

chrmarti (changes requested):

src/test/cli.up.test.ts
chrmarti: Thanks!

src/spec-common/dotfiles.ts
chrmarti: Agreed, I thought that would be covered by the `cd` you added. 👍

src/spec-common/dotfiles.ts
```diff
@@ -44,20 +44,34 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
...
+	then
+		echo Setting '${targetPath}'/'${installCommand}' as executable
+		chmod +x ${targetPath}/${installCommand}
+	fi
+	echo Executing command '${targetPath}/${installCommand}'...\n
+	${allEnv}${targetPath}/${installCommand}
 else
-	echo Error: ${installCommand} not executable
-	exit 126
+	if [ ! -x ${installCommand} ]
```
chrmarti, line 60: Add `elif [ -f "${installCommand}" ]` here and have an `else` handling when the file does not exist.

src/spec-common/dotfiles.ts
```diff
@@ -44,20 +44,34 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
...
 [ -e ${targetPath} ] || ${allEnv}git clone ${repository} ${targetPath} || exit $?
-if [ -x ${installCommand} ]
+echo Setting current directory to ${targetPath}
+cd ${targetPath}
+
+if [ -f ${targetPath}/${installCommand} ]
 then
-	echo Executing script ${installCommand}
-	cd ${targetPath} && ${allEnv}${installCommand}
+	if [ ! -x ${targetPath}/${installCommand} ]
```
chrmarti, line 52: We should quote `"${targetPath}/${installCommand}"` (and other variable usages), so things don't break when there are spaces in the path or filename.

chrmarti (changes requested):

src/spec-common/dotfiles.ts
joshspicer: I've tried several techniques but I think the way I have it now is the most correct, otherwise it will be unhappy with one of the following examples:

1. `install.sh`
2. `./install.sh`
3. `/home/node/dotfiles/install.sh`

src/test/cli.up.test.ts
joshspicer: I refactored and wrote several new tests under dotfiles.test.ts.  I think this is solid coverage and moving over this test verbatim doesn't add any additional value.

src/spec-common/dotfiles.ts
```diff
@@ -40,16 +40,33 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
...
 then
-	echo Executing script ${installCommand}
-	cd ${targetPath} && ${allEnv}${installCommand}
+	if [ ! -x "${installCommand}" ]
+	then
+		echo Setting '${installCommand}' as executable
+		chmod +x ${installCommand}
+	fi
+	echo Executing command '${installCommand}'...\n
+	${allEnv}${installCommand}
```
chrmarti, line 67: nit: Also quote here.

src/spec-common/dotfiles.ts
```diff
@@ -40,16 +40,33 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
...
+	echo Executing command './${installCommand}'..\n
+	${allEnv}"./${installCommand}"
+elif [ -f "${installCommand}" ]
 then
-	echo Executing script ${installCommand}
-	cd ${targetPath} && ${allEnv}${installCommand}
+	if [ ! -x "${installCommand}" ]
+	then
+		echo Setting '${installCommand}' as executable
+		chmod +x ${installCommand}
```
chrmarti, line 64: nit: Also quote here.

src/spec-common/dotfiles.ts
```diff
@@ -40,16 +40,33 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
 			status: 'running',
 		});
 		if (installCommand) {
-			await shellServer.exec(`# Clone & install dotfiles
+			await shellServer.exec(`# Clone & install dotfiles via '${installCommand}'
 ${createFileCommand(markerFile)} || (echo dotfiles marker found && exit 1) || exit 0
 command -v git >/dev/null 2>&1 || (echo git not found && exit 1) || exit 0
 [ -e ${targetPath} ] || ${allEnv}git clone ${repository} ${targetPath} || exit $?
```
chrmarti, line 46: Actually `${targetPath}` should probably also be quoted to not break on paths with spaces. We overlooked these cases previously. Same with `${repository}`, more cases further down.

chrmarti (changes requested):
More missing quotes, we missed some previously, so probably not critical, but would be good to fix now that we are looking into it.

chrmarti:
> Using both single quotes or double quotes to quote `targetPath `during the `git clone ${targetPath}` commands don't work well with shell characters (`~`) used in the tests. I tried manually making the directory as well, but that doesn't work well either (still unhappy when quoting the

Ack, I wasn't aware of that. Let's keep it unquoted then for now, not supporting spaces in the path as we are doing now. (One option to support both might be to check for a leading `~` and put it outside quotes. Or escape spaces with a backslash.)

Maybe still make quoting of `installCommand` consistent (see my unresolved comments).

joshspicer:
Yes, made the `installCommand` quoting changes (didn't realize I hadn't pushed that).

chrmarti (approved):

PR #544, opened 2023-06-02
Pass secret values correcly to dotfile clone and install

eljog:
There was a bug, where we were not passing secrets correctly to the dotfiles clone and install command. This PR fixes it

### DIFF
--- a/src/test/cli.up.test.ts
+++ b/src/test/cli.up.test.ts
@@ -17,6 +17,7 @@ describe('Dev Containers CLI', function () {
 	const tmp = path.relative(process.cwd(), path.join(__dirname, 'tmp'));
 	const cli = `npx --prefix ${tmp} devcontainer`;
 
+
 	before('Install', async () => {
 		await shellExec(`rm -rf ${tmp}/node_modules`);
 		await shellExec(`mkdir -p ${tmp}`);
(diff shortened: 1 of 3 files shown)


src/test/cli.up.test.ts
```diff
@@ -44,6 +45,25 @@ describe('Dev Containers CLI', function () {
...
+		it('should execute successfully with valid config and dotfiles with secrets', async () => {
+			const testFolder = `${__dirname}/configs`;
+			let containerId: string | null = null;
+			await shellExec(`rm -f ${testFolder}/*.testMarker`, undefined, undefined, true);
+			const secrets = {
+				'SECRET1': 'SecretValue1',
+			};
+			await shellExec(`printf '${JSON.stringify(secrets)}' > ${testFolder}/test-secrets-temp.json`, undefined, undefined, true);
+
+			const res = await shellExec(`${cli} up --workspace-folder ${__dirname}/configs/image-with-git-feature --dotfiles-repository https://github.com/codspace/test-dotfiles --secrets-file ${testFolder}/test-secrets-temp.json`);
```
joshspicer, line 57: Could we perhaps update the test repo to ensure we're passing these variables properly?

eljog: This test is very basic. It only tests dotfiles still work with secrets file.
May be it is a good idea to make the dotfiles fail if expected secrets are not passed
Let me check

eljog: Test added in https://github.com/devcontainers/cli/pull/544/commits/55ba3f86f2e329e16d75bbc7fc75d4e057d9218f

joshspicer (approved):
Awesome!

PR #551, opened 2023-06-14
Mask secrets when writing logs

eljog:
Secrets must be handled securely at all times and should not be written in plain text into logs.


src/spec-common/secrets.ts
```diff
@@ -0,0 +1,45 @@
...
+			description: 'Failed to read/parse secrets',
+			originalError: e
+		});
+	}
+}
+
+export async function maskSecrets(secretsP: Promise<Record<string, string>>, text: string) {
+	const secretValues = Object.values(await secretsP);
+	secretValues.forEach(secret => {
+		const regex = new RegExp(secret, 'g');
```
chrmarti, line 40: `secret` needs some escaping. Maybe extract a helper function for common use from: https://github.com/devcontainers/cli/blob/236b85162a945a1af9e72fcbe02eb5c7c864b31d/src/spec-shutdown/dockerUtils.ts#L349.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -200,6 +201,12 @@ async function provision({
 	const addCacheFroms = addCacheFrom ? (Array.isArray(addCacheFrom) ? addCacheFrom as string[] : [addCacheFrom]) : [];
 	const additionalFeatures = additionalFeaturesJson ? jsonc.parse(additionalFeaturesJson) as Record<string, string | boolean | Record<string, string | boolean>> : {};
 	const providedIdLabels = idLabel ? Array.isArray(idLabel) ? idLabel as string[] : [idLabel] : undefined;
+
+	const cwd = workspaceFolder || process.cwd();
+	const cliHost = await getCLIHost(cwd, loadNativeModule);
+	const secretsP = readSecretsFromFile({ secretsFile, cliHost });
+	const log = async (text: string) => process.stderr.write(await maskSecrets(secretsP, text));
```
chrmarti, line 208: The output here can be JSON. That creates the edge case of a secret matching a JSON property (or some other part of the JSON syntax). Maybe add a helper function to log.ts: `function replaceAllLog(origin: LogHandler, values: string[], replacement: string): LogHandler`. (See `createCombinedLog` in that file for an example.)

`replaceAllLog` can apply the replacements to the log event's text and pass a modified log event on. You can then wrap the `handler` in `createLogFrom` in devContainer.ts with `replaceAllLog`.

chrmarti (changes requested):

src/spec-common/secrets.ts
eljog: What does this escaping actually do?

chrmarti: The RexExp will interpret its input string as a regular expression, the escaping takes care of the regex special characters by prepending them with a backslash.

src/spec-node/devContainersSpecCLI.ts
chrmarti: Maybe (untested):
```ts
export function replaceAllLog(origin: LogHandler, values: string[], replacement: string): LogHandler {
	const r = new RegExp(values.map(v => v.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')).join('|'), 'g');
	return {
		event: e => {
			if ('text' in e) {
				origin.event({
					...e,
					text: e.text.replace(r, replacement),
				});
			} else {
				origin.event(e);
			}
		}
	};
}
```

`stepDetails` might need the same treatment.

src/spec-common/secrets.ts
eljog: That makes sense. I will address this suggestion in a new commit int he morning

src/spec-node/devContainersSpecCLI.ts
eljog: I just committed a version of my attempt.. (before your last comment). Would you mind taking a look?

src/spec-node/devContainers.ts
```diff
@@ -199,24 +199,40 @@ export interface LogOptions {
...
 		},
 	};
 	return log;
 }
+
+async function maskSecrets(secretsP: Promise<Record<string, string>>, text: string) {
+	const MASK = '********';
+	const secretValues = Object.values(await secretsP);
+	secretValues.forEach(secret => {
+		const regex = new RegExp(secret, 'g');
```
chrmarti, line 234: You could create the RexExp once and then reuse it to avoid parsing by `new RegExp()` for each log event. (Left a comment with an example of using a single RegEx, so you only have to keep around one.)

src/spec-utils/log.ts
```diff
@@ -111,6 +111,10 @@ export function createCombinedLog(logs: LogHandler[], header?: string): LogHandl
 					text: header!,
 				}));
 			}
+
+			if (e.type !== 'progress') {
```
chrmarti, line 115: Does `stepDetail` for `'progress'` events need the same treatment? Not sure anymore what could be in there. /cc @samruddhikhandale

chrmarti (changes requested):

src/spec-utils/log.ts
```diff
@@ -98,10 +98,10 @@ export const nullLog: Log = {
 
 export const terminalEscapeSequences = /(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]/g; // https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python/33925425#33925425
 
-export function createCombinedLog(logs: LogHandler[], header?: string): LogHandler {
+export function createCombinedLog(logs: LogHandler[], maskSecrets: (text: string) => Promise<string>, header?: string): LogHandler {
 	let sendHeader = !!header;
 	return {
-		event: e => {
+		event: async e => {
```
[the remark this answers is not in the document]
eljog: Done

src/spec-node/devContainersSpecCLI.ts
eljog: Implemented the suggestion. Thank you

src/spec-node/devContainers.ts
```diff
@@ -199,24 +199,34 @@ export interface LogOptions {
...
 			// TODO: wait for write() to finish.
 		},
 	};
 	return log;
 }
+
+function maskSecrets(handler: LogHandler, secrets?: Record<string, string>): LogHandler {
+	if (secrets) {
+		const mask = '********';
+		const secretValues = secrets ? Object.values(secrets) : [];
```
chrmarti, line 227: nit: `secrets` is already checked for not being undefined above.

chrmarti (changes requested):
One additional comment and I'm not sure if stepDetails could contain secrets. Looks good otherwise!

src/spec-node/devContainers.ts
eljog: Done

src/spec-utils/log.ts
eljog: Added masking for stepDetail as well, since we also write progress events into json logs

chrmarti (approved):

PR #553, opened 2023-06-14
Features Test Cmd: "Duplicate" test mode to test Feature Idempotence

joshspicer:
ref: https://github.com/github/codespaces/issues/14081

This change adds a new mode that will be automatically executed if a `duplicate.sh` test script is added to a Feature's test directory.  This change will generate a dev container where a given Feature is installed twice in a row.   This is to allow Feature authors to make assertions about what should happen if the exact Feature (with different options) is executed more than once.

The first Feature installed will have random options based on the information in the `devcontainer-feature.json`.  The second Feature will be installed with the default options.  The test script `duplicate.sh` will have [both the random options and the default options available as environment variables](https://github.com/devcontainers/cli/pull/553/files#diff-71d3aadcfa6f5dda4c15d6367a46b1faa106d4009262dff4b5ccb50a0d057667R10-R20) to help write an accurate test.

The intention here is is that authors can write tests asserting (1) every Feature that opts-in to this test by adding a `duplicate.sh` will always succeed when it is installed >1 times and (2) there is predictable behavior (what Feature is on the path?) no matter what combination of options is provided


src/spec-node/featuresCLI/testCommandImpl.ts
```diff
@@ -103,9 +105,105 @@ async function runGlobalFeatureTests(args: FeaturesTestCommandInput, testResults
...
+	const options = featureMetadata.options || {};
+
+	// For each possible option, generate a random value for each Feature
+	const randomizedOptions: { [key: string]: string | boolean } = {};
+	Object.entries(options).forEach(([key, value]) => {
+		if (value.type === 'boolean') {
+			randomizedOptions[key] = !value.default;
+		}
+		if (value.type === 'string' && 'proposals' in value && value?.proposals?.length) {
+			const randomIndex = Math.floor(Math.random() * value.proposals.length);
```
chrmarti, line 132: Using randomized options might result in unexpected failures. Not sure if these will been seen as helpful. Maybe add command line option to enable/disable randomization and consider having it disabled by default to avoid unhelpful failures.

chrmarti (changes requested):
Looks good, left one comment.

src/spec-node/featuresCLI/testCommandImpl.ts
joshspicer: Yea I agree. My goal is making it easy to test installing two different "versions" of a given Feature.  The alternatives are:

- Two of the exact same Feature with same options (less interesting)
- One Feature with default value and one with a non-default but static set of options.

joshspicer: I've added a flag that can be passed to add in randomness, otherwise it will be non-default, but predictable values (unless the devcontainer-feature.json is updated)

chrmarti (dismissed):

marketoptiontend (approved):

chrmarti (approved):

PR #565, opened 2023-06-26
Outdated command

chrmarti:
Similar to yarn's and NPM's outdated command. https://github.com/microsoft/vscode-remote-release/issues/8640


src/spec-node/devContainersSpecCLI.ts
```diff
@@ -1042,6 +1043,87 @@ async function readFeaturesConfig(params: DockerCLIParameters, pkg: PackageConfi
...
+}
+
+type OutdatedArgs = UnpackArgv<ReturnType<typeof outdatedOptions>>;
+
+function outdatedHandler(args: OutdatedArgs) {
+	(async () => outdated(args))().catch(console.error);
+}
+
+async function outdated({
+	// 'user-data-folder': persistedFolder,
```
samruddhikhandale, line 1065: Curious, why have we added `user-data-folder` as an option, but have this bit commented out? 🤔

samruddhikhandale (approved):
left one question, besides that looks good to me!

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -1042,6 +1043,87 @@ async function readFeaturesConfig(params: DockerCLIParameters, pkg: PackageConfi
...
+		output = createLog({
+			logLevel: mapLogLevel(logLevel),
+			logFormat,
+			log: text => process.stderr.write(text),
+			terminalDimensions: terminalColumns && terminalRows ? { columns: terminalColumns, rows: terminalRows } : undefined,
+		}, pkg, sessionStart, disposables);
+
+		const workspace = workspaceFromPath(cliHost.path, workspaceFolder);
+		const configPath = configFile ? configFile : await getDevContainerConfigPathIn(cliHost, workspace.configFolderPath);
+		const configs = configPath && await readDevContainerConfigFile(cliHost, workspace, configPath, /* mountWorkspaceGitRoot */ true, output) || undefined;
```
joshspicer, line 1094:
```suggestion
		const configs = configPath && await readDevContainerConfigFile(cliHost, workspace, configPath, true, output) || undefined;
```

joshspicer (commented):
A couple questions:

1. Features in my `devcontainer.json` without a `version` are omitted.  Should we assume `latest` in this case and resolve the values accordingly?

image
vs.
image

2. Features that exist but that you don't have local access to don't fail the entire command.  This is probably ok but thought i'd highlight.  I feel the most logical thing to do here is mirror what npm/yarn does?

image

3. Features with a sha hash as the version are also ignored.  
image

src/spec-node/devContainersSpecCLI.ts
chrmarti: We have that for all other commands and it might make sense in the future, e.g., when we add a config file for the CLI.

chrmarti:
> A couple questions:
> 
> 1. Features in my `devcontainer.json` without a `version` are omitted.  Should we assume `latest` in this case and resolve the values accordingly?

Fixed.

> 2. Features that exist but that you don't have local access to don't fail the entire command.  This is probably ok but thought i'd highlight.  I feel the most logical thing to do here is mirror what npm/yarn does?

I agree. Not sure actually what npm/yarn do (seems my setup doesn't work with GitHub package repo atm), but having some resilience towards missing credentials makes sense.

> 3. Features with a sha hash as the version are also ignored.

Fixed.

Also adding an `--output-format` option with `text` being the default (following along the `features info` command).

joshspicer (approved):

PR #570, opened 2023-06-29
Add `--cache-to` option to `devcontainer build` command

### LINKED ISSUE #563: Export buildx remote cache when using devcontainer build command
## Problem
Currently, we are using the `--cache-to` option of the docker buildx build command to export the build cache to a registry or local file system. However, the devcontainer build command only has the --cache-from option and not the --cache-to option, which makes exporting the build cache impossible.

## Proposal
We would like to request the addition of the `--cache-to` option to the devcontainer build command, which would allow us to export the build cache. This would enable developers to export and reuse the build cache.

yokonao:
Close #563

I added `--cache-to` option to `devcontainer build` command.

If buildx is used for building image, we can pass `--cache-to` option in order to export and reuse buildx cache.

## Checks
- [ ] Create docker-container driver builx builder
  - `docker buildx create  --driver=docker-container --use `
- [ ] Build with `--cache-to` option
  - `node devcontainer.js build --workspace-folder . --cache-to=type=local,dest=/tmp/build-cache`
- [ ] Confirm cache directory exists
  - `ls /tmp/build-cache`

```.sh
❯ ls /tmp/build-cache
blobs  index.json  ingest  oci-layout
```


yokonao:

src/spec-node/singleContainer.ts
```diff
@@ -193,6 +193,9 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 				args.push('--load'); // (short for --output=docker, i.e. load into normal 'docker images' collection)
 			}
 		}
+		if (buildParams.buildxCacheTo) {
+			args.push('--cache-to', buildParams.buildxCacheTo);
```
joshspicer, line 197: Do you think it would make sense to include this passing this flag in the `src/spec-node/docker-compose.ts` case too?

Somewhere around here perhaps? https://github.com/devcontainers/cli/blob/main/src/spec-node/dockerCompose.ts#L235

yokonao: Thank you for comment!

- Devcontainer uses `docker compose build` for building image from `compose.yaml`. However, `docker compose build` does not have `--cache-to` option.
- Instead, it may make sense to include `cache-to` section in overriden docker compose files which devcontainer command renders.
- In my recoginition, devcontainer does not support `--push` option when building from docker compose config file.
- `--cache-to` option is usually used with `--push` option in CI. So, it is a little wired that `--cache-to` is supported, but `--push` is not.

We don't need to add `--cache-to` for Docker Compose cases  as my conclusion.
Objections welcome! Please let me know if there are any misunderstanding.

joshspicer: Thanks for the detailed response, makes sense to me now.

joshspicer (changes requested):
I've run CI on your changes and looks like there's a couple minor tweaks needed.  Linter issues can be detected locally with `yarn lint`. There's also a test issue but that probably just needs some minor reworking

yokonao:
Sorry! I fixed lint issue. Waiting for checks.

src/spec-node/devContainers.ts
```diff
@@ -52,6 +52,7 @@ export interface ProvisionOptions {
 	buildxPlatform: string | undefined;
 	buildxPush: boolean;
```
marketoptiontend, line 53: Ultra-Tight Spreads & Fast Execution
An account type for every trader – Standard, Pro-ECN, ECN & Platinum!

marketoptiontend (approved):

yokonao:
@joshspicer 
Thank you for checking on CI!
What should I do next before merging?

joshspicer (approved):
Thanks! I'll ask @chrmarti to take a quick peek to

yokonao:
@chrmarti 
Hi! Could you review this PR?

src/spec-node/singleContainer.ts
chrmarti, line 197: This also works without BuildKit, doesn't it? If so, you can move it out of the if/else.

chrmarti (changes requested):
Thanks for the PR! Left a comment. Additionally:

- Please add a check and error at https://github.com/devcontainers/cli/blob/eafbf6414d02703f967fcdf3be8c3b1fca647101/src/spec-node/devContainersSpecCLI.ts#L620 since it is not supported.

- Should the flag also be added at https://github.com/devcontainers/cli/blob/eafbf6414d02703f967fcdf3be8c3b1fca647101/src/spec-node/containerFeatures.ts#L90? (This runs when the user has an image without a Dockerfile, but we still want to add features to the image.)

src/spec-node/singleContainer.ts
yokonao: `--cache-to` is an option for `buildx build` command. I think it does not work without BuildKit.

e.g.

```.sh
❯ docker build --cache-to some-cache .
unknown flag: --cache-to
See 'docker build --help'.
```

yokonao:
Thanks for your review. I pushed two commits!

- check and raise with docker compose file and `--cache-to`
- append `--cache-to` also with an image without a Dockerfile

chrmarti (approved):
Approving PR, thanks!

@joshspicer / @samruddhikhandale Could you take another look? Thanks.

joshspicer (approved):

yokonao:
@samruddhikhandale 
Hi!, Take a look at this PR, please.

samruddhikhandale (approved):

mering:
This seems to be missing from `devcontainer build --help` documentation.

PR #572, opened 2023-06-29
Update supported node engines to ^16.13.0 || >=18.0.0

eljog:
With the upgrade to ES 2021 the CLI no longer support older node versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 	},
 	"license": "SEE LICENSE IN LICENSE.txt",
 	"engines": {
-		"node": ">=14.17.0"
+		"node": "^16.13.0 || >=18.0.0"
 	},
 	"scripts": {
 		"compile": "npm-run-all clean-dist definitions compile-dev",


package.json
```diff
@@ -15,7 +15,7 @@
 	},
 	"license": "SEE LICENSE IN LICENSE.txt",
 	"engines": {
-		"node": ">=14.17.0"
+		"node": "^16.13.0 || >=18.0.0"
```
joshspicer, line 18: For my knowledge, when is this value read/enforced?

joshspicer (approved):

package.json
jkeech: It's more of a declaration of which Node versions are supported. Not strictly enforced at runtime. https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines.

Since we are dropping support for Node 14 though, it's a breaking change and we need to call it out in the changelog and a minor version bump (would be a major version bump, but since we are on 0.x, minor is treated as major and can include breaking changes).

eljog: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines

As mentioned in this doc it is advisory only, and will only produce warnings since we don't specify `engine-strict`

joshspicer: gotcha, thanks

PR #576, opened 2023-07-03
Case-insensitive instructions

chrmarti:
Fixes microsoft/vscode-remote-release#6850


samruddhikhandale (approved):
⚡

joshspicer (approved):
Interesting, I wouldn't have expected dockerfile keywords to be case-insensitive, but it is so stated here https://docs.docker.com/engine/reference/builder/#format

PR #585, opened 2023-07-04
Publish without node-pty dependency

chrmarti:
https://github.com/microsoft/vscode-remote-release/issues/8698


joshspicer (approved):
LGTM - I think this will require some extra testing in Codespaces once published. Do you anticipate any other changes we'll need to make to pick this up?

chrmarti:
@joshspicer From what I saw you install `node-pty` separately before you install the CLI and I think that might make everything continue to work as before. 🤞

PR #586, opened 2023-07-05
Fix #543: build with correct image names

### LINKED ISSUE #543: "--image-name" is ignored when devcontainer.json is using "image" statement
Hi there,

i was following this guide to setup our internal environment accordingly:
https://containers.dev/guide/gitlab-ci

This works correctly since the devcontainer.json contains the following part:
```
  "build": {
    "context": "..",
    "dockerfile": "Dockerfile"
  },
```

But as soon as i switch this to just referencing an image name instead of an dockerfile in the devcontainer.json, e.g. via:
```
"image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
```

The referenced image name via:
```
devcontainer build --workspace-folder . --push true --image-name ${CI_REGISTRY_IMAGE}:latest
```

is completly ignored and just some random image name containing a hashes is generated. This block the usage of gitlab-ci, since the resulting image is not pushable to the gitlab registry, since the naming of the docker image does not correspond anymore correctly to the project name. Nevertheless a workaround is at the moment to create a dummy Dockerfile just containing the FROM statement referencing the image name.

But it would be beneficial if the `--image-name` works for both scenarios ;)

best,
Christian

invasy:
Closes https://github.com/devcontainers/cli/issues/543

# Problem
- Method `extendImage` created image with the default name ignoring `--image-name` option.
- The image could not be tagged after building, because CLI exited with `ERROR: denied: requested access to the resource is denied` after pushing attempt.

# Solution
- Added image names from `--image-names` option at building stage.

### DIFF
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -30,7 +30,7 @@ export const getSafeId = (str: string) => str
 	.replace(/^[\d_]+/g, '_')
 	.toUpperCase();
 
-export async function extendImage(params: DockerResolverParameters, config: SubstitutedConfig<DevContainerConfig>, imageName: string, additionalFeatures: Record<string, string | boolean | Record<string, string | boolean>>, canAddLabelsToContainer: boolean) {
+export async function extendImage(params: DockerResolverParameters, config: SubstitutedConfig<DevContainerConfig>, imageName: string, imageNames: string[] | undefined, additionalFeatures: Record<string, string | boolean | Record<string, string | boolean>>, canAddLabelsToContainer: boolean) {
 	const { common } = params;
 	const { cliHost, output } = common;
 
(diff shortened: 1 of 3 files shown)


invasy:

src/spec-node/containerFeatures.ts
```diff
@@ -90,14 +88,18 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
 	for (const buildArg in featureBuildInfo.buildArgs) {
 		args.push('--build-arg', `${buildArg}=${featureBuildInfo.buildArgs[buildArg]}`);
 	}
+
+	const folderImageName = getFolderImageName(common);
+	const updatedImageName: string[] = imageNames ?? [`${imageName.startsWith(folderImageName) ? imageName : folderImageName}-features`];
```
joshspicer, line 93: Looking quickly at [CI](https://github.com/devcontainers/cli/actions/runs/5469255000/job/15377989792?pr=586#step:7:822), I think `imageName` might need a null check here

Alirakz181016 (approved):

samruddhikhandale (commented):
Hi @invasy 👋 

Apologies on missing out on this PR, would you be still interested in checking this PR?

chrmarti:
Addressed by https://github.com/devcontainers/cli/issues/791.

PR #594, opened 2023-07-14
Record features referenced by tarball URI

chrmarti:
https://github.com/devcontainers/spec/issues/236


samruddhikhandale (approved):
Looks great to me!

PR #595, opened 2023-07-14
Set npm license field to `MIT`.

### LINKED ISSUE #542: Use of license identifier in package.json metadata.
The package.json [license field](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#license) says `SEE LICENSE IN LICENSE.txt`. The use of this sentence usually indicates that there is not a standard [SPDX identifier](https://spdx.org/licenses/) for the license text.

Instead, the value should like be the SPDX value of `MIT`. The LICENSE.txt file is identical to [oci](https://opensource.org/license/mit/)'s listing of the MIT license.

CodeMan99:
Fixes #542

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
 	"bugs": {
 		"url": "https://github.com/devcontainers/cli/issues"
 	},
-	"license": "SEE LICENSE IN LICENSE.txt",
+	"license": "MIT",
 	"engines": {
 		"node": "^16.13.0 || >=18.0.0"
 	},


package.json
```diff
@@ -13,7 +13,7 @@
 	"bugs": {
 		"url": "https://github.com/devcontainers/cli/issues"
 	},
-	"license": "SEE LICENSE IN LICENSE.txt",
+	"license": "MIT",
```
samruddhikhandale, line 16: // cc @bamurtaugh

bamurtaugh (approved):
LGTM, thank you!

PR #618, opened 2023-08-21
RFC: Build as single executable application

ZauberNerd:
This commit adds a build script to package the `devcontainer` CLI as a single executable application using the new (and experimental) Node.js feature of the same name.
https://nodejs.org/api/single-executable-applications.html

Distributing the devcontainer CLI as a single executable application allows to use it on systems where Node.js is not installed. For example on immutable systems like Fedora Silverblue or openSUSE MicroOS.

I'm aware of another PR that also tries to package standalone binaries via `@vercel/pkg` here: https://github.com/devcontainers/cli/pull/343 but it seems that PR hasn't received updates in a while and I'm not sure if it is stale or not. Feel free to close my PR if the other approach is preferable.

TODOs:
- [ ] Code signing on Windows/MacOS (see point 5 https://nodejs.org/api/single-executable-applications.html)
- [ ] Update CLI tests to test against the executable
- [ ] Update GitHub workflows to build, test and publish the executable

Let me know what you think about either this PR or #343 and whether I should continue with the above outlined TODOs.

### DIFF
--- a/build/single-executable.js
+++ b/build/single-executable.js
@@ -1,0 +1,49 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+"use strict";
+
+const fs = require("fs");
+const cp = require("child_process");
+
+/**
+ * Create a single executable file for the devcontainer CLI.
+ * https://nodejs.org/api/single-executable-applications.html
+ */
+
+process.chdir("dist/spec-node");
+
+const name = "devcontainer";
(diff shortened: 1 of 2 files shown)


ZauberNerd:

ZauberNerd:
> Let me know what you think about either this PR or https://github.com/devcontainers/cli/pull/343 and whether I should continue with the above outlined TODOs.

cc @trxcllnt @jkeech because you were both involved in the above mentioned PR

ZauberNerd:
> One thing I'm not sure about is the `node-pty` native dependency and how that would work in this packaging format. I'm guessing that it wouldn't work, but it would be good to double check.

Correct, native modules don't work with this packaging format. But https://github.com/devcontainers/cli/pull/585 causes the `node-pty` to be optional, which makes this PR feasible.

JustDontsnitchonme (approved):

PR #631, opened 2023-09-02
updateUID: remove redundant `SHELL` instruction

vilgotf:
Docker already defaults to `/bin/sh -c` on Linux... This option is not part of the OCI spec and produces a warning by default when building on podman.

### DIFF
--- a/scripts/updateUID.Dockerfile
+++ b/scripts/updateUID.Dockerfile
@@ -8,7 +8,6 @@ USER root
 ARG REMOTE_USER
 ARG NEW_UID
 ARG NEW_GID
-SHELL ["/bin/sh", "-c"]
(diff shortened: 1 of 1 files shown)


chrmarti:
We added it there to make sure we use `/bin/sh`. The user might changed it to some other shell using `SHELL`.

vilgotf:
> We added it there to make sure we use `/bin/sh`. The user might changed it to some other shell using `SHELL`.

First of all, I'm not even sure that the `SHELL` instruction propogates across dockerfiles. Secondly, even if that's the case, silently overriding the user's `SHELL` instruction in a downstream image is HORRIBLE. It would be much better to do one of the following:

1. restore the `SHELL` just before after exiting from this image (however, the `SHELL` override is only used for one `RUN`, so...)
2. replace `RUN [...]` with `CMD ["sh", "-c", "[...]"]` such that the user's `SHELL` override is bypassed
3. error if the upstream's `SHELL` doesn't support `eval` (overriding `SHELL` before building dependencies is code smell, so those affected by this should update their images either way)

I can update this PR with whatever option you'd think is best.

PR #638, opened 2023-09-13
Fix: propagate --cache-from to buildx build

dhhyi:
I am trying to use the new `--cache-to` feature from #570 with a [registry cache](https://docs.docker.com/build/cache/backends/registry/).

If I understand it correctly this only works when using both `--cache-to` and `--cache-from`, but the devcontainers-cli does not send `--cache-from` to `docker buildx build`.

### DIFF
--- a/src/spec-node/containerFeatures.ts
+++ b/src/spec-node/containerFeatures.ts
@@ -85,6 +85,9 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
 		if (params.buildxCacheTo) {
 			args.push('--cache-to', params.buildxCacheTo);
 		}
+		if (!params.buildNoCache) {
+			params.additionalCacheFroms.forEach(cacheFrom => args.push('--cache-from', cacheFrom));
+		}
 
 		for (const buildContext in featureBuildInfo.buildKitContexts) {
 			args.push('--build-context', `${buildContext}=${featureBuildInfo.buildKitContexts[buildContext]}`);


src/spec-node/singleContainer.ts
```diff
@@ -202,6 +202,9 @@ async function buildAndExtendImage(buildParams: DockerResolverParameters, config
 		if (buildParams.buildxCacheTo) {
 			args.push('--cache-to', buildParams.buildxCacheTo);
 		}
+		if (buildParams.additionalCacheFroms) {
+			buildParams.additionalCacheFroms.forEach(cacheFrom => args.push('--cache-from', cacheFrom));
```
chrmarti, line 206: Line 231 below already does this. 🤔

src/spec-node/containerFeatures.ts
```diff
@@ -80,6 +80,9 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
 		if (params.buildxCacheTo) {
 			args.push('--cache-to', params.buildxCacheTo);
 		}
+		if (params.additionalCacheFroms) {
```
chrmarti, line 83: Skip if `params.buildNoCache` is set.

chrmarti (changes requested):

src/spec-node/containerFeatures.ts
chrmarti: Just noticed `params.additionalCacheFroms` is always set, no need to check for that.

src/spec-node/singleContainer.ts
dhhyi: hmm, I don't remember why I added it, sorry. It also seems to be working without. 🤷

src/spec-node/containerFeatures.ts
dhhyi: ✅

chrmarti (approved):
Thanks!

samruddhikhandale (approved):
Looks good, thanks!

Issue #659, opened 2023-10-19
Suggestion about findComposeContainer


NerdToMars:
It's a bit inadequate to search docker container only with project name and services when project name is the basename of the working directory.
 
https://github.com/devcontainers/cli/blob/71180e791b13371f8313d981dfcf06b3c8fcd2ab/src/spec-node/dockerCompose.ts#L613

Can I suggest to either use  longer name for project:
https://github.com/devcontainers/cli/blob/71180e791b13371f8313d981dfcf06b3c8fcd2ab/src/spec-node/dockerCompose.ts#L643-L645

like replace the use of basename with a representation of the full path where each subdirectory is connected with an underscore (_)
```
function pathToUnderscore(p: string): string {
    return p.split(path.sep).join('_');
}

...
...

if (equalPaths(cliHost.platform, workingDir, cliHost.path.join(configDir, '.devcontainer'))) {
    return toProjectName(`${pathToUnderscore(configDir)}_devcontainer`, newProjectName);
}
return toProjectName(pathToUnderscore(workingDir), newProjectName);
```

samruddhikhandale:
Hi 👋 

Thanks for the suggestion, it makes sense to me.

@NerdToMars would you be interested in contributing the change to the CLI? We appreciate support from the community, thanks!

NerdToMars:
or is it possible to add more filter in `findComposeContainer` function

like filter based on devcontainer.local_folder

chrmarti:
This is how Docker Compose identifies containers. We can't change that.

PR #661, opened 2023-10-21
use whole dir as project base name

NerdToMars:
To address https://github.com/devcontainers/cli/issues/659


samruddhikhandale (commented):
Thanks, can you help add tests to validate this change?

chrmarti:
If someone had existing containers using only the basename, wouldn't this change abandon these and create new ones with the new project name?

Csaba1205 (changes requested):
1

PR #670, opened 2023-10-31
fix version sorting and report major version in `outdated`

joshspicer:
Ref: https://github.com/devcontainers/cli/issues/676

Reported in https://github.com/devcontainers/spec/discussions/327

Previously the comparision done to sort versions would incorrectly place patch versions >=10 above versions x.x.2-x.x.9.  This is due to the versions seemingly being sorted as string.

Eg:
image
Note that `1.0.11` is not the final entry.


samruddhikhandale (approved):
Looks great! ⚡

chrmarti (approved):
Great, thank you!

PR #681, opened 2023-11-14
Force deterministic order for 'outdated' command

joshspicer:
Currently the position of each Feature in the `outdated` command is non-deterministic due to the `Promise.all`.  This isn't too big a deal for a user, but it's more intuitive to have the output be consistent between runs.

It's also a helpful assumption that the positions will match (ie: writing units tests, which is how I noticed this).

This change mirror the position of each Feature in the config (the input).


chrmarti (approved):
Makes sense, thanks!

PR #684, opened 2023-11-14
Edit a Feature pinned version via upgrade cmd

joshspicer:
This change adds new arguments to the `upgrade` command to facilitate updating a single Feature in a `devcontainer.json`.  After the edit to the `devcontainer.json` is made, the lockfile is updated according to the requirements (version tags) of each Feature. 

https://github.com/devcontainers/cli/assets/23246594/68073ae6-3590-4c6a-9193-edeed5dd7883

This is useful in dependbot's [`file_updater`](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/file_updaters/README.md) step, allowing us to do shell out to the CLI directly for this operation. This is similar to the [Swift's `file_updater` implementation in dependabot core](https://github.com/joshspicer/dependabot-core/blob/joshspicer/devcontainers/swift/lib/dependabot/swift/file_updater/lockfile_updater.rb#L37)

I've marked this arguments as 'hidden' while I do some initial experimentation. Today the entire lockfile will be regenerated even if `--feature` is used.

ref: https://github.com/github/codespaces/issues/15791

### DIFF
--- a/src/spec-configuration/lockfile.ts
+++ b/src/spec-configuration/lockfile.ts
@@ -13,20 +13,8 @@ export interface Lockfile {
 	features: Record<string, { version: string; resolved: string; integrity: string }>;
 }
 
-export async function writeLockfile(params: ContainerFeatureInternalParams, config: DevContainerConfig, featuresConfig: FeaturesConfig, forceInitLockfile?: boolean, dryRun?: boolean): Promise<string | undefined> {
-	const lockfilePath = getLockfilePath(config);
-	const oldLockfileContent = await readLocalFile(lockfilePath)
-		.catch(err => {
-			if (err?.code !== 'ENOENT') {
-				throw err;
-			}
-		});
-
-	if (!forceInitLockfile && !oldLockfileContent && !params.experimentalLockfile && !params.experimentalFrozenLockfile) {
-		return;
-	}
-
-	const lockfile: Lockfile = featuresConfig.featureSets
+export async function generateLockfile(featuresConfig: FeaturesConfig): Promise<Lockfile> {
+	return featuresConfig.featureSets
 		.map(f => [f, f.sourceInformation] as const)
 		.filter((tup): tup is [FeatureSet, OCISourceInformation | DirectTarballSourceInformation] => ['oci', 'direct-tarball'].indexOf(tup[1].type) !== -1)
 		.map(([set, source]) => {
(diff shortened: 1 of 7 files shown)


src/spec-configuration/lockfile.ts
```diff
@@ -13,20 +13,8 @@ export interface Lockfile {
 	features: Record<string, { version: string; resolved: string; integrity: string }>;
 }
 
-export async function writeLockfile(params: ContainerFeatureInternalParams, config: DevContainerConfig, featuresConfig: FeaturesConfig, forceInitLockfile?: boolean, dryRun?: boolean): Promise<string | undefined> {
```
joshspicer, line 16: Strange diff here.  What I did was extract out `generateLockfile` from what was `writeLockfile`.  That eliminates the need to pass in the `dryRun` bool that I added a few commits ago.

src/spec-node/upgradeCommand.ts
```diff
@@ -123,4 +146,72 @@ async function featuresUpgrade({
...
+
+	if (!configPath || !(await isLocalFile(configPath))) {
+		throw new ContainerError({ description: `Error running upgrade command.  Config path '${configPath}' does not exist.` });
+	}
+
+	const configText = await readLocalFile(configPath);
+	const previousConfigText: string = configText.toString();
+	let updatedText: string = configText.toString();
+
+	// Clear Features object to make editing easier
```
chrmarti, line 168: Wouldn't it be easier to only edit the affected property inside "features"? (Also: jsonc.parse drops all formatting and comments, so these would be lost from all features' options.)

chrmarti (changes requested):

src/spec-node/upgradeCommand.ts
joshspicer: Probably :) Will give it a try

joshspicer: Made a change, not sure if it's much better though.  Do you think I should do this a different way (eg: without jsonc)?

joshspicer: Changed it to a simple regex in https://github.com/devcontainers/cli/pull/684/commits/c13b2c228899026829c9e6ca7384f9d9db856d15

chrmarti: LGTM! I realize I took a similar approach for the 'code lenses' in the Dev Containers extension (doing string replacements).

chrmarti (approved):

src/spec-node/upgradeCommand.ts
joshspicer: I've done the same thing in the past for the `template apply` function

Issue #691, opened 2023-11-22
"devcontainer features publish" fails to log in with Artifactory


imphil:
We're using Artifactory as container registry and I'm trying to use it to store a Dev Container Feature. Unfortunately, the login fails.

I have setup a Docker credential helper to return the username and token, but we're not even getting there:

```
$ devcontainer features publish --registry docker-eu.artifactory.example.com -n my-repo/my-project devcontainer-features/src/git-absorb --log-leve
l=trace
[3 ms] @devcontainers/cli 0.54.0. Node.js v20.9.0. linux 4.18.0-477.27.1.el8_8.x86_64 x64.
[3 ms] Packaging single feature...
[27 ms] Packaged feature 'git-absorb'
[34 ms] Processing feature: git-absorb...
[34 ms] > input: docker-eu.artifactory.example.com/my-repo/my-project/git-absorb
[34 ms] >
[34 ms] > resource: docker-eu.artifactory.example.com/my-repo/my-project/git-absorb
[34 ms] > id: git-absorb
[34 ms] > owner: my-repo
[34 ms] > namespace: my-repo/my-project
[34 ms] > registry: docker-eu.artifactory.example.com
[34 ms] > path: my-repo/my-project/git-absorb
[34 ms] >
[34 ms] > version: latest
[34 ms] > tag?: latest
[34 ms] > digest?: undefined
[35 ms] Feature Annotations: {"dev.containers.metadata":"{\"id\":\"git-absorb\",\"name\":\"Install git-absorb\",\"version\":\"1.0.0\",\"description\":\"Install git absorb (https://github.com/tummychow/git-absorb)\",\"installsAfter\":[\"ghcr.io/devcontainers/features/common-utils\"]}"}
[35 ms] Fetching published versions...
[251 ms] [httpOci] Attempting to authenticate via 'Bearer' auth.
[252 ms] [httpOci] WWW-Authenticate header is not in expected format. Got:  Bearer realm="https://docker-eu.artifactory.example.com/artifactory/api/docker/null/v2/token",service="docker-eu.artifactory.example.com"
[252 ms] Request failed
[252 ms] (!) ERR: Failed to publish 'docker-eu.artifactory.example.com/my-repo/my-project/git-absorb'
```

The `scope` key is not part of the response, which seems to be allowed according to https://datatracker.ietf.org/doc/html/rfc6750#section-3 (as referenced in https://distribution.github.io/distribution/spec/auth/token/).

[...]

PR #692, opened 2023-11-22
Fix authentication against Artifactory

imphil:
To authenticate against an OCI registry the code currently requires a
`scope` attribute in the `WWW-Authenticate` header. This key is optional
according to RFC 6750 [1], an empty string can be used if it's not
present.

Artifactory does not include the `scope` attribute and hence
authentication fails currently. Fix that by making the attribute
optional.

[1] "Use of the "scope" attribute is OPTIONAL."

Fixes https://github.com/devcontainers/cli/issues/691


src/spec-configuration/httpOCIRegistry.ts
```diff
@@ -95,15 +95,15 @@ export async function requestEnsureAuthenticated(params: CommonParams, httpOptio
...
+			if (!realmGroup || !serviceGroup) {
 				output.write(`[httpOci] WWW-Authenticate header is not in expected format. Got:  ${wwwAuthenticate}`, LogLevel.Trace);
 				return;
 			}
 
 			const wwwAuthenticateData = {
 				realm: realmGroup[1],
 				service: serviceGroup[1],
-				scope: scopeGroup[1],
+				scope: scopeGroup ? scopeGroup[1] : "",
```
samruddhikhandale, line 106: If we allow an empty scope, it could mean that the client is requesting a token without any specific access rights. Depending on implementation, this could result in a token with default access rights being issued, or it could result in an error or a token with no access rights.

I definitely agree that empty `scope` is allowed in the `WWW-Authenticate header`, and it's great to see that it fixes auth issues for Artifactory. However, I wonder if this should also be reviewed from security perspective (in case, there's another alternative to fix the issue for Artifactory)

Looping in @joshspicer as he has the most context.

samruddhikhandale (commented):
Thank you for taking the time in contributing this change, appreciate it! ✨

src/spec-configuration/httpOCIRegistry.ts
joshspicer, line 106:
```suggestion
				scope: scopeGroup ? scopeGroup[1] : '',
```

joshspicer (commented):
Thanks for the change and pointer to the section [in the RFC](https://www.rfc-editor.org/rfc/rfc6750.html#section-3)!

One little change needed for the linter to pass, then it looks good to me 👍 

I'll try to look into setting up a test Artifactory to add as a [regression test here](https://github.com/devcontainers/cli/blob/main/src/test/container-features/registryCompatibilityOCI.test.ts#L40)

joshspicer (approved):

samruddhikhandale (approved):

imphil:
> I'll try to look into setting up a test Artifactory to add as a [regression test here](https://github.com/devcontainers/cli/blob/main/src/test/container-features/registryCompatibilityOCI.test.ts?rgh-link-date=2023-11-27T17%3A08%3A34Z#L40)

One more note: Please feel free to reach out to me directly if you cannot reproduce the behavior I'm seeing. There seem to be a number of ways to set up Artifactory auth, so maybe I can dig up the right details to help you reproduce the problem.

samruddhikhandale:
`v0.54.1` of the @devcontainers/cli now includes this PR code changes 🚀

PR #700, opened 2023-12-06
Escape * to prevent linux globbing from dumping random files into argument

limarkxx:
While attempting to "Reopen in Container" in VSCode (Running Ubuntu through WSL with Docker CLI, no Docker Desktop) on a devcontainer.json targeting a docker-compose.yml, similar to the following: 
```json
{
	"name": "xyzdef",
        "service": "xyzdef",
        "workspaceFolder": "/xyzdef",
        "dockerComposeFile": [ "../../docker-compose.yml" ] //docker-compose has a service named xyzdef
}
```
I got the error:
image
further up:
image
image
The `*` being placed in the docker compose cli call was getting globbed as a file wildcard, resulting in a random file from the workspace folder root being passed to the `--profile` parameter

Editing `~/.vscode-remote-containers/dist/dev-containers-cli-0.321.0/dist/spec-node/devContainersSpecCLI.js` with the following change fixed the problem.

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -567,7 +567,7 @@ export async function readDockerComposeConfig(params: DockerCLIParameters, compo
 		}
 		const composeCLI = await params.dockerComposeCLI();
 		if ((parseVersion(composeCLI.version) || [])[0] >= 2) {
-			composeGlobalArgs.push('--profile', '*');
+			composeGlobalArgs.push('--profile', '\\*');
 		}
 		try {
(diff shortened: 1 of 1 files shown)


chrmarti:
This command is run without a shell, so no glob expansion should be applied. Is `docker-compose` a shell script?

limarkxx:
Thanks, that's right

limarkxx:
This may still be of interest to you since adding the `\\` doesn't affect compatibility with the v1 docker-compose syntax and only adds support for cases where v2 docker compose syntax was shimmed in using a script

PR #707, opened 2023-12-13
Log `docker start` output

chrmarti:
Fixes microsoft/vscode-remote-release#5887

### DIFF
--- a/src/spec-node/singleContainer.ts
+++ b/src/spec-node/singleContainer.ts
@@ -305,7 +305,8 @@ async function startExistingContainer(params: DockerResolverParameters, labels: 
 	if (start) {
 		const starting = 'Starting container';
 		const start = common.output.start(starting);
-		await dockerCLI(params, 'start', container.Id);
+		const infoParams = { ...toExecParameters(params), output: makeLog(common.output, LogLevel.Info), print: 'continuous' as 'continuous' };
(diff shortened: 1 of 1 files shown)


src/spec-node/singleContainer.ts
```diff
@@ -305,7 +305,8 @@ async function startExistingContainer(params: DockerResolverParameters, labels:
 	if (start) {
 		const starting = 'Starting container';
 		const start = common.output.start(starting);
-		await dockerCLI(params, 'start', container.Id);
+		const infoParams = { ...toExecParameters(params), output: makeLog(common.output, LogLevel.Info), print: 'continuous' as 'continuous' };
```
eljog, line 308: Not sure if I fully follow this
Can this use the log that has secret masking support?
Like:
https://github.com/devcontainers/cli/blob/09e9821598a056e8d11dfd41e05ccbf370d57562/src/spec-node/devContainers.ts#L221

chrmarti: Yes, `common.output` has that built-in. Here `makeLog()` is only used to change the default log level of messages to Info, the new Log object passes all log messages on to `common.output` (which is passed as the first argument to makeLog()).

eljog (approved):

Issue #85, opened 2022-07-15
Support additional Docker build options


sanyer:
It would be a great and useful ability to add extra args to docker build command.
It seems there is already a variable `additionalBuildArgs` in the code for this at https://github.com/devcontainers/cli/blob/6d18acf/src/spec-node/singleContainer.ts#L143

The two main use cases for myself are adding `--progress=plan` and `--secret` [arguments](https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information).
Of course, I can do it myself by executing docker build outside the vscode, but it is not the same.

ontoneio:
I would like to second this. Running into an issue where I would like to provide runtime environment BUILD_ARGS. Dates and dynamic versioning type metadata that seems to be difficult to implement with the current API.

rastislav-vcrs:
+1 on this. My use-case is around network isolation. Due to network settings I need to use host network and not default container network while building. Requires `--network=host` to be supplied for build argument to reach out to internet.

rastislav-vcrs:
Any update on this?

ericfeunekes:
Adding my +1. I'm building a bunch of centrally managed devcontainers for my team and would like to dynamically create containers for different runtime versions that are passed to something similar to Docker build args. 

In an ideal world the devcontainer args could be used in any field so that I could, for example specify a different base image or base image tag based on incoming args. Not the end of the world if it's just a pass through to docker though since you could get the same effect by creating a Dockerfile and passing the args through. 

currently I have a single base image with my build args set, then I have to create a separate devcontainer.json for each runtime version that I want to support.

chrmarti:
Implementing proposal https://github.com/devcontainers/spec/pull/328 here.

chrmarti:
Available in the Dev Containers CLI 0.56.0.

PR #718, opened 2024-01-15
Support additional Docker build options

chrmarti:
#85


eljog (approved):
⚡

tonynajjar:
Thank you for this! When can we expect the Dev container extension to be able to leverage this feature? I can't seem to find which devcontainer/cli version the extension uses and if it's possible to force it to use a rece recent version (0.56.0). Thanks

PR #728, opened 2024-01-22
Add hidden `--omit-syntax-directive` flag

joshspicer:
Unblocks bug with `#syntax` directives + buildkit + user namespace remapping by not auto-injecting the `#syntax` directive in the extended (Features) intermediate Dockerfile, and ignoring user-contributed syntax directives.  

See https://github.com/moby/buildkit/issues/4556 for more info.

Enable with hidden `--omit-syntax-directive` `up` and `build` flag.  **This flag will be removed once upstream is fixed.**

ref: https://github.com/github/codespaces/issues/15748


chrmarti (approved):
Looks good!

Issue #712, opened 2023-12-21
`devcontainer outdated` gives incorrect current version information for some deprecated features


deivid-rodriguez:
With the following `devcontainer.json` file:

https://github.com/dependabot/dependabot-core/blob/d7a1fe636f86b8b7cfafe70ffaea6fbc8cbb126b/.devcontainer/devcontainer.json

Running `outdated --workspace-folder . --config devcontainer.json --output-format json`, I get the following result:

```
[2 ms] @devcontainers/cli 0.55.0. Node.js v20.8.0. darwin 23.2.0 arm64.
[13 ms] (!) WARNING: Using the deprecated 'docker-from-docker' Feature. See https://github.com/devcontainers/features/tree/main/src/docker-from-docker#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'github-cli' Feature. See https://github.com/devcontainers/features/tree/main/src/github-cli#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'node' Feature. See https://github.com/devcontainers/features/tree/main/src/node#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'golang' Feature. See https://github.com/devcontainers/features/tree/main/src/go#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'ruby' Feature. See https://github.com/devcontainers/features/tree/main/src/ruby#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'rust' Feature. See https://github.com/devcontainers/features/tree/main/src/rust#example-usage for the updated Feature.
[14 ms] (!) WARNING: Using the deprecated 'dotnet' Feature. See https://github.com/devcontainers/features/tree/main/src/dotnet#example-usage for the updated Feature.
{
  "features": {
    "docker-from-docker": {
      "current": "1.3.1",
      "wanted": "1.3.1",
      "wantedMajor": "1",
      "latest": "1.3.1",
      "latestMajor": "1"
    },
    "github-cli": {
      "current": "1.0.11",
      "wanted": "1.0.11",
      "wantedMajor": "1",
      "latest": "1.0.11",
      "latestMajor": "1"
    },
    "node": {
      "current": "1.3.1",
      "wanted": "1.3.1",
      "wantedMajor": "1",
      "latest": "1.3.1",
      "latestMajor": "1"
    },
    "golang": {
      "current": "1.2.2",
[...]

PR #731, opened 2024-01-24
Ignore non-versioned Features when generating `outdated` report

joshspicer:
Made a couple changes to the `outdated` command.

- Addresses https://github.com/devcontainers/cli/issues/712 by omitting the deprecated "short-hand" Feature syntax from the output of `outdated`.  
    - This is a bit opinionated, but my reasoning here is that - since there is no way for the user to alter these Feature versions (they're "held back" to the major version from when they were deprecated) - it doesn't make sense to surface in `outdated`.  
    - Omitting these Features from `outdated` will subsequently cause the [dependabot integration](https://github.com/dependabot/dependabot-core/blob/main/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb#L47) to not consider them for version updates, which [is already being done as an exceptional case](https://github.com/dependabot/dependabot-core/blob/main/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb#L70).
- Updates to `getRef()` to better filter out non-OCI Features, namely "local Features".
- Fix a sorting bug that caused `--output-format text` to exit unsuccessfully if a "local Feature" was present in the config.
    -  Added a test to exercise that code path

### DIFF
--- a/src/spec-configuration/containerCollectionsOCI.ts
+++ b/src/spec-configuration/containerCollectionsOCI.ts
@@ -142,6 +142,12 @@ export function getRef(output: Log, input: string): OCIRef | undefined {
 	// Normalize input by downcasing entire string
 	input = input.toLowerCase();
 
+	// Invalid if first character is a dot
+	if (input.startsWith('.')) {
+		output.write(`Input '${input}' failed validation.  Expected input to not start with '.'`, LogLevel.Error);
+		return;
+	}
+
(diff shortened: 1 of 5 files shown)


chrmarti (approved):
LGTM!

samruddhikhandale (approved):
⚡

samruddhikhandale (approved):

PR #750, opened 2024-02-12
Remove dependency on ip package

chrmarti:
- Removing dependency on `ip` package to address https://github.com/advisories/GHSA-78xj-cgh5-2h22.
- Fixing tests to make `docker-in-docker` work again.


samruddhikhandale (approved):
Looks good!

PR #768, opened 2024-03-01
Updates "outdated" command to display stale image versions

samruddhikhandale:


.github/workflows/dev-containers.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           "src/test/cli.test.ts",
           "src/test/cli.up.test.ts",
           "src/test/imageMetadata.test.ts",
+          "src/test/cli.outdated.test.ts",
           "src/test/container-features/containerFeaturesOCIPush.test.ts",
           # Run all except the above:
           "--exclude src/test/container-features/containerFeaturesOrder.test.ts --exclude src/test/container-features/registryCompatibilityOCI.test.ts --exclude src/test/container-features/containerFeaturesOCIPush.test.ts --exclude src/test/container-features/e2e.test.ts --exclude src/test/container-features/featuresCLICommands.test.ts --exclude src/test/cli.build.test.ts --exclude src/test/cli.exec.buildKit.1.test.ts --exclude src/test/cli.exec.buildKit.2.test.ts --exclude src/test/cli.exec.nonBuildKit.1.test.ts --exclude src/test/cli.exec.nonBuildKit.2.test.ts --exclude src/test/cli.test.ts --exclude src/test/cli.up.test.ts --exclude src/test/imageMetadata.test.ts 'src/test/**/*.test.ts'",
```
joshspicer, line 61: If you're isolating a test file here you'll want to also exclude it down below like:

```suggestion
          "--exclude src/test/container-features/containerFeaturesOrder.test.ts --exclude src/test/container-features/registryCompatibilityOCI.test.ts --exclude src/test/container-features/containerFeaturesOCIPush.test.ts --exclude src/test/container-features/e2e.test.ts --exclude src/test/container-features/featuresCLICommands.test.ts --exclude src/test/cli.build.test.ts --exclude src/test/cli.exec.buildKit.1.test.ts --exclude src/test/cli.exec.buildKit.2.test.ts --exclude src/test/cli.exec.nonBuildKit.1.test.ts --exclude src/test/cli.exec.nonBuildKit.2.test.ts --exclude src/test/cli.test.ts --exclude src/test/cli.up.test.ts --exclude src/test/imageMetadata.test.ts --exclude src/test/cli.outdated.test.ts 'src/test/**/*.test.ts'",
```

src/spec-node/dockerfileUtils.ts
```diff
@@ -119,6 +119,54 @@ export function findBaseImage(dockerfile: Dockerfile, buildArgs: Record<string,
...
+		}).reverse()
+		.reduce((str, { begin, end, value }) => str.substring(0, begin) + value + str.substring(end), str);
+}
+
+function findValueInImage(dockerfile: Dockerfile, buildArgs: Record<string, string>, variable: string, previousStage: Stage | undefined) {
+	if (buildArgs !== undefined && variable in buildArgs) {
+		return buildArgs[variable];
+	}
+
+	if (previousStage !== undefined) {
```
chrmarti, line 155: I'm not sure the previous stage should play a role here, but I might be missing something given the complexity of Dockerfile's build arg support.

I wonder if we could reuse more of the existing code to do what outdatedCommandImpl.ts needs. E.g., by adding a variation of the existing `findBaseImage`:
```ts
export function findBaseImages(dockerfile: Dockerfile, buildArgs: Record<string, string>) {
	const resolvedBaseImages = {} as Record<string, string>;
	for (const stage of dockerfile.stages) {
		const image = replaceVariables(dockerfile, buildArgs, /* not available in FROM instruction */ {}, stage.from.image, dockerfile.preamble, dockerfile.preamble.instructions.length);
		const nextStage = dockerfile.stagesByLabel[image];
		if (!nextStage) {
			resolvedBaseImages[stage.from.image] = image;
		}
	}
	return resolvedBaseImages;
}
```

outdatedCommandImpl.ts could just call this if I understand correctly.

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -0,0 +1,441 @@
...
+			output,
+			env: cliHost.env,
+			skipFeatureAutoMapping: false,
+			platform: cliHost.platform,
+		};
+
+		const outdatedFeatures = await loadFeatureVersionInfo(params, configs.config.config);
+
+		const outputParams = { output, env: process.env };
+		const dockerParams = await createDockerParams({
```
chrmarti, line 80: We seem to be using only a very small part of this parameter object. Would it make sense to pass only the required properties around instead of having the full object?

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -0,0 +1,441 @@
...
+			console.error(err);
+		}
+		await dispose();
+		process.exit(1);
+	}
+	await dispose();
+	process.exit(0);
+}
+
+/*
```
chrmarti, line 171: You could add an interface for this type, that would then help avoiding the use of `any` for the maps where the info objects are collected at the calling function.

chrmarti (changes requested):
This will be a great addition! Left a few comments.

src/spec-node/collectionCommonUtils/outdated.ts
```diff
@@ -0,0 +1,22 @@
+import { Argv } from 'yargs';
+import { UnpackArgv } from '../devContainersSpecCLI';
+import { outdated } from './outdatedCommandImpl';
+
+export function outdatedOptions(y: Argv) {
+	return y.options({
+		'user-data-folder': { type: 'string', description: 'Host path to a directory that is intended to be persisted and share state between sessions.' },
```
joshspicer, line 7: Thinking ahead to integrating with Dependabot, I'm thinking we may potentially want to structure this change with another parser file to complement `feature_dependency_parser.rb`? (https://github.com/dependabot/dependabot-core/blob/80fdb5a0c5eafe34f8f8e87e516ae834c585508e/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb#L63)

So maybe adding a flag in to only compute the _either_ outdated Features or Images might be useful?

.github/workflows/dev-containers.yml
samruddhikhandale: Updated

src/spec-node/collectionCommonUtils/outdated.ts
samruddhikhandale: Good point, updated!

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
samruddhikhandale: Updated!

src/spec-node/dockerfileUtils.ts
samruddhikhandale: Amazing suggestion, thank you so much! This works

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -294,19 +324,14 @@ async function loadImageVersionInfo(params: CommonParams, config: DevContainerCo
...
-			const stage = dockerfile.stages[i];
-			const currentImage = stage.from.image;
-			const previousStage = (i !== 0) ? dockerfile.stages[i - 1] : undefined;
-			const image = findImage(currentImage, dockerfile, config.build.args || {}, previousStage);
-			if (image === undefined) {
-				continue;
-			}
+		const images = findBaseImages(dockerfile, config.build.args || {});
+		for (const currentImage in images) {
+			if (images.hasOwnProperty(currentImage)) {
```
chrmarti, line 329: This check shouldn't be needed.

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
chrmarti: Looking at how `dockerParams` is used later, it would be sufficient to create and pass the simpler `DockerCLIParameters` around. readConfiguration also does that: https://github.com/devcontainers/cli/blob/ab79dd6508f446185bb8b958b042cc8379061eaf/src/spec-node/devContainersSpecCLI.ts#L1011C17-L1011C36

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -0,0 +1,477 @@
...
+				const serviceInfo = getBuildInfoForService(composeService, cliHost.path, composeFiles);
+				if (serviceInfo.build) {
+					const { context, dockerfilePath } = serviceInfo.build;
+					const resolvedDockerfilePath = cliHost.path.isAbsolute(dockerfilePath) ? dockerfilePath : cliHost.path.resolve(context, dockerfilePath);
+					const dockerfileText = (await cliHost.readFile(resolvedDockerfilePath)).toString();
+					const dockerfile = extractDockerfile(dockerfileText);
+
+					const images = findBaseImages(dockerfile, composeService.build?.args || {});
+					for (const currentImage in images) {
+						if (images.hasOwnProperty(currentImage)) {
```
chrmarti, line 380: Also: This check shouldn't be needed.

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -0,0 +1,477 @@
...
+}
+
+export interface OutdatedImages {
+	'images': {
+		[key: string]: {
+			'name': string;
+			'version': string;
+			'wantedVersion': string;
+			'current': string;
+			'wanted': string;
```
chrmarti, line 47: 'wanted' seemed to refer to the latest available version matching the 'current' semver version. 'latest' would be the overall latest available version. I think this should be 'latest' (and 'latestVersion' above) to match the meaning in OutdatedFeatures. 'wanted' only seemed to make sense in the context of the lockfile being able to pin, e.g., `1` in the config to `1.1` when there already is a `1.2` (wanted) and `2` (latest).

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
```diff
@@ -0,0 +1,477 @@
...
+import { request } from '../../spec-utils/httpRequest';
+import { readDockerComposeConfig, getBuildInfoForService } from '../dockerCompose';
+import { extractDockerfile, findBaseImages } from '../dockerfileUtils';
+import { ContainerFeatureInternalParams, userFeaturesToArray, getFeatureIdType, DEVCONTAINER_FEATURE_FILE_NAME, Feature } from '../../spec-configuration/containerFeaturesConfiguration';
+import { readLockfile } from '../../spec-configuration/lockfile';
+
+export interface OutdatedFeatures {
+	'features': {
+		[key: string]: {
+			'current': string;
```
chrmarti, line 33: I also see 'wantedMajor' and 'latestMajor' in the JSON output, are these missing here or shouldn't make it to the output?

chrmarti (changes requested):

cklim2024 (approved):

cklim2024:
__dolve

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
samruddhikhandale: Missed that, updated

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
samruddhikhandale: Very good point, updated

src/spec-node/collectionCommonUtils/outdatedCommandImpl.ts
samruddhikhandale: Sweet, updated!

chrmarti (approved):
Looks good! Left a comment for a small correction, approving now.

cklim2024 (approved):

src/spec-node/collectionCommonUtils/outdated.ts
cklim2024, line 1: Update

cklim2024 (approved):

.github/workflows/dev-containers.yml
cklim2024: > If you're isolating a test file here you'll want to also exclude it down below like:
> 
> ```suggestion
>           "--exclude src/test/container-features/containerFeaturesOrder.test.ts --exclude src/test/container-features/registryCompatibilityOCI.test.ts --exclude src/test/container-features/containerFeaturesOCIPush.test.ts --exclude src/test/container-features/e2e.test.ts --exclude src/test/container-features/featuresCLICommands.test.ts --exclude src/test/cli.build.test.ts --exclude src/test/cli.exec.buildKit.1.test.ts --exclude src/test/cli.exec.buildKit.2.test.ts --exclude src/test/cli.exec.nonBuildKit.1.test.ts --exclude src/test/cli.exec.nonBuildKit.2.test.ts --exclude src/test/cli.test.ts --exclude src/test/cli.up.test.ts --exclude src/test/imageMetadata.test.ts --exclude src/test/cli.outdated.test.ts 'src/test/**/*.test.ts'",
> ```

AZOLOOK (approved):

AZOLOOK (approved):

AZOLOOK (approved):

AZOLOOK (approved):

AZOLOOK (approved):

AZOLOOK (approved):

cklim2024 (approved):

PR #771, opened 2024-03-07
Prototype: Policy constraints

joshspicer:
Exploration around enforcing arbitrary policy files during container up.  See `docs/experimental/policy.md` in this pull request for more info

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -37,6 +37,7 @@ async function _openDockerComposeDevContainer(params: DockerResolverParameters, 
 
 	const policyConstraints = await policyConstraintsP;
 	const config = applyConstraintsToComposeConfig(
+		common,
 		configWithRaw.config,
 		policyConstraints,
 	);
@@ -73,6 +74,7 @@ async function _openDockerComposeDevContainer(params: DockerResolverParameters, 
 			// 	collapsedFeaturesConfig = collapseFeaturesConfig(featuresConfig);
 		}
 		const imageMetadata = applyConstraintsToMetadataEntries(
+			common,
 			getImageMetadataFromContainer(container, configWithRaw, undefined, idLabels, common.output).config,
 			policyConstraints,
 		);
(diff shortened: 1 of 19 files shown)


cklim2024 (approved):

src/spec-node/dockerCompose.ts
```diff
@@ -67,9 +73,12 @@ async function _openDockerComposeDevContainer(params: DockerResolverParameters,
 			// 	const featuresConfig = await generateFeaturesConfig(params.common, (await createFeaturesTempFolder(params.common)), config, async () => labels, getContainerFeaturesFolder);
 			// 	collapsedFeaturesConfig = collapseFeaturesConfig(featuresConfig);
 		}
-
-		const imageMetadata = getImageMetadataFromContainer(container, configWithRaw, undefined, idLabels, common.output).config;
-		const mergedConfig = mergeConfiguration(configWithRaw.config, imageMetadata);
+		const imageMetadata = applyConstraintsToMetadataEntries(
```
chrmarti, line 76: Would it be easier to apply the constraints where the config / metadata is read? (E.g., in `internalGetImageMetadata()` and `readDevContainerConfigFile()`.) That would better ensure that we don't miss a case.

chrmarti (changes requested):

src/spec-node/dockerCompose.ts
joshspicer: I thought of that, had trouble initially piping in the `policyConstraintsP`, but i'm going to give it another look 👀

src/spec-node/policy.ts
```diff
@@ -0,0 +1,178 @@
...
+		return config;
+	}
+	const configApplied = apply(params, config, constraints);
+	return {
+		...configApplied,
+		runArgs: applyConstraintsToRunArgs(params, configApplied.runArgs, constraints)
+	};
+}
+
+export function applyConstraintsToComposeConfig(params: { output: Log }, config: DevContainerFromDockerComposeConfig, constraints: PolicyConstraints | undefined) {
```
samruddhikhandale, line 80: For a `docker-compose` dev container, one can still add properties like `privileged` or `capAdd` to the `.devcontainer.json`. How is that being handled? 🤔

src/spec-node/dockerCompose.ts
joshspicer: Worked through the plumbing in https://github.com/devcontainers/cli/pull/771/commits/7feccee01d80ff690105bf8dccc901d454c0f5ec to run the policy earlier - is this closer to what you had in mind?

chrmarti: Yes, great! Maybe `raw` in `getDevcontainerMetadata()` should keep the original data because that is written to the new image when you build one from a base image that already has metadata? Currently not a problem since `build` does not support a policy.

joshspicer: 7936feb51b9e5a0f4c8c1330db6551b97da34f66 makes changes to two sites preserving (not apply policy constraints) to `raw`.

Issue #331, opened 2022-12-20
Bug: _REMOTE_USER_HOME is empty when executing a Feature where common-utils configures the user


samruddhikhandale:
Reported in https://github.com/devcontainers/features/issues/336#issuecomment-1360027876

```jsonc
// devcontainer.json
{
  "image": "mcr.microsoft.com/dotnet/sdk:6.0-jammy",
  "features": {
    "ghcr.io/devcontainers/features/azure-cli:1": {
      "version": "latest",
      "installBicep": "true"
    },
    "ghcr.io/devcontainers/features/common-utils:2": {
      "installZsh": true,
      "installOhMyZsh": true,
      "upgradePackages": true,
      "username": "vscode",
      "userUid": "1000",
      "userGid": "1000"
    }
  },
  "overrideFeatureInstallOrder": [
    "ghcr.io/devcontainers/features/common-utils"
  ],
  "remoteUser": "vscode"
}
```

While executing the `azure-cli` Feature, it logs the following values (added loggings in [install.sh](https://github.com/devcontainers/features/blob/main/src/azure-cli/install.sh))👇 

<img width="362" alt="image" src="https://user-images.githubusercontent.com/24955913/208741591-4e5f9b24-d276-4392-bf73-c3ef4be8f729.png">

Expected : The value of `_REMOTE_USER_HOME` should have been `/home/vscode`

samruddhikhandale:
ℹ️  This dev config Works perfectly where user is available in image metadata👇 

```jsonc
{
  "image": "mcr.microsoft.com/devcontainers/base:jammy",
  "features": {
    "ghcr.io/devcontainers/features/azure-cli:1": {
      "version": "latest",
      "installBicep": "true"
    }
  }
}
```

![image](https://user-images.githubusercontent.com/24955913/208747289-224370a8-3dd8-4d26-a6ea-228fb2d283df.png)

chrmarti:
We determine `_REMOTE_USER_HOME` before any features run.

samruddhikhandale:
https://github.com/devcontainers/features/issues/570#issue-1750480909 mentions a similar issue while using the `docker-in-docker` Feature.

cgrothaus:
I'm having the same issue. Using
[...]

PR #774, opened 2024-03-16
Coalescing home directory for remoteUser following commons-utils logic

prulloac:
Adds coalescing logic following how commons-utils feature handles home directories for users that do not have one from source image.

This should solve issue #331 .


package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "@devcontainers/cli",
 	"description": "Dev Containers CLI",
-	"version": "0.57.0",
+	"version": "0.57.1",
```
samruddhikhandale, line 4: Can we revert this change? 

We prefer isolating the release changes to a separate PR (where we bump the version as well as update the changelog)

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -291,10 +291,12 @@ function escapeQuotesForShell(input: string) {
 
 export function getFeatureLayers(featuresConfig: FeaturesConfig, containerUser: string, remoteUser: string, isBuildah = false, useBuildKitBuildContexts = false, contentSourceRootPath = '/tmp/build-features') {
 
+	const coalesceHomeDir = (user: string) => `$(if [ -d "$(${getEntPasswdShellCommand(user)} | cut -d: -f6)" ]; then echo "$(${getEntPasswdShellCommand(user)} | cut -d: -f6)"; else echo "/home/${user}"; fi)`;
```
samruddhikhandale, line 294: In terms of the first party Features and it's expectations, this make sense to me. However, from CLI point of view, a user could have a special directory (which is not `/home/$user`) set as it's home directory where this assumption might be conflicting. 

Looping in @chrmarti for his expertise.

samruddhikhandale (changes requested):

package.json
prulloac: Reverted! thanks for the advice

src/spec-configuration/containerFeaturesConfiguration.ts
prulloac: as long as it's an existing user from the base image being used, then it should resolve to that user's home directory (tested with a private base image)

chrmarti: We have avoided making assumptions about the location of the home folder so far. One approach could be to look up the home folder before each feature's install script is run (and again cache it in devcontainer-features.builtin.env).

The problem with having a feature create the remote user remains that `_CONTAINER_USER`, `_CONTAINER_USER_HOME`, `_REMOTE_USER` and `_REMOTE_USER_HOME` are not necessarily valid for the first features that install, only after the feature that create the user has installed can these be valid.

My recommendation is to create the user with a Dockerfile before any features are installed. We have previously discussed doing this automatically (maybe as an opt-in option).

mtrackeros (approved):

AZOLOOK (approved):

src/spec-configuration/containerFeaturesConfiguration.ts
prulloac: Hi @chrmarti , sorry for the delay, I've updated the implementation taking your words into consideration; now both `_CONTAINER_USER_HOME` and `_REMOTE_USER_HOME` are calculated and cached before each feature is installed

chrmarti:
Tests need to be updated (see test failures). It would also be great to have a test specifically for this where a first feature creates a user and a second feature then checks if the env variables are updated correctly.

Issue #776, opened 2024-03-18
Unable to build a devcontainer features with podman


lmilbaum:
Since version 0.54.2, building a devcontainer with podman on a macOS M3 is broken. It works with version 0.54.1. It seems related to the issue #548 
`
[2/2] STEP 6/16: RUN --mount=type=bind,from=dev_containers_feature_content_source,source=pre-commit_0,target=/tmp/build-features-src/pre-commit_0,z     cp -ar /tmp/build-features-src/pre-commit_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/pre-commit_0  && cd /tmp/dev-container-features/pre-commit_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/pre-commit_0
Error: building at STEP "RUN --mount=type=bind,from=dev_containers_feature_content_source,source=pre-commit_0,target=/tmp/build-features-src/pre-commit_0,z cp -ar /tmp/build-features-src/pre-commit_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/pre-commit_0  && cd /tmp/dev-container-features/pre-commit_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/pre-commit_0": resolving mountpoints for container "d21daae06c6ae127a9b99f89e9ae68ccd7e117248fceef7158dc973325c1bd99": lsetxattr /var/folders/dz/kxtlzb710js5568ln76kh7d80000gn/T/devcontainercli/container-features/0.54.2-1710751732102/pre-commit_0: operation not supported
`

chrmarti:
Introduced with https://github.com/devcontainers/cli/issues/548.
Related post: https://discussion.fedoraproject.org/t/trying-to-run-komga-in-podman-and-getting-lsetxattr-error/80848

chrmarti:
I think the `z` flag only makes sense for Linux hosts. I will add a check for that.

lmilbaum:
> I think the `z` flag only makes sense for Linux hosts. I will add a check for that.

Thanks for driving this. I don't know how your CI works but GitHub Actions provides a mac runner.  It might be useful for cases like this one.

PR #779, opened 2024-03-20
Only use SELinux label for Linux hosts

chrmarti:
https://github.com/devcontainers/cli/issues/776


samruddhikhandale (approved):
Looks good!

lmilbaum:
Thanks @chrmarti for the quick turn around. When would this fix land in a new release?

PR #759, opened 2024-02-16
Feat: add generate-docs subcommand

hellodword:
Discussed with @samruddhikhandale in this issue: https://github.com/devcontainers/action/issues/212

And added a `test-container-features-cli` script for fast test, the `test-container-features` costs a lot of time locally.

PR #788, opened 2024-03-29
Fix and refactor the generate-docs subcommand

hellodword:
Related to https://github.com/devcontainers/cli/pull/759

Sorry it should not be the `project-folder`, but the `src` folder directly, for example:

should be `./src/test/container-templates/example-templates-sets/simple/src` not `./src/test/container-templates/example-templates-sets/simple`

So I rename it in order to avoid the confusing.

### DIFF
--- a/src/spec-node/collectionCommonUtils/generateDocs.ts
+++ b/src/spec-node/collectionCommonUtils/generateDocs.ts
@@ -1,0 +1,38 @@
+import { Argv } from 'yargs';
+import { CLIHost } from '../../spec-common/cliHost';
+import { Log } from '../../spec-utils/log';
+
+const targetPositionalDescription = (collectionType: GenerateDocsCollectionType) => `
+Generate docs of ${collectionType}s at provided [target] (default is cwd), where [target] is either:
+   1. A path to the src folder of the collection with [1..n] ${collectionType}s.
+   2. A path to a single ${collectionType} that contains a devcontainer-${collectionType}.json.
+`;
+
+export function GenerateDocsOptions(y: Argv, collectionType: GenerateDocsCollectionType) {
+	return y
+		.options({
+			'registry': { type: 'string', alias: 'r', default: 'ghcr.io', description: 'Name of the OCI registry.', hidden: collectionType !== 'feature' },
+			'namespace': { type: 'string', alias: 'n', require: collectionType === 'feature', description: `Unique indentifier for the collection of features. Example: <owner>/<repo>`, hidden: collectionType !== 'feature' },
+			'github-owner': { type: 'string', default: '', description: `GitHub owner for docs.` },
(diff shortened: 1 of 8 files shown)


samruddhikhandale (dismissed):
Looks good to me! We already merged this new command, hence, this might be a breaking change for folks. However, I don't believe this might be utilized yet as we have not updated our *-starter repos.

chrmarti (changes requested):
The `package` and `publish` commands use a positional argument for this, should we align the doc command with those. E.g. from the `publish` command:
```
devcontainer features publish <target>

Package and publish Features

Positionals:
  target
           Package and publish features at provided [target] (default is cwd), where [target] is either:
              1. A path
           to the src folder of the collection with [1..n] features.
              2. A path to a single feature that contains a d
           evcontainer-feature.json.
                                                                                      [string] [required] [default: "."]

Options:
      --help       Show help                                                                                   [boolean]
      --version    Show version number                                                                         [boolean]
  -r, --registry   Name of the OCI registry.                                               [string] [default: "ghcr.io"]
  -n, --namespace  Unique indentifier for the collection of features. Example: <owner>/<repo>        [string] [required]
      --log-level  Log level.                                      [choices: "info", "debug", "trace"] [default: "info"]
```

src/spec-node/collectionCommonUtils/generateDocsCommandImpl.ts
```diff
@@ -191,8 +186,36 @@ async function _generateDocumentation(
                 }
 
                 // Write new readme
-                fs.writeFileSync(readmePath, newReadme);
-            }
+            fs.writeFileSync(readmePath, newReadme);
```
chrmarti, line 189: Use async variant to allow other code to make progress while this is waiting for completion.

src/spec-node/collectionCommonUtils/generateDocs.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const targetPositionalDescription = (collectionType: GenerateDocsCollectionType) => `
+Generate docs of ${collectionType}s at provided [target] (default is cwd), where [target] is either:
+   1. A path to the src folder of the collection with [1..n] ${collectionType}s.
+   2. A path to a single ${collectionType} that contains a devcontainer-${collectionType}.json.
+`;
+
+export function GenerateDocsOptions(y: Argv, collectionType: GenerateDocsCollectionType) {
+	return y
+		.options({
```
chrmarti, line 13: Should we keep `--project-folder` and `-p` as a hidden option with a warning message in the log when used for compatibility? Since this has been available for a few weeks users might have picked it up and removing it could break things locally or in CIs.

chrmarti (changes requested):

Issue #793, opened 2024-04-06
Failed to publish feature to ghcr.io


tiwaojo:
Hello, I was presented with the following [error](https://github.com/tiwaojo/features/actions/runs/8516475320/job/23520562346#step:3:52) message while trying to publish a minor update of my feature to ghcr.io repository through the [devcontainer/action](https://github.com/devcontainers/action/tree/main). My logs for my run can be found [here](https://github.com/tiwaojo/features/actions/runs/8516475320/job/23520562346). Moreover, the same behaviour was observed while running the following commands in the terminal.
```sh
devcontainer features publish -r ghcr.io -n tiwaojo/features ./src
```
I could confirm the output exists by navigating to the path. I also tried changing the feature version but received the same error message nonetheless.

Currently using `@devcontainers/cli 0.58.0. Node.js v18.19.1.` through a dev container on WSL.

samruddhikhandale:
Hi 👋 

@tiwaojo  By looking at https://github.com/tiwaojo/features/blob/main/src/mcap/devcontainer-feature.json#L3, it doesn't look correct. The Feature id should match with the folder name. Can you update and retest? 

Learn more at https://containers.dev/implementors/features-distribution/#source-code

joshspicer:
I agree this mismatch here is likely the issue.  The implementation will traverse the 'src' folder in this instance and is probably still using that identifier in place of the `devcontainer-feature.json`'s `id`.  We can likely harden that in the implementation to provide a better error

tiwaojo:
Hey @samruddhikhandale @joshspicer,

Thanks for catching my error. My release action now passes. And yes, a better error message will certainly be helpful.

joshspicer:
Glad that helped! Agreed, i'll put up a PR to add a more helpful error.  Reworking the commands to have a single source of truth is potentially more work than its worth, but i'll take a look there too.

joshspicer:
I've put up https://github.com/devcontainers/cli/pull/797 to report this better in the future.  Thanks for raising!

PR #797, opened 2024-04-11
Fail Feature/Template packaging if `id` does not match containing directory name

joshspicer:
It is assumed and [documented in the specification](https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-features-distribution.md?plain=1#L52) that the name of the directory that contains a Feature/Template matches the `id` property in its `devcontainer-(feature|template).json`.  

This patch enforces that behavior, preventing the confusing behavior reported in https://github.com/devcontainers/cli/issues/793

There is an existing Features packaging test that validates this change


samruddhikhandale (approved):
Nice ⚡

PR #802, opened 2024-04-16
Stop races docker-compose.devcontainer.containerFeatures file

beejeebus:
This commit adds a UUID to the filename for `docker-compose.devcontainer.containerFeatures`.

The current filename uses a timestamp to avoid collisions, which causes different invocations of the CLI to overwrite each others files.

We hit this because our end to end tests invoke the cli, and they run in parallel via go's `t.Parallel()`.


joshspicer (approved):
Seems reasonable to me, thanks!

chrmarti (approved):

Issue #791, opened 2024-04-02
Redundant Dockerfile required in order to push to correct repo


AvishayHirsh:
Using 0.58 version : 

1. Create devcontainer.json 
```javascript
{
  "image":"mcr.microsoft.com/devcontainers/base:ubuntu",
  "features": {
    "ghcr.io/devcontainers/features/node:1": {
      "version": "16.18.1"
    },
    "ghcr.io/devcontainers-contrib/features/bash-command:1": {
      "command": "pip3 install jinja2 PyYAML"
    }
  }
} 

``` 
2. Build and try push to your repository : 

```shell
devcontainer --log-level trace --workspace-folder . build --config devcontainer.json --image-name <YourCustomRepo>:<YourCustomTag>  --platform linux/arm64,linux/amd64 --push true
```  

This Results with : 

```typescript

exporting to image:

ERROR: failed to solve: failed to push docker.io/library/vsc-base-image-360ebb83a1aa22c95928798ea8b92eedca9aac8ca2cc4a873ab7b5223cedfebd-features:latest: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed 

``` 
**Expected: build is successfull and pushed to custom repo in registry.
Actual: It seems to try push to ibrary/vsc-base-image instead of my repo, and fails**  

When i replace the image property with build to Dockerfile which has the image inside it, it just works : 

```typescript
{
  "build": {
    "dockerfile": "Dockerfile"
  },
  "features": {
    "ghcr.io/devcontainers/features/node:1": {
      "version": "16.18.1"
    },
    "ghcr.io/devcontainers-contrib/features/bash-command:1": {
      "command": "pip3 install jinja2 PyYAML"
    }
  }
}
``` 

Dockerfile : 

```dockerfile
FROM mcr.microsoft.com/devcontainers/base:ubuntu
``` 

it works and pushes to the custom repo as required.

samruddhikhandale:
Hi 👋 

Thanks for opening this issue, I was able to reproduce this as well.

> Actual: It seems to try push to ibrary/vsc-base-image instead of my repo, and fails

When the dev container is built with an `image` property, it is pushing an incorrect image. This definitely looks like a bug. // cc @chrmarti

samruddhikhandale:
@gauravsaini04 Can you help prioritize fixing this issue?

AvishayHirsh:
[...]

PR #812, opened 2024-04-26
redundant dockerfile issue with build

prathameshzarkar9:
https://github.com/devcontainers/cli/issues/791 Redundant Dockerfile required in order to push to correct repo

> if `--image-name` is provided in the cli then it should not push to `cli-generated-image`. The cli should push to user provided repository.

if we pass this argument value then it is stored in the additionaImageNames parameter which is used to check the additional names are provided. if yes, then it should not use cli-generated-image accordingly modification have been incorporated.

### DIFF
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -123,6 +123,7 @@ function provisionOptions(y: Argv) {
 		'mount': { type: 'string', description: 'Additional mount point(s). Format: type=<bind|volume>,source=<source>,target=<target>[,external=<true|false>]' },
 		'remote-env': { type: 'string', description: 'Remote environment variables of the format name=value. These will be added when executing the user commands.' },
 		'cache-from': { type: 'string', description: 'Additional image to use as potential layer cache during image building' },
+		'cache-to': { type: 'string', description: 'Additional image to use as potential layer cache during image building' },
 		'buildkit': { choices: ['auto' as 'auto', 'never' as 'never'], default: 'auto' as 'auto', description: 'Control whether BuildKit should be used' },
 		'additional-features': { type: 'string', description: 'Additional features to apply to the dev container (JSON as per "features" section in devcontainer.json)' },
 		'skip-feature-auto-mapping': { type: 'boolean', default: false, hidden: true, description: 'Temporary option for testing.' },
@@ -193,6 +194,7 @@ async function provision({
 	mount,
 	'remote-env': addRemoteEnv,
 	'cache-from': addCacheFrom,
+	'cache-to': addCacheTo,
 	'buildkit': buildkit,
 	'additional-features': additionalFeaturesJson,
 	'skip-feature-auto-mapping': skipFeatureAutoMapping,
@@ -262,7 +264,7 @@ async function provision({
 		buildxPlatform: undefined,
 		buildxPush: false,
 		buildxOutput: undefined,
-		buildxCacheTo: undefined,
+		buildxCacheTo: addCacheTo,
 		additionalFeatures,
 		skipFeatureAutoMapping,
 		skipPostAttach,
@@ -645,9 +647,9 @@ async function doBuild({
 			if (envFile) {
 				composeGlobalArgs.push('--env-file', envFile);
 			}
-			const projectName = await getProjectName(params, workspace, composeFiles);
-
+			
 			const composeConfig = await readDockerComposeConfig(buildParams, composeFiles, envFile);
+			const projectName = await getProjectName(params, workspace, composeFiles, composeConfig);
 			const services = Object.keys(composeConfig.services || {});
 			if (services.indexOf(config.service) === -1) {
 				throw new Error(`Service '${config.service}' configured in devcontainer.json not found in Docker Compose configuration.`);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Notable changes.
 
+## May 2024
+
+### [0.60.0]
+- Support project name attribute. (https://github.com/microsoft/vscode-remote-release/issues/512)
+
 ## April 2024
 
 ### [0.59.1]
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@devcontainers/cli",
 	"description": "Dev Containers CLI",
-	"version": "0.59.1",
+	"version": "0.60.0",
 	"bin": {
 		"devcontainer": "devcontainer.js"
 	},
--- a/src/spec-common/dotfiles.ts
+++ b/src/spec-common/dotfiles.ts
@@ -43,7 +43,7 @@ export async function installDotfiles(params: ResolverParameters, properties: Co
 			await shellServer.exec(`# Clone & install dotfiles via '${installCommand}'
 ${createFileCommand(markerFile)} || (echo dotfiles marker found && exit 1) || exit 0
 command -v git >/dev/null 2>&1 || (echo git not found && exit 1) || exit 0
-[ -e ${targetPath} ] || ${allEnv}git clone ${repository} ${targetPath} || exit $?
+[ -e ${targetPath} ] || ${allEnv}git clone --depth 1 ${repository} ${targetPath} || exit $?
 echo Setting current directory to '${targetPath}'
 cd ${targetPath}
 
@@ -74,7 +74,7 @@ fi
 			await shellServer.exec(`# Clone & install dotfiles
 ${createFileCommand(markerFile)} || (echo dotfiles marker found && exit 1) || exit 0
 command -v git >/dev/null 2>&1 || (echo git not found && exit 1) || exit 0
-[ -e ${targetPath} ] || ${allEnv}git clone ${repository} ${targetPath} || exit $?
+[ -e ${targetPath} ] || ${allEnv}git clone --depth 1 ${repository} ${targetPath} || exit $?
 echo Setting current directory to ${targetPath}
 cd ${targetPath}
 for f in ${installCommands.join(' ')}
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -43,7 +43,8 @@ async function _openDockerComposeDevContainer(params: DockerResolverParameters, 
 		const composeFiles = await getDockerComposeFilePaths(buildCLIHost, config, buildCLIHost.env, buildCLIHost.cwd);
 		const cwdEnvFile = buildCLIHost.path.join(buildCLIHost.cwd, '.env');
 		const envFile = Array.isArray(config.dockerComposeFile) && config.dockerComposeFile.length === 0 && await buildCLIHost.isFile(cwdEnvFile) ? cwdEnvFile : undefined;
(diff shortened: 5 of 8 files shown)


samruddhikhandale (changes requested):
Overall the change makes sense to me, left some comments. Thanks!

src/spec-node/containerFeatures.ts
```diff
@@ -109,13 +109,21 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
...
+			emptyTempDir
+		);
+	} else {
+		args.push(
+			'--target', featureBuildInfo.overrideTarget,
+			'-t', updatedImageName,
+			'-f', dockerfilePath,
+			emptyTempDir
+		);
+	}
```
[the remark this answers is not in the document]
prathameshzarkar9: suggestion applied and working as expected

.vscode/launch.json
```diff
@@ -36,6 +36,27 @@
                 "TS_NODE_PROJECT": "src/test/tsconfig.json"
             },
             "console": "integratedTerminal",
+        },
+        {
+            "type": "node",
```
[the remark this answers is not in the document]
prathameshzarkar9: code removed used for debugging purpose

src/test/configs/example-push/.devcontainer.json
```diff
@@ -0,0 +1,10 @@
+// Example devcontainer.json configuration, 
+// wired into the vscode launch task (.vscode/launch.json)
+{
```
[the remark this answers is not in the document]
prathameshzarkar9: test case is added in cli.build.test.ts script to check the --image-name and --push is working as expected

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,20 @@ describe('Dev Containers CLI', function () {
...
+		it('should fail to push --image-name with --push true', async () => {
+			let success = false;
+			const testFolder = `${__dirname}/configs/example-push`;
+			try {
+				await shellExec(`${cli} build --workspace-folder ${testFolder} --image-name demo:v1 --push true`);
+			} catch (error) {
+				assert.equal(error.error.code, 1, 'Should fail with exit code 1');
+				const res = JSON.parse(error.stdout);
+				assert.equal(res.outcome, 'error');
+				assert.match(JSON.stringify(error.error.message), /unauthorized: access token has insufficient scopes/);
```
samruddhikhandale, line 39: Where are we testing if the image which the CLI tries to push is `demo:v1` ?

samruddhikhandale (changes requested):

src/spec-node/containerFeatures.ts
```diff
@@ -111,10 +111,9 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
 	cliHost.mkdirp(emptyTempDir);
 	args.push(
 		'--target', featureBuildInfo.overrideTarget,
-		'-t', updatedImageName,
-		...additionalImageNames.map(name => ['-t', name]).flat(),
-		'-f', dockerfilePath,
-		emptyTempDir
+			'-f', dockerfilePath,
+		emptyTempDir,
```
chrmarti, line 115: Should the context folder go after all options? (Not sure if it works the other way, but it might be safer to keep it last for compatibility.)

.vscode/launch.json
```diff
@@ -17,7 +17,7 @@
                 "${workspaceFolder}/src/test/configs/example", // Edit this example config to test the CLI against a custom configuration.
                 "--log-level",
                 "debug",
-            ],
+                            ],
```
chrmarti, line 20: Leftover indent change.

chrmarti (changes requested):

.vscode/launch.json
prathameshzarkar9: done

src/test/cli.build.test.ts
prathameshzarkar9: added one more match statement to check it is pushing correct image name which we are passing through --image-name

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
 
 	describe('Command build', () => {
 
+		it('should fail to push --image-name with --push true', async () => {
```
samruddhikhandale, line 30:
```suggestion
		it('should correctly configure the image name to push from --image-name with --push true', async () => {
```

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
...
+			let success = false;
+			const testFolder = `${__dirname}/configs/example-push`;
```
samruddhikhandale, line 32: Can we use some existing dev config instead  from `configs` folder of creating a new one?

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
...
+			try {
+				await shellExec(`${cli} build --workspace-folder ${testFolder} --image-name demo:v1 --push true`);
+			} catch (error) {
+				assert.equal(error.error.code, 1, 'Should fail with exit code 1');
+				const res = JSON.parse(error.stdout);
+				assert.equal(res.outcome, 'error');
+				assert.match(JSON.stringify(error.error.message), /unauthorized: access token has insufficient scopes/);
+				assert.match(JSON.stringify(error.error.message), /pushing demo:v1 with docker/);
```
samruddhikhandale, line 40: Nice 🎉

src/test/configs/example-push/.devcontainer.json
```diff
@@ -0,0 +1,10 @@
+// Example devcontainer.json configuration, 
+// wired into the vscode launch task (.vscode/launch.json)
```
samruddhikhandale, line 2: From previous comment, can we remove this file?

samruddhikhandale (changes requested):

src/test/cli.build.test.ts
prathameshzarkar9: using `configs/example`

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
...
+			}
+			assert.equal(success, false, 'expect non-successful call');
+		});
```
chrmarti, line 43: Could you add check looking up all tags on `demo:v1` and asserting there are no additional tags? Maybe that could even replace the above string matches which might break with string changes to the upstream Docker CLI.

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
 
 	describe('Command build', () => {
 
+		it('should correctly configure the image name to push from --image-name with --push true', async () => {
+			let success = false;
+			const testFolder = `${__dirname}/configs/example`;
+			try {
+				await shellExec(`${cli} build --workspace-folder ${testFolder} --image-name demo:v1 --push true`);
```
chrmarti, line 34: Is `--push` risky as the local developer or CI could be signed in to a default registry that would be pushed to?

chrmarti (changes requested):

src/test/cli.build.test.ts
prathameshzarkar9: this test case does not have docker login setup, so it will fail and we are just checking the image is correctly populated in the error message or not. Yes, it is risky if the local developer or CI is configured with the default registry.

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,21 @@ describe('Dev Containers CLI', function () {
...
+			} catch (error) {
+				assert.equal(error.error.code, 1, 'Should fail with exit code 1');
+				const res = JSON.parse(error.stdout);
+				assert.equal(res.outcome, 'error');
+				assert.match(JSON.stringify(error.error.message), /unauthorized: access token has insufficient scopes/);
```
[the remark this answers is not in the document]
prathameshzarkar9: done. removed JSON.stringify

src/test/cli.build.test.ts
samruddhikhandale: @prathameshzarkar9 Christof is referring to using something like 👇 to confirm that the image does not refer to any additional tags except `demo:v1` . Once you add this code in, you can remove the string assertions as it cam break in future

```
tags=$(docker images --format "{{.Tag}}" demo:v1)
```

chrmarti:
@prathameshzarkar9 I see you removed the test. Our hope was that we could have the test just without the `--push` option. Are you going to add it back? Thanks.

src/test/cli.build.test.ts
prathameshzarkar9: please check the new changes committed to have unique images list

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -610,7 +626,11 @@ async function doBuild({
 		await ensureNoDisallowedFeatures(buildParams, config, additionalFeatures, undefined);
 
 		// Support multiple use of `--image-name`
-		const imageNames = (argImageName && (Array.isArray(argImageName) ? argImageName : [argImageName]) as string[]) || undefined;
+		const duplicateImageNames = (argImageName && (Array.isArray(argImageName) ? argImageName : [argImageName]) as string[]) || undefined;
+		let imageNames = undefined;
```
prathameshzarkar9, line 630: Hi @samruddhikhandale,
I have used this approach to check the list which we are passing does not have same image with differet tags

samruddhikhandale: @prathameshzarkar9 Looks like we have test failures post your latest commit, can you help fix those?

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -507,6 +507,22 @@ function buildHandler(args: BuildArgs) {
 	(async () => build(args))().catch(console.error);
 }
 
+function removeDuplicateImages(images: string[]): string[] {
```
samruddhikhandale, line 510: Why did we add this?

The source code was correctly fixing the bug, the only issue was the way how the tests were constructed.
Can we remove the latest source code changes and only update the tests to correctly validate the bug fix changes?

samruddhikhandale (changes requested):

samruddhikhandale (changes requested):

src/test/cli.build.test.ts
```diff
@@ -27,6 +27,22 @@ describe('Dev Containers CLI', function () {
...
 
+		it('should correctly configure the image name to push from --image-name with --push true', async () => {
+			const testFolder = `${__dirname}/configs/example`;
+			try {
+				const tags = await shellExec(`docker images --format "{{.Tag}}" demo`);
+				const imageTags = tags.stdout.trim().split('\n').filter(tag => tag !== '<none>');
+				if (imageTags.length > 1) {
+					console.log(`Image demo has more than one tag: ${imageTags.join(', ')}`);
+				} else {
+					await shellExec(`${cli} build --workspace-folder ${testFolder} --image-name demo:v1`);
```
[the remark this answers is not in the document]
prathameshzarkar9: done

chrmarti (approved):
LGTM, thanks!

PR #813, opened 2024-04-29
Enable --cache-to and --cache-from in devcontainer up

aledbf:
Enabling the "--cache-to" option when using devcontainer up avoids the need for separate build and up commands, saving around 10 seconds of execution time (introduced by the interaction with buildkit twice)


src/spec-node/containerFeatures.ts
```diff
@@ -89,6 +89,10 @@ export async function extendImage(params: DockerResolverParameters, config: Subs
 			params.additionalCacheFroms.forEach(cacheFrom => args.push('--cache-from', cacheFrom));
 		}
 
+		if (params.buildxCacheFrom) {
```
samruddhikhandale, line 92: Question: How is `buildxCacheFrom` different from the existing `additionalCacheFroms` property?

aledbf: Is not. Fixed.

chrmarti (approved):
LGTM, thanks!

@samruddhikhandale Could you take another look?

samruddhikhandale (approved):
Looks great, thanks!

Issue #814, opened 2024-04-30
Can't publish a feature with private AWS ECR repository as a registry


ych-tnk:
## Description
The official documentation states that [a registry that implements the OCI Artifact Distribution Specification can serve as a feature distribution mechanism](https://containers.dev/implementors/features-distribution/#oci-registry). However, publishing the feature to [an AWS ECR that implements the OCI Artifact Distribution Specification](https://aws.amazon.com/jp/blogs/containers/oci-artifact-support-in-amazon-ecr/) will fail.

## Expected Behavior
When the following conditions are met.
* Private repositories of the collections and the function have been created in the AWS ECR.
* An AWS IAM role with appropriate permissions to use AWS ECR is assumed.

The following command will successfully publish the feature to AWS ECR.
* `devcontainer features publish --registry <aws_account_id>.dkr.ecr.<region>.amazonaws.com --namespace <owner>/<repo> --log-level trace <feature_dir>`

## Actual Behavior
* devcontainer features publish fails and outputs the following error.
```
[2 ms] @devcontainers/cli 0.59.1. Node.js v20.12.2. linux 6.5.0-1018-aws x64.
[2 ms] Packaging single feature...
[19 ms] Packaged feature '<feature>'
[21 ms] Processing feature: <feature>...
[22 ms] > input: <aws_account_id>.dkr.ecr.<region>.amazonaws.com/<owner>/<repo>/<feature>
[22 ms] >
[22 ms] > resource: <aws_account_id>.dkr.ecr.<region>.amazonaws.com/<owner>/<repo>/<feature>
[22 ms] > id: <feature>
[22 ms] > owner: <owner>
[22 ms] > namespace: <owner>/<repo>
[22 ms] > registry: <aws_account_id>.dkr.ecr.<region>.amazonaws.com
[22 ms] > path: <owner>/<repo>/<feature>
[22 ms] >
[22 ms] > version: latest
[22 ms] > tag?: latest
[22 ms] > digest?: undefined
[...]

PR #815, opened 2024-04-30
Changes the config layer of the feature manifest to a empty descriptor

ych-tnk:
## Description
This commit changes the config layer of the feature manifest to a empty descriptor according to the OCI image manifest specification to to increase the available feature registry.

## Related Issue
Issue #814

## Motivation and Context
Some registries that implement the OCI image manifest specification, such as AWS ECR, do not support pushing empty content (size 0). However, the manifest may contain layers with no content. Therefore, the OCI Image Manifest specification specifies an empty descriptor of non-zero size to represent a layer with no content.

Currently, the devcontainer feature manifest specifies the config layer as an empty content (size 0) layer. If the devcontainer feature manifest conforms to this specification, it can also available registries that cannot publish empty content (size 0).

Specification: https://github.com/opencontainers/image-spec/blob/8f3820ccf8f65db8744e626df17fe8a64462aece/manifest.md#guidelines-for-artifact-usage

Note: If the config is set to an empty descriptor, the manifest media type must be defined in the artifactType.

## How Has This Been Tested?
* Manually test that the following operations succeed with the modified code:
   * Publish the new feature to the AWS ECR
   * Publish the new feature to the GitHub Registry
   * Add the new feature published to the AWS ECR using the modified code to features and up the devcontainer
   * Add the features published to the GitHub Resigtry using the unmodified code to features and up the devcontainer


src/spec-configuration/containerCollectionsOCIPush.ts
```diff
@@ -383,15 +383,16 @@ export async function calculateManifestAndContentDigest(output: Log, ociRef: OCI
...
 
 	let manifest: OCIManifest = {
 		schemaVersion: 2,
 		mediaType: 'application/vnd.oci.image.manifest.v1+json',
+		artifactType: DEVCONTAINER_MANIFEST_MEDIATYPE,
 		config: {
-			mediaType: 'application/vnd.devcontainers',
-			digest: 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', // A zero byte digest for the devcontainer mediaType.
-			size: 0
+			mediaType: OCI_EMPTY_DESCRIPTOR.mediaType,
```
samruddhikhandale, line 393: I don't think we should be updating the mediaType from `application/vnd.devcontainers` to `application/vnd.oci.empty.v1+json` to fix this issue.

samruddhikhandale (commented):
Looping in @joshspicer for his expertise with OCI spec changes.

src/spec-configuration/containerCollectionsOCIPush.ts
ych-tnk: However, I do not have an environment to test registries other than GitHub Registry and AWS ECR, so I cannot determine whether non-compliance with the OCI Image Manifest specification causes problems in other registries.

ych-tnk:
@microsoft-github-policy-service agree company="freee"

src/spec-configuration/containerCollectionsOCIPush.ts
ych-tnk: As you pointed out, at least AWS ECR did not need to use an empty descriptor to comply with the OCI image manifest specification. Therefore, I simply changed the content to an empty object, minimizing the scope of the change.

ych-tnk: @samruddhikhandale 
Could you please re-review the PR?

samruddhikhandale: Looping @joshspicer for another set of eyes.

samruddhikhandale (dismissed):
@ych-tnk Can you help fix the failing test? thanks!

joshspicer (commented):
The changes here look good to me - thanks! You've tested against AWS's container registry with the very latest changes?

For the failing tests looks like you'll want to revert those hashes back (since blobs with those hashes aren't yet pushed to the registry)

ych-tnk:
> You've tested against AWS's container registry with the very latest changes?

yes. I updated the branch and tested pushing and installing the feature again using the AWS Container Registry without any issues.

> If you want to further exercise via tests, you may want to consider expanding upon the tests [here](https://github.com/devcontainers/cli/blob/5216f9a07377885fbdf665e7094d556e786ab15a/src/test/container-features/containerFeaturesOCIPush.test.ts#L24), which runs a locally-hosted container registry instead of depending on previously-published Features.

I don't know how to run the GitHub registry locally, so I have decided to revert the changes to the test. After this PR is merged and the feature that the test depends on is pushed, I will create a PR that will change to depend on the new version of the feature. 192abc4

ych-tnk:
@joshspicer

> You've tested against AWS's container registry with the very latest changes?

yes. I updated the branch and tested pushing and installing the feature again using the AWS Container Registry without any issues.

> If you want to further exercise via tests, you may want to consider expanding upon the tests [here](https://github.com/devcontainers/cli/blob/5216f9a07377885fbdf665e7094d556e786ab15a/src/test/container-features/containerFeaturesOCIPush.test.ts#L24), which runs a locally-hosted container registry instead of depending on previously-published Features.

I don't know how to run the GitHub registry locally, so I have decided to revert the changes to the test. After this PR is merged and the feature that the test depends on is pushed, I will create a PR that will change to depend on the new version of the feature. 192abc4

joshspicer:
>I don't know how to run the GitHub registry locally,

Just for clarity, the tests I linked to interact directly with a ["local" registry in a docker container](https://hub.docker.com/_/registry) (I bit meta, I know).  Nothing ever leaves the test infrastructure, so it's a nice place to write tests since it's all "emulated" in containers instead of hitting a real registry

joshspicer:
I used your changes here to publish to ghcr.io as a final smoke test 🚀 

I pushed these Features:
https://github.com/orgs/codspace/packages?tab=packages&q=non-empty

And consumed the Feature in this test repo:
https://github.com/codspace/non-empty-config-layer-project/blob/main/.devcontainer/devcontainer.json

I don't see any concerns with older CLI backwards compatibility 👍

ych-tnk:
@joshspicer 

I changed the tests to depend on the package you created for the smoke test. If there seems to be a problem, I will replace the test with a local registry on the docker container.

joshspicer:
Given that you're now just wrestling with tests that i've set up, i'm happy to take over the PR to get it over the line :)  Appreciate all the work and communication!

joshspicer (approved):

Issue #829, opened 2024-05-16
Speed up installing dotfiles by loading only latest commit


ai:
Right now CLI uses `git clone` to download dotfiles repo. It will download the whole history of it, which could take a while.

By using `git clone --depth 1` we can speed up process a lot.

PR #830, opened 2024-05-16
Use --depth 1 to make dotfiles install process faster

ai:
Fixes https://github.com/devcontainers/cli/issues/829


samruddhikhandale (approved):
Makes sense to me, thanks! 

Looping in @chrmarti for another set of 👀

chrmarti (approved):
LGTM, thanks!

Issue #826, opened 2024-05-11
Should `devcontainers/cli` uses `docker compose` always instead `docker-compose`?


duduribeiro:
Hey devcontainers/cli folks 👋 

I have a CI that uses `devcontainers/ci` action (which uses the devcontainers/cli) to validate a devcontainer generated for the project.

I'm trying to add the top-level name property into the compose file https://github.com/rails/rails/pull/51791

but it is failing on CI with the error of:
```
he Compose file .. is invalid because:
  'name' does not match any of the regexes: '^x-'
```
https://github.com/rails/rails/actions/runs/9044514158/job/24853256908?pr=51791

I also tried to specify the version 3.8 on the compose but without success. 

I see that `devcontainers/cli` code base uses `docker-compose` instead `docker compose`. should we replace the codebase to use `docker compose` instead so we avoid to use the old compose which is deprecate since jul 23? https://docs.docker.com/compose/migrate/

I also see a test on devcontainers/cli codebase https://github.com/devcontainers/cli/blob/main/src/test/configs/compose-with-name/.devcontainer/docker-compose.yml which tests the exact same scenario but in my case it is not working.

thoughts?

samruddhikhandale:
Hi 👋 

Looking at 👇 , I think the request makes sense for the CLI to switch to the newer Docker compose V2 syntax  (ie `docker compose`) given https://docs.docker.com/compose/migrate/

https://github.com/devcontainers/cli/blob/be1f2034e10395f42843ebf26600a42e11e66531/src/spec-node/devContainers.ts#L161

@gauravsaini04 Can you help take a look at this request? Thanks! 

// cc @chrmarti Let us know if there are any concerns, thanks!

chrmarti:
Sounds good! We currently check for `docker-compose` and if that doesn't exist we use `docker compose`. To minimize breakage, we could turn this around. Similar code exists in the Dev Containers extension which will also need updating to stay consistent.

The check is here: https://github.com/devcontainers/cli/blob/be1f2034e10395f42843ebf26600a42e11e66531/src/spec-node/dockerCompose.ts#L687

gauravsaini04:
pr [#848](https://github.com/devcontainers/cli/pull/848) was merged to the codebase. closing the issue.

PR #848, opened 2024-06-18
cli- docker-compose - swap of installation packages

gauravsaini04:
This pr is a solution to the following issue: 
https://github.com/devcontainers/cli/issues/826

Order of installations has been reversed as proposed as a solution to the issue

### DIFF
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -710,12 +710,12 @@ export function dockerComposeCLIConfig(params: Omit<PartialExecParameters, 'cmd'
 					...params,
 					cmd: dockerCLICmd,
 				}, 'compose', 'version', '--short')).stdout;
-				v2 = true;
 			} catch (err) {
 				stdout = (await dockerComposeCLI({
 					...params,
 					cmd: dockerComposeCLICmd,
 				}, 'version', '--short')).stdout;
+				v2 = false;
 			}
 			const version = stdout.toString().trim();
 			params.output.write(`Docker Compose version: ${version}`);
--- a/src/spec-node/dockerCompose.ts
+++ b/src/spec-node/dockerCompose.ts
@@ -703,30 +703,26 @@ export function dockerComposeCLIConfig(params: Omit<PartialExecParameters, 'cmd'
 	let result: Promise<DockerComposeCLI>;
 	return () => {
 		return result || (result = (async () => {
-			let v2 = false;
+			let v2 = true;
 			let stdout: Buffer;
-			dockerCLICmd='docker-compose';
 			try {
+				stdout = (await dockerComposeCLI({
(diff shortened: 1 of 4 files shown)


src/spec-node/dockerCompose.ts
```diff
@@ -705,6 +705,7 @@ export function dockerComposeCLIConfig(params: Omit<PartialExecParameters, 'cmd'
 		return result || (result = (async () => {
 			let v2 = false;
 			let stdout: Buffer;
+			dockerCLICmd='docker-compose';
```
samruddhikhandale, line 708: `function dockerComposeCLIConfig` already accepts `dockerComposeCLICmd` as parameter Can you instead update all the places where this function is called? eg. https://github.com/devcontainers/cli/blob/58e0a470e19473514af63a0d9ea0ff4a488de3cb/src/spec-node/devContainers.ts#L165C36-L165C53

samruddhikhandale, line 708: Here `dockerCLICmd` should not be set to a docker compose command. Can we revert this?

samruddhikhandale, line 708: Can you fix the failing tests?

samruddhikhandale (changes requested):

src/spec-node/dockerCompose.ts
chrmarti: `dockerComposeCLICmd` is passed in from the `--docker-compose-path` CLI option. I think we need to keep the CLI option specify the Docker Compose v1 path to not break existing setups. The existing `--docker-path` (passed in here as `dockerCLICmd`) implies the Docker Compose v2 path.

I think we need to:
- Swap the two calls to 'version'.
- Start with `v2 = true`.
- Remove the check for `ENOENT`? Since `docker` likely exists, but if v2 is missing then `docker compose` will return an error.

chrmarti (changes requested):
Thanks for addressing this. Left a comment.

src/spec-node/dockerCompose.ts
```diff
@@ -703,22 +703,19 @@ export function dockerComposeCLIConfig(params: Omit<PartialExecParameters, 'cmd'
...
-				}, 'version', '--short')).stdout;
-			} catch (err) {
-				if (err?.code !== 'ENOENT') {
-					throw err;
-				}
 				stdout = (await dockerComposeCLI({
 					...params,
 					cmd: dockerCLICmd,
 				}, 'compose', 'version', '--short')).stdout;
 				v2 = true;
```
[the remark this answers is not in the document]
gauravsaini04: thanks !

chrmarti (approved):
LGTM!

samruddhikhandale (approved):
Nice!

Issue #845, opened 2024-05-14
missing root cause of error


behrica:
I have 
devcontainer snippet like this:
```
"postStartCommand":  
  { 
    "poetry setup":  ["/bin/bash", "-i", "-c",  "python3 -m venv $HOME/.local && source $HOME/.local/bin/activate && poetry install"]
  }
```
and forgot to setup `poetry`, so it is not available.
But the log does not show this, it does not show the root error message in this situation. (`poetry: command not found`)
I took me while to figure this out.

```
2024-05-14T08:35:36.7508615Z Error: Command failed: /bin/bash -i -c mkdir -p /home/vscode/.local && python3 -m venv /home/vscode/.local && source /home/vscode/.local/bin/activate && poetry install
2024-05-14T08:35:36.7511344Z     at N7 (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:235:130)
2024-05-14T08:35:36.7513197Z     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2024-05-14T08:35:36.7515326Z     at async Am (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:227:4393)
2024-05-14T08:35:36.7517736Z     at async k7 (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:227:3910)
2024-05-14T08:35:36.7520153Z     at async $w (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:227:3102)
2024-05-14T08:35:36.7522547Z     at async fa (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:227:2386)
2024-05-14T08:35:36.7524984Z     at async DtA (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:464:1496)
2024-05-14T08:35:36.7660547Z     at async NH (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:464:964)
2024-05-14T08:35:36.7662569Z     at async KtA (/usr/local/lib/node_modules/@devcontainers/cli/dist/spec-node/devContainersSpecCLI.js:481:3692)
[...]

PR #852, opened 2024-07-03
Poetry error message

prathameshzarkar9:
Issue #845 

infoOutput has been updated and logged the message of the command if the command fails inside container

image


src/spec-common/injectHeadless.ts
```diff
@@ -535,12 +535,15 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 			if (err && (err.code === 130 || err.signal === 2)) { // SIGINT seen on darwin as code === 130, would also make sense as signal === 2.
 				infoOutput.raw(`\r\n\x1b[1m${lifecycleHookName} interrupted.\x1b[0m\r\n\r\n`);
 			} else {
+				if (err.code === 127) {
+					infoOutput.write(toErrorText(`${err.cmdOutput}`));
```
samruddhikhandale, line 539: Curious, why don't we always output the `err.cmdOutput` when there exists one ? That way even if the error code is not 127, everyone would benefit.

src/test/configs/dockerfile-with-target/.devcontainer.json
```diff
@@ -8,7 +8,14 @@
 		"options": [ "--label", "test_build_options=success" ]
 	},
 	"postCreateCommand": "touch /tmp/postCreateCommand.testmarker",
-	"postStartCommand": "touch /tmp/postStartCommand.testmarker",
```
samruddhikhandale, line 11: Shouldn't we still keep this `"touch /tmp/postStartCommand.testmarker` ?

src/test/configs/example/.devcontainer.json
```diff
@@ -3,8 +3,16 @@
 {
 	"image": "mcr.microsoft.com/devcontainers/base:latest",
 	"features": {
-		"ghcr.io/devcontainers/features/go:1": {
+		"ghcr.io/devcontainers/features/python:1": {
 			"version": "latest"
 		}
+	},
+	"postStartCommand": {
+		"poetry setup": [
```
samruddhikhandale, line 11: Why did we update this? The tests using this config would fail as python Feature doesn't install poetry.

src/test/configs/dockerfile-with-target/.devcontainer.json
```diff
@@ -8,7 +8,14 @@
 		"options": [ "--label", "test_build_options=success" ]
```
samruddhikhandale, line 8: Can we fix the failing tests?

samruddhikhandale (changes requested):

src/spec-common/injectHeadless.ts
prathameshzarkar9: done

src/test/configs/example/.devcontainer.json
prathameshzarkar9: reverted

src/test/configs/dockerfile-with-target/.devcontainer.json
prathameshzarkar9: reverted

src/test/configs/dockerfile-with-target/.devcontainer.json
prathameshzarkar9: fixed

src/spec-common/injectHeadless.ts
```diff
@@ -535,12 +535,13 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 			if (err && (err.code === 130 || err.signal === 2)) { // SIGINT seen on darwin as code === 130, would also make sense as signal === 2.
 				infoOutput.raw(`\r\n\x1b[1m${lifecycleHookName} interrupted.\x1b[0m\r\n\r\n`);
 			} else {
+				infoOutput.write(toErrorText(`${err.cmdOutput}`));
```
samruddhikhandale, line 538: Should we check if `err.cmdOutput` exists before we output? We should try to avoid null refs.

src/test/cli.up.test.ts
```diff
@@ -25,6 +25,13 @@ describe('Dev Containers CLI', function () {
 	});
 
 	describe('Command up', () => {
+		it('should succeed (dockerfile) with supported --error for poetry', async () => {
```
samruddhikhandale, line 28: What's the purpose of this test with respect to the PR? 

If you want to test the poetry scenario, then you can create a new dev container config which fails similar to https://github.com/devcontainers/cli/issues/845

Then, validate if you the `poetry: command not found` error. However, in this case it's alright to not add the tests.

samruddhikhandale (changes requested):

src/spec-common/injectHeadless.ts
prathameshzarkar9: done

src/test/cli.up.test.ts
prathameshzarkar9: test removed

src/spec-common/injectHeadless.ts
```diff
@@ -535,12 +535,15 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 			if (err && (err.code === 130 || err.signal === 2)) { // SIGINT seen on darwin as code === 130, would also make sense as signal === 2.
 				infoOutput.raw(`\r\n\x1b[1m${lifecycleHookName} interrupted.\x1b[0m\r\n\r\n`);
 			} else {
+				if (err.cmdOutput) {
```
samruddhikhandale, line 538:
```suggestion
				if (err?.cmdOutput) {
```

samruddhikhandale (changes requested):
Let's wait for another review from @chrmarti. Thanks!

src/spec-common/injectHeadless.ts
chrmarti, line 538: Should we move the error handling into `runSingleCommand` above, so it is done for each command? We might have missed that when adding the object syntax for user commands.

`runSingleCommand` uses `const printMode = name ? 'off' : 'continuous';`, so `cmdOutput` should already make it to the terminal when there are no names with the string/string-array syntax. Could you check?

chrmarti (changes requested):
Left a comment, maybe the error handling should move up a few lines. Thanks.

src/spec-common/injectHeadless.ts
prathameshzarkar9: The event is thrown in runRemoteCommand function, so I have added a cmdOutput to the message to show what exactly is failed in the command.
please check the new changes committed to fix this issue.

prathameshzarkar9: moved the cmdOutput to runRemoteCommand function and appended in the message.

chrmarti: I suggest to keep `message` of the error short and single line. Would it instead make sense to put https://github.com/devcontainers/cli/blob/main/src/spec-common/injectHeadless.ts#L506-L517 in a try-catch block and log `cmdOutput` in the catch? Much or all of the outer catch block's code then needs to move to the inner catch: https://github.com/devcontainers/cli/blob/main/src/spec-common/injectHeadless.ts#L533-L548.

I think the 'progress' events should move out of the `runSingleCommand` function. They look like they should only be sent once for all commands of a lifecycle hook. (This is confirmed by code in the Dev Containers extension.)

prathameshzarkar9: Hi @chrmarti ,
I have checked in new changes as per your suggestion. Kindly help to review it and let me know if any change is required will accordingly address the same.

src/spec-common/injectHeadless.ts
```diff
@@ -512,38 +512,41 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
-				if (err?.code) {
-					infoOutput.write(toErrorText(`${lifecycleHookName} failed with exit code ${err.code}. Skipping any further user-provided commands.`));
-				}
-				throw new ContainerError({
-					description: `The ${lifecycleHookName} in the ${userCommandOrigin} failed.`,
-					originalError: err,
-				});
-			}
 		}
+		await Promise.all(commands);
```
chrmarti, line 549: Keep this in a try-catch and move all infoOutput.event() calls here. These events don't seem to be per command but per lifecycle hook.

src/spec-common/injectHeadless.ts
```diff
@@ -512,38 +512,41 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 					name: progressName,
 					status: 'succeeded',
 				});
+			} catch (err) {
+				infoOutput.event({
+					type: 'progress',
+					name: progressName,
+					status: 'failed',
+				});
+				if (err?.cmdOutput) {
```
chrmarti, line 521: Only write the output when `name` is set. Otherwise the output was already written (see `printMode` above).

chrmarti (changes requested):
Thanks @prathameshzarkar9! I have left a few comments.

src/spec-common/injectHeadless.ts
prathameshzarkar9: Done

prathameshzarkar9:
@chrmarti , can you please review the requested changes and approve the same.

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
-				infoOutput.event({
-					type: 'progress',
-					name: progressName,
-					status: 'running',
-					stepDetail: progressDetail
-				});
+		const remoteCwd = containerProperties.remoteWorkspaceFolder || containerProperties.homeFolder;
+		async function runSingleCommand(postCommand: string | string[], name?: string) {
+			const progressDetail = typeof postCommand === 'string' ? postCommand : postCommand.join(' ');
+			infoOutput.event({
```
chrmarti, line 490: Move this event to right before `await Promise.all(commands);`.

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
 					infoOutput.raw(`\x1b[1mRunning ${name} from ${userCommandOrigin}...\x1b[0m\r\n${cmdOutput}\r\n`);
 				}
-
-				infoOutput.event({
-					type: 'progress',
-					name: progressName,
-					status: 'succeeded',
-				});
+			} catch (err) {
+				if (err?.cmdOutput) {
```
chrmarti, line 510: This is still missing an additional check. Maybe:
```suggestion
				if (printMode === 'off' && err?.cmdOutput) {
```

src/spec-common/injectHeadless.ts
```diff
@@ -525,24 +536,18 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 					return runSingleCommand(command, name);
 				});
 			}
+			infoOutput.event({
```
chrmarti, line 539: This event needs to move to right after `await Promise.all(commands);`.

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
-					type: 'progress',
-					name: progressName,
-					status: 'succeeded',
-				});
+			} catch (err) {
+				if (err?.cmdOutput) {
+					infoOutput.raw(`\r\n\x1b[1m${err.cmdOutput}\x1b[0m\r\n\r\n`);
+				}
+				if (err && (err.code === 130 || err.signal === 2)) { // SIGINT seen on darwin as code === 130, would also make sense as signal === 2.
+					infoOutput.raw(`\r\n\x1b[1m${lifecycleHookName} interrupted.\x1b[0m\r\n\r\n`);
```
chrmarti, line 514:
```suggestion
					infoOutput.raw(`\r\n\x1b[1m${name ? `${name} of ${lifecycleHookName}` : lifecycleHookName} from ${userCommandOrigin} interrupted.\x1b[0m\r\n\r\n`);
```

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
+				} else {
+					if (err?.code) {
+						infoOutput.write(toErrorText(`${lifecycleHookName} failed with exit code ${err.code}. Skipping any further user-provided commands.`));
```
chrmarti, line 517:
```suggestion
						infoOutput.write(toErrorText(`${name ? `${name} of ${lifecycleHookName}` : lifecycleHookName} from ${userCommandOrigin} failed with exit code ${err.code}. Skipping any further user-provided commands.`));
```

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
+					}
+					throw new ContainerError({
+						description: `The ${lifecycleHookName} in the ${userCommandOrigin} failed.`,
```
chrmarti, line 520:
```suggestion
						description: `${name ? `${name} of ${lifecycleHookName}` : lifecycleHookName} from ${userCommandOrigin} failed.`,
```

src/spec-common/injectHeadless.ts
```diff
@@ -484,38 +484,49 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
+			// we need to hold output until the command is done so that the output
+			// doesn't get interleaved with the output of other commands.
+			const printMode = name ? 'off' : 'continuous';
+			const env = { ...(await remoteEnv), ...(await secrets) };
+			try {
 				const { cmdOutput } = await runRemoteCommand({ ...lifecycleHook, output: infoOutput }, containerProperties, typeof postCommand === 'string' ? ['/bin/sh', '-c', postCommand] : postCommand, remoteCwd, { remoteEnv: env, pty: true, print: printMode });
 
 				// 'name' is set when parallel execution syntax is used.
 				if (name) {
 					infoOutput.raw(`\x1b[1mRunning ${name} from ${userCommandOrigin}...\x1b[0m\r\n${cmdOutput}\r\n`);
```
chrmarti, line 507:
```suggestion
					infoOutput.raw(`\x1b[1mRunning ${name} of ${lifecycleHookName} from ${userCommandOrigin}...\x1b[0m\r\n${cmdOutput}\r\n`);
```

chrmarti (changes requested):
A few more comments and some message polishing.

src/spec-common/injectHeadless.ts
prathameshzarkar9: Done

src/spec-common/injectHeadless.ts
prathameshzarkar9: Done

src/spec-common/injectHeadless.ts
prathameshzarkar9: Done

src/spec-common/injectHeadless.ts
prathameshzarkar9: done. moved this before `await Promise.all(commands);`

src/spec-common/injectHeadless.ts
prathameshzarkar9: Done.

src/spec-common/injectHeadless.ts
chrmarti: Actually, since `progressDetail` includes the command to run, maybe this should stay here.

@samruddhikhandale Could you advise on this? I think Codespaces is using the progress events in the UI?

@prathameshzarkar9 I don't see your latest changes on GitHub, did you push the commits?

prathameshzarkar9: Hi @chrmarti ,
changes suggested by you are incorporated and the new changes also checked-in here in the latest commits.
PR is now updated can you please review it once @samruddhikhandale ?

samruddhikhandale: > Actually, since progressDetail includes the command to run, maybe this should stay here.
> @samruddhikhandale Could you advise on this? I think Codespaces is using the progress events in the UI?

@chrmarti  Good catch here! 
Yes, Codespaces is still using `progressDetail` so let's keep it as is.

prathameshzarkar9: @samruddhikhandale , @chrmarti 
I kept the progressDetails as it is. can you review it once?

samruddhikhandale:
@prathameshzarkar9 Can you fix the failing tests? Thanks

src/spec-common/injectHeadless.ts
```diff
@@ -528,24 +538,20 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
 				});
 			}
+
 			await Promise.all(commands);
-		} catch (err) {
+
+			infoOutput.event({
+				type: 'progress',
+				name: progressName,
+				status: 'succeeded',
```
samruddhikhandale, line 547: @prathameshzarkar9 Why was this added?

Mathiyarasy:
@chrmarti Could you please review this PR ..

src/spec-common/injectHeadless.ts
```diff
@@ -528,24 +538,14 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 					return runSingleCommand(command, name);
 				});
 			}
+
 			await Promise.all(commands);
-		} catch (err) {
+		} catch {
```
chrmarti, line 543: I think the error needs to be rethrown after reporting progress to preserve the previous behavior. Only if it is rethrown will the "further user-provided commands" be skipped. Please check.

src/spec-common/injectHeadless.ts
```diff
@@ -487,38 +487,48 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
+					if (err?.code) {
+						infoOutput.write(toErrorText(`${name ? `${name} of ${lifecycleHookName}` : lifecycleHookName} from ${userCommandOrigin} failed with exit code ${err.code}. Skipping any further user-provided commands.`));
+					}
+					throw new ContainerError({
+						description: `${name ? `${name} of ${lifecycleHookName}` : lifecycleHookName} from ${userCommandOrigin} failed.`,
+						originalError: err
+					});
 				}
-
-				infoOutput.event({
```
chrmarti, line 513: Success still needs to be reported after `await Promise.all(commands)` finished successfully below. Please move it there.

chrmarti (changes requested):
Left two more comments.

chrmarti (approved):

eljog (approved):

PR #859, opened 2024-07-16
feat: support `forwardPorts` config

### LINKED ISSUE #186: forwardPorts configuration option not working
With the recent release (the one with features) I tried experimenting with the devcontainers CLI. But I just can't get the `forwardPorts` feature to work. I created the following minimal `devcontainer.json` file:
```
{
  "name": "my-project-devcontainer",
  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
  "features": {
    "ghcr.io/devcontainers/features/sshd:1": {
      "version": "latest"
    }
  },
  "forwardPorts": [2222],
  "remoteUser": "vscode"
}
```

Maybe I am misunderstanding the `forwardPorts` option, but I would expect that this configuration will forward the port 2222 (this is the default defined by the sshd feature) from the devcontainer to my host. But a simple `ssh localhost -p 2222` did not work.

I tried this on Windows 11.

agilgur5:
### Summary

Fixes #186 by adding `-p` args for all `forwardPorts`, not just `appPort`

### Details

- split out a `normalizePorts` function used on both `appPort` and `forwardPorts`
- then `concat` and `flatMap` those two together into args

Credits to @MunsMan for the original variant listed in this comment https://github.com/devcontainers/cli/issues/186#issuecomment-1455910147: this commit: https://github.com/MunsMan/devcontainercli/commit/6909283588fc02695f48ca9572515acdeff1d4a8
- I simplified it and fixed some styling etc

### Notes for Reviewers

There didn't seem to be any existing tests for `appPort` that I could extend for `forwardPorts`, afaict. Let me know if there's a proper place to add tests for this


chrmarti:
Thanks for the PR! A few considerations:

The Dev Containers extension for VS Code forwards the listed ports to a free port on the local machine (which might be different from where Docker runs). GitHub Codespaces also has its own port forwarding. Additionally using `-p` might conflict with other containers running on the same Docker host.

Could this be added behind a CLI option, so the above scenarios don't break? `forwardPorts` is also available for dev containers using Docker Compose (unlike `appPort`).

agilgur5:
I would think that all of those scenarios should be handled already downstream, as the behavior is identical to `appPort` and so they'd need to handle it already, no?

> Could this be added behind a CLI option, so the above scenarios don't break?

For a reference implementation, that seems pretty confusing to me. The spec is the source of truth, yet you also need a flag?
An env var to disable might make more sense, although, per above, I'm not sure that would really be necessary.

chrmarti:
The extension uses VS Code's port forwarding which is temporary while VS Code is connected and also works when the VS Code window runs on a different machine than the Docker daemon. (It is not use Docker's port forwarding.)

I think another approach would be to replicate the temporary forwarding as part of an additional CLI command which would keep running until the port forwarding is no longer required and only then would be terminated.

agilgur5:
> The extension uses VS Code's port forwarding

Then it should already be ignoring this, no? As it already ignores `appPort` for the same reason

> I think another approach would be to replicate the temporary forwarding as part of an additional CLI command

While this makes sense to me, this doesn't match the existing behavior of `appPort`. As far as I could tell, `forwardPorts` is partly an array replacement for it, since [the spec](https://github.com/devcontainers/spec/blob/3174ebbb25377e52b3637889772098da01d957dc/docs/specs/devcontainerjson-reference.md#publishing-vs-forwarding-ports) says: 
> In most cases, we recommend using the new [forwardPorts property](https://github.com/devcontainers/spec/blob/3174ebbb25377e52b3637889772098da01d957dc/docs/specs/devcontainerjson-reference.md#general-devcontainerjson-properties).

> The property is most useful for forwarding ports that cannot be auto-forwarded 

But I don't think I realized `appPort` could be an array as well, that workaround might suffice for my case at least. I feel like I tried that and it didn't work for some reason though 🤔 (I do use the same spec for an extension and Codespaces when necessary, but tend to prefer the more minimal CLI)

I had also looked at https://github.com/microsoft/vscode-remote-release/issues/1809, where you did say (almost 5 years ago):

> I agree, we can't make published ports disappear, they are part of the basic functionality of Docker.

and someone else said a bit further up in the thread:

> and the fact that it's done with docker publishing the port seems like an implementation detail. Instead of surfacing a confusing implementation detail (publish vs. forward)

So I assumed they're equivalent in a Docker-based implementation. I'm not sure if that's changed since then or that's specific to the VS Code extension, but since the devcontainer CLI is similarly Docker-based, I would think that applies the same?

chrmarti:
Just to clarify: The Dev Containers extension for VS Code and GitHub Codespaces both use the Dev Containers CLI to create the dev container. We moved away (`appPort` > `forwardPorts`) from Docker-based forwarding ("publishing") because these can't be added to an already running container and they only work locally, not when the dev container runs on a remote machine connected via SSH.

agilgur5:
A few questions for clarity:

> both use the Dev Containers CLI to create the dev container.

Gotcha. And then they have their own forwarding implementation, right?

How do they handle the existing `appPort` publishing right now though? That is throwing me for a loop, since I just reused tha existing behavior in the PR

> because these can't be added to an already running container

Yea that was my other concern -- with a separate command for this, it would be done while already running, meaning we can't use the Docker implementation and have to reimplement forwarding (which is a bit of a rabbit hole). Does it make sense to move/share the downstream implementations here?

almathie:
Is there any progress on this ? On our side we are trying to leverage devcontainers for a pretty standard Rails app. The required ports are exposed through the `forwardPorts` directive as recommended here: https://containers.dev/implementors/json_reference/

Everything works fine for VSCode / Codespace (aka:Microsoft lead products) users but developers using vim/emcas or similar can't properly spin up a working devcontainer from their terminal.

It feels weird to me that the documentation recommends an approach that is currently unsupported by the reference implementation. Additionally, some tools and services are apparently leveraging the cli behind the scene, and they are thus subject to the same limitations (we tried).

At the moment it feels like using VSCode as the "devcontainer manager" is required to have `forwardPorts` working at all which goes a bit against again the idea of having devcontainers as an open specification.

agilgur5:
### reference implementation

> It feels weird to me that the documentation recommends an approach that is currently unsupported by the reference implementation

> At the moment it feels like using VSCode as the "devcontainer manager" is required to have `forwardPorts` working at all which goes a bit against again the idea of having devcontainers as an open specification.

Yea, I generally agree with both these statements per my above comments

### summary

> Is there any progress on this ?

Let me summarize the comments to address this (and refresh my memory and make sure everyone's on the same page):

> The Dev Containers extension for VS Code and GitHub Codespaces both use the Dev Containers CLI to create the dev container.

> We left `appPort` as-is at the time to not break anyone and started to recommend users use `forwardPorts` instead.

> That code assumes Node.js is available in the container - which is true when using VS Code and the VS Code Server. (The extension is not OSS.)

As I understand, this basically means that the `forwardPorts` implementations in GH Codespaces and the VS Code extension are not compatible with the Docker publishing of this PR. And as they do use this CLI under-the-hood, that is why it was recommended to put these behind a flag or separate command.

A separate command would be incompatible with Docker publishing and so requires re-implementing forwarding (which is a bit of rabbit hole and may require various dependencies on the host _and_ container).

Both a separate command or behind a flag would make the reference implementation not quite correspond to the spec by default though.

### next steps

As such, I would go back to my earlier proposal of an env var (or flag) to _disable_ `forwardPorts`. The default then follows the spec and other implementations, like GH Codespaces and the VS Code extension, can disable this if they simultaneously use the CLI and also have alternate implementations.

That might seem like a breaking change from the GH Codespaces / VS Code extension perspective, but it is a feature.

@chrmarti would that proposal work?

I'm willing to re-write it as an opt-in flag instead at this point just to get the behavior in the CLI, but that would still be odd as a reference implementation.
And of course, you can release as a breaking change either way if desired to require downstream consumers to be aware of the change.

agilgur5:
I added a `--skip-forward-ports` flag in https://github.com/devcontainers/cli/pull/859/commits/ae893d9bd934b39019af17eae98e8f6440d68d03.

The function arg drilling is not the most ideal, but changing that would require a lot of refactoring that I don't think I'm best suited to handle at this time.

apriendeau:
@chrmarti is there anyway that we can review this? This blocks all editors that are not VS Code from developing.

agilgur5:
following up @chrmarti

chrmarti:
Publishing the ports by default would break existing setups. Could you invert the logic and make it `--publish-forward-ports` or something?

agilgur5:
> Could you invert the logic and make it `--publish-forward-ports` or something?

As I wrote [above](https://github.com/devcontainers/cli/pull/859#issuecomment-2596866818), this would not be _spec compliant_ and therefore especially confusing for a _reference implementation_. If you really want me to I can, as this _blocks_ lots of users and tools as others wrote [above](https://github.com/devcontainers/cli/pull/859#issuecomment-2669614124), but I am quite strongly against that option due to its lack of spec compliance.

> Publishing the ports by default would break existing setups.

I think it's important to qualify this statement -- as I and others wrote above and in the issue, many existing setups are _already broken_ because this feature doesn't exist in the reference implementation.
Afaik, it would only break VS Code and GH Codespaces, and that's only if they were to update and not add `--skip-forward-ports`. It would only break those two because they have their own custom implementation of `forwardPorts` that _differs from the reference_. Nominally and conventionally speaking, it would be the responsibility of those tools to disable reference features if they implement them differently on their own.

As I wrote above, if you really want to, you can certainly release this feature as a breaking change with a major version bump, thereby requiring downstream users to acknowledge the breakage and adjust appropriately.
As I wrote, this isn't strictly breaking in the sense that it's adding a missing feature, but could be considered so in the sense that it changes default behavior (to be spec compliant).

adhxtry:
Hey @agilgur5 
I think `forwardPorts` is not exactly same as `appPorts`

https://containers.dev/implementors/json_reference/#publishing-vs-forwarding-ports

`docker -p` exposes a container port to the network, while `forwardPorts` tunnels traffic so it appears as localhost inside the container.

agilgur5:
@adhxtry I'm aware they're not the same; that is covered in this thread. See the above comments such as [my summary](https://github.com/devcontainers/cli/pull/859#issuecomment-2596866818)

adhxtry:
@agilgur5 

This PR is using `-p` for `forwardPorts` as well as `appPort`.

When you use `forwardPorts` with say `8000` in VS Code, something like:

```
// Node / Express example
app.listen(8000, "localhost")
```

Should work inside the dev container.
However, it won't work for published ports, you have to use `"0.0.0.0"` as IP.

That is why adding `--publish-forward-ports` flag for this functionality makes sense. As it doesn't do what spec requires it to do. Instead it publishes the ports.

I have prepared a small demo using LLM, for how the `forwardPorts` should actually work, with tunnelling using `socat`.

[forwardPort_demo.zip](https://github.com/user-attachments/files/24356226/forwardPort_demo.zip)

PR #865, opened 2024-07-31
publish Templates with metadata attached to manifest

joshspicer:
Just like with Features, attach a complete copy of the metadata (the `devcontainer-template.json`) to the OCI manifest.  This is useful for tooling to query.

This will be immediately useful for https://github.com/microsoft/vscode-remote-release/issues/10095, allowing the extension to read various metadata properties before applying the Template.

Example Template with the metadata attached can be seen here (displayed under 'labels' on GitHub): https://github.com/codspace/templates/pkgs/container/templates%2Fmytemplate/251579251?tag=latest

PR #866, opened 2024-08-01
add 'templates metadata' subcommand

joshspicer:
Follow up to https://github.com/devcontainers/cli/pull/865.  Exposes Template metadata cached on a manifest with the `templates metadata` subcommand

Also relevant for https://github.com/microsoft/vscode-remote-release/issues/10095, by allowing tools to query the metadata of a Template before downloading its contents.

image


samruddhikhandale (dismissed):
Looks great, should we add some test coverage?

samruddhikhandale (approved):
⚡

PR #870, opened 2024-08-12
Fix registry prompting when running with `podman`

AndesKrrrrrrrrrrr:
Issue https://github.com/containers/podman/issues/23055 outlines that `podman` doesn't search the local repository for the dynamically built image. This could be fixed by prefixing the dynamically generated image with `localhost/`.

Container images should be fully qualified at all times, anyways (see e.g. [this RedHat blog](https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name)).


chrmarti:
This should already be handled here: https://github.com/devcontainers/cli/blob/fbd8dfe91ec239bb1ca8d6a82bafe45887f0fab8/src/spec-node/containerFeatures.ts#L434

Maybe the check for Podman didn't work. What output do you get from `podman -v`?

zvictor:
>  What output do you get from podman -v?

I tried in both `podman version 5.0.3` and `podman version 5.3.1`, and I have the issue described.

typedrat:
Could the issue be that it is checking for `docker` first? A lot of the time, `podman` is installed with a `docker` wrapper that lies about its identity.

```console
❯ podman -v
podman version 5.4.1
❯ docker -v
docker version 5.4.1
```

PR #875, opened 2024-08-21
Enhance Template metadata

joshspicer:
Related: https://github.com/devcontainers/spec/pull/484

- Updates Template interface
- Includes array of file names on published Template metadata (potentially useful for consuming tools)
- When an `optionalPaths` directory entry implies only a single file is optional, automatically resolve that full file name on the metadata.  Useful for consuming tools to better highlight what file is optional
- Fixed code that appends `featureIds` to Template metadata (now uses `getRef(..)` function to cover edge cases)
- Add new test case to validate the above


samruddhikhandale (approved):
Looks good!

chrmarti (approved):

joshspicer:
Validated that the `optionalPaths` single-file resolution is working as-expected e2e:
image

PR #882, opened 2024-08-29
Allow variable substitution default value containing a colon

grthr:
The current implementation does not support default values in localEnv or containerEnv substitution that include a colon.

Example:
```
 "WEBSITE_URL": "${localEnv:WEBSITE_URL:https://example.com"
```

If `$WEBSITE_URL` is not present, the variable is just set to `https`

### DIFF
--- a/src/test/variableSubstitution.test.ts
+++ b/src/test/variableSubstitution.test.ts
@@ -88,6 +88,21 @@ describe('Variable substitution', function () {
 		assert.strictEqual(result.foo, 'bardefaultbar');
 	});
 
+	it(`environment variables with default value containing colon if they do not exist`, async () => {
+		const raw = {
+			foo: 'bar${localEnv:baz:de:fault}bar'
+		};
+		const result = substitute({
+			platform: 'linux',
+			localWorkspaceFolder: '/foo/bar',
+			containerWorkspaceFolder: '/baz/blue',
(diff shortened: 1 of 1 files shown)


src/spec-common/variableSubstitution.ts
```diff
@@ -82,7 +82,7 @@ function evaluateSingleVariable(replace: Replace, match: string, variable: strin
 
 	// try to separate variable arguments from variable name
 	let args: string[] = [];
-	const parts = variable.split(':');
+	const parts = variable.split(':', 3);
```
chrmarti, line 85: This would prevent us from introducing additional arguments in the future because we would break existing configs.

Also: `split()` with the limit argument does not include the remaining part of the text after the limit was hit. This would need a closer look.

chrmarti (changes requested):
Thanks for the PR. Left a comment with my thoughts.

src/spec-common/variableSubstitution.ts
grthr: If we would need additional arguments in the future _after_ the current ones it would be nice to be able to escape the `:` in the default value. Would this be an approach to go for?

something like ` "WEBSITE_URL": "${localEnv:WEBSITE_URL:https\://example.com"`

PR #884, opened 2024-08-30
Add more leniency towards registries that malform WWW-Authenticate.

EyeDevelop:
Some OCI registries return a mangled WWW-Authenticate header. For example, versions of Sonatype Nexus Repository Manager return:

```http
WWW-Authenticate: BASIC realm="..."
```

The current handling of the WWW-Authenticate header leaves the CLI to error when pushing to these registries. This commit adds more leniency and accepts weirdly capitalised versions of the header also.


chrmarti (approved):
LGTM, thanks!

samruddhikhandale (approved):
Looks good, thanks!

PR #896, opened 2024-09-15
fix: change increment syntax in test library script

koralowiec:
Function `checkMultiple` which is available in `dev-container-features-test-lib` (test library script) doesn't work as intended

It exits while looping through the 1st command. For example, execution of this script:

```bash
#!/bin/bash

set -e

# Optional: Import test library
source dev-container-features-test-lib

checkMultiple "Test multiple" 2 "echo First" "echo Second"
```
Results in error: 
Pasted image 20240915161640

It seems that it's caused by the way how the `PASSED` variable is incremented: `((PASSED++))` - The value is returned first, then `PASSED` is incremented

For the first iteration the value of the expression is 0, so the return status is set to 1

> (( expression ))
> If the value of the expression is non-zero, the return status is 0; otherwise the return status is 1.
> https://www.gnu.org/software/bash/manual/bash.html#Conditional-Constructs

I changed it to `((PASSED+=1))`, and tested locally. It fixes the issue:

image

### DIFF
--- a/src/spec-node/featuresCLI/utils.ts
+++ b/src/spec-node/featuresCLI/utils.ts
@@ -98,7 +98,7 @@ checkMultiple() {
     shift; MINIMUMPASSED=$1
     shift; EXPRESSION="$1"
     while [ "$EXPRESSION" != "" ]; do
-        if $EXPRESSION; then ((PASSED++)); fi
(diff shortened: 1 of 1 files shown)


chrmarti (approved):
LGTM, thanks!

samruddhikhandale (approved):
Thanks!

PR #929, opened 2024-11-07
Added test for labels with dockerfile

Roemer:
This PR adds a test to add labels via cli to a devcontainer that is based on a Dockerfile. The test currently fails.


johnitvn:
I have encountered this error in practice. Hope this error will be fixed soon.

Roemer:
I actually gave it a try and implemented it in this PR, so it is ready for review.

PR #933, opened 2024-11-18
issue missing root cause of error #845 solution

gauravsaini04:
Ref: #845

**Description**
* As a correction to earlier raised pr #852 
* As a solution to the above issue, pr achieves printing of the correct error message in the above scenario

**_changelog_**
* Added changes to the `src/spec-common/injectHeadless.ts` file as well as a launch.json configuration for cli debugging at `src/test/configs/poetry-example/.devcontainer.json`

**checklist**
* Checked that applied changes work as expected

### DIFF
--- a/src/spec-common/injectHeadless.ts
+++ b/src/spec-common/injectHeadless.ts
@@ -503,18 +503,26 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
 				// doesn't get interleaved with the output of other commands.
 				const printMode = name ? 'off' : 'continuous';
 				const env = { ...(await remoteEnv), ...(await secrets) };
-				const { cmdOutput } = await runRemoteCommand({ ...lifecycleHook, output: infoOutput }, containerProperties, typeof postCommand === 'string' ? ['/bin/sh', '-c', postCommand] : postCommand, remoteCwd, { remoteEnv: env, pty: true, print: printMode });
 
-				// 'name' is set when parallel execution syntax is used.
-				if (name) {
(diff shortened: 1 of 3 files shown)


chrmarti:
@gauravsaini04 It looks like you started from a different change than the very latest in https://github.com/devcontainers/cli/pull/852. Please make sure to pull in the latest from there and address my latest comment on top. Thanks!

src/spec-common/injectHeadless.ts
```diff
@@ -503,18 +503,26 @@ async function runLifecycleCommand({ lifecycleHook }: ResolverParameters, contai
...
-				infoOutput.event({
-					type: 'progress',
-					name: progressName,
-					status: 'succeeded',
-				});
+					// 'name' is set when parallel execution syntax is used.
+					if (name) {
+						infoOutput.raw(`\x1b[1mRunning ${name} of ${lifecycleHookName} from ${userCommandOrigin}...\x1b[0m\r\n${cmdOutput}\r\n`);
+					}
+				} catch (err) {
```
gauravsaini04, line 514: so basically what i understood is that you'd like me to push this code of catching the error inside `runSingleCommand` method to be after https://github.com/devcontainers/cli/blob/main/src/spec-common/injectHeadless.ts#L531 i.e. await Promise.all(commands), in the catch block that's written after it

chrmarti: I think we are very close in https://github.com/devcontainers/cli/pull/852. I suggest we continue there since that has all the history of the discussion. I have pushed the suggested fixes there, please take a look and leave a review. Thanks!

gauravsaini04: closing this as #852 was already merged.

PR #950, opened 2025-01-30
fix: add check for missing FROM instructions in Dockerfile parsing

krewi1:
This fix ensures that the Dockerfile parser checks for the presence of at least one FROM instruction. If no FROM instructions are found, an error is thrown with a clear message, guiding the user to fix the issue by adding a valid FROM statement. This improves robustness and helps users quickly identify and resolve misconfigured or incomplete Dockerfiles.


krewi1:
@microsoft-github-policy-service agree company="Customink"

chrmarti (approved):
Thanks!

eljog (approved):

felipecrs:
Unfortunately this PR actually prevents a legit use case https://github.com/devcontainers/cli/issues/991.

IMO using `docker build --check` would make better sense than implementing its own checker.

PR #959, opened 2025-02-16
Make dev-container-features-test-lib script POSIX compliant (fixes #438)

### LINKED ISSUE #438: Exec'ing test scripts assumes `/bin/bash`
ref: https://github.com/devcontainers/feature-starter/issues/31

garymacindoe:
I've rewritten the `dev-container-features-test-lib` script included in the `devcontainer` CLI to be POSIX-compliant (using [shellcheck](https://www.shellcheck.net/)).

This fixes tests against images where the default shell is not Bash (such as in #438).

### DIFF
--- a/src/spec-node/featuresCLI/utils.ts
+++ b/src/spec-node/featuresCLI/utils.ts
@@ -57,35 +57,27 @@ export interface LaunchResult {
 
 // dev-container-features-test-lib
 export const testLibraryScript = `
-#!/bin/bash
-SCRIPT_FOLDER="$(cd "$(dirname $0)" && pwd)"
-USERNAME=\${1:-root}
+#!/bin/sh
+
+set -eu
 
 if [ -z $HOME ]; then
     HOME="/root"
 fi
 
-FAILED=()
-
-echoStderr()
-{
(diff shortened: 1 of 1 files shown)


garymacindoe:

garymacindoe:
I have forked the devcontainers/features repository and fixed most of the tests as they fail on the default branch with the current version of the devcontainers-cli.  The latest test run is https://github.com/garymacindoe/features/actions/runs/13595482653.

The main things that were failing were installing node LTS then checking that node 20 is installed when node 22 is the current LTS.  As the LTS version of node is hardcoded in the test script it will start to fail again when a new version of node enters LTS.

There were a couple of missing dependencies to successfully build git from source.

The Python scenarios still fail for me in GitHub Actions because I don't have access to a big enough runner but pass for me locally.

The dotnet LTS install also fails because the install script correctly installs version 8 which is the current LTS but the test script queries https://builds.dotnet.microsoft.com/dotnet/Sdk/LTS/latest.version which currently returns `9.0.200`.  I haven't been able to fix that as I'm not sure of the correct API to use.

PR #1042, opened 2025-07-01
Fix: Honor COMPOSE_PROFILE environment variable

varkrish:
This commit addresses an issue where the `devcontainer exec` command would always override the `COMPOSE_PROFILE` environment variable by defaulting to `--profile *` when reading the Docker Compose configuration.

The `readDockerComposeConfig` function in `src/spec-node/dockerCompose.ts` has been updated to check for the presence of the `COMPOSE_PROFILE` environment variable. If `COMPOSE_PROFILE` is set, the `--profile *` argument will not be added, allowing users to specify their desired profiles through the environment variable.

Additionally, new tests have been added to `src/test/dockerComposeUtils.test.ts` to verify this behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8732 @@
+{
+	"name": "@devcontainers/cli",
+	"version": "0.78.0",
+	"lockfileVersion": 3,
+	"requires": true,
+	"packages": {
+		"": {
+			"name": "@devcontainers/cli",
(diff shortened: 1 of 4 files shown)


varkrish:

varun krishnamurthy

SebTM:
What besides rebase is needed to get this merged? Would someone be willing to do the review if author or me rebase it?

PR #1063, opened 2025-08-20
fix missing manifest when performing multi-platform build and push

### LINKED ISSUE #1062: Manifest missing when building and pushing multi-platform images
The following build command results in images lacking manifests.

However, the cache image does have a manifest.
```shell
$ devcontainer build \
--workspace-folder . \
--cache-from registry.gitlab.com/jshbrntt/devcontainers/base/cache:latest \
--cache-to registry.gitlab.com/jshbrntt/devcontainers/base/cache:latest \
--platform linux/amd64 \
--platform linux/arm64 \
--image-name registry.gitlab.com/jshbrntt/devcontainers/base:latest
```

jshbrntt:
Fixes #1062.

### DIFF
--- a/src/spec-node/devContainersSpecCLI.ts
+++ b/src/spec-node/devContainersSpecCLI.ts
@@ -647,6 +647,17 @@ async function doBuild({
 		// Support multiple use of `--label`
 		params.additionalLabels = (buildxLabel && (Array.isArray(buildxLabel) ? buildxLabel : [buildxLabel]) as string[]) || [];
 
+		// Detect multiple platforms for buildx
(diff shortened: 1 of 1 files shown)


chrmarti:
Thanks for the PR! I think going from `--load` to automatic `--push` as a side-effect of adding another platform might be unexpected to users.

jshbrntt:
@chrmarti how else would you suggest we fix the linked issue?

Doesn't `--load` and `--push` effectively result in the same operation?

I think the images built will already automatically loaded into the local `dockerd` so `--load` is kind of redundant?

Issue #275, opened 2022-11-09
`findBaseImage` in dockerfileUtils.ts does not handle buildkit automatic variables in layers name


sylver:
Giving a simple example `Dockerfile` like this leveraging the `TARGETARCH` [automatic variable](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) provided in the global scope by buildkit :

```dockerfile
ARG APT_FLAGS="-y -qq --no-install-recommends --no-install-suggests"

FROM node AS base-arm64
ONBUILD ARG PACKAGES="\
  build-essential \
  python3 \
"

FROM node AS base-amd64
ONBUILD ARG PACKAGES=""

FROM base-${TARGETARCH} AS base

 apt-get install ${APT_FLAGS} ${PACKAGES}

# do common stuff
```

The `TARGETARCH` global scope variable (Nor for that matter any others) does not seem available for replacement when searching the base image in the `findBaseImage` function called by `internalGetImageBuildInfoFromDockerfile` during the build
https://github.com/devcontainers/cli/blob/f3be29d61167bef3ca430bf106ea2fc203ae0f3b/src/spec-node/dockerfileUtils.ts#L95

in our case ending up returning the value `base-` which does not exist and then fails:

```
[1197 ms] Start: Run: docker inspect --type image base-
[3802 ms] Error fetching image details: No manifest found for docker.io/library/base-.
[3802 ms] Start: Run: docker pull base-
invalid reference format
[3903 ms] []
[3903 ms] Error response from daemon: no such image: base-: invalid reference format
```

Also, another issue I found along the way when inspecting available values, parsing does not handle quoted values (nor multilines). Based on the previous `Dockerfile` example, we end up in the provided replacements to [`replaceVariables` function](https://github.com/devcontainers/cli/blob/f3be29d61167bef3ca430bf106ea2fc203ae0f3b/src/spec-node/dockerfileUtils.ts#L141) with
```js
  // [...]
  { instruction: 'ARG', name: 'APT_FLAGS', value: '"-y' },
  { instruction: 'ARG', name: 'PACKAGES', value: '"\\' },
  // [...]
```

sylver:
@chrmarti Additionally, trying to find a workaround in the meantime, I just noticed that parsing of variables with [replacements](https://docs.docker.com/engine/reference/builder/#environment-replacement) (ie. `${FOO:-bar}`) does not work either.

With something like:

```dockerfile
[...]

PR #1064, opened 2025-08-31
fix: partial support multi-platform build with automatic platform args

tky2240:
## Summary

This PR fix missing platform arguments (e.g., `$TARGETARCH`) for multi-platform builds. 

## Key Changes

- **Fixed Argument Substitution:** Platform arguments are now resolved within the `findBaseImage` function based on the CLI  execution environment or `--platform` args.
- **Added Base Image Validation:** The build will now fail with a error if the base image resolves to a different across the  platforms.

## How to Test

- ✅ Confirm a multi-platform build succeeds when using a base image with a platform variable (e.g., `FROM ${TARGETARCH}-base`).
    - `src/test/configs/dockerfile-with-automatic-platform-args/`
- ❌ Verify that a build fails as expected when using base images that are inconsistent across platforms.
    - `src/test/configs/dockerfile-with-inconsistent-base-image/`

## Related Issus

- https://github.com/microsoft/vscode-remote-release/issues/7582
- https://github.com/microsoft/vscode-remote-release/issues/9605
- https://github.com/devcontainers/cli/issues/275

## Note

- This PR is out of scope for Docker Compose.

### DIFF
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -423,8 +423,15 @@ export async function internalGetImageBuildInfoFromDockerfile(inspectDockerImage
 		delete dockerfile.preamble.directives.syntax;
 	}
 	const images: string[] = [];
-	for (const platform of buildxPlatforms) {
-		const image = findBaseImage(dockerfile, dockerBuildArgs, targetStage, platform);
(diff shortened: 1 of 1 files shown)


tky2240:

PR #1111, opened 2025-11-28
Implement lookupHomeFolder for localHomeFolder case

trajano:
Add lookupHomeFolder function to retrieve home directory.

https://github.com/devcontainers/spec/issues/335


kapouer:
That would have been awesome.

kapouer:
@trajano I think you're supposed to rebase on main ?

Issue #63, opened 2022-06-26
Simplify install process


Chuxel:
Currently this CLI has a native module dependency which drives the need for node-gyp dependencies like Python and C++. This makes installing the CLI far more difficult than intended.

To solve this issue, I'd propose we have a "distro" that is already compiled, and could even bundle Node.js.  Installation can then just be an unzip/untar to keep things easy to begin with.  We could then keep the npm option around for those that prefer to deal with the dependenceis.

The alternative would be to look into ways to remove the native dependency so that only Node.js is required.

//cc: @chrmarti @bamurtaugh

eddiegroves:
This would be great for the use case of using `devcontainer build` in CI/CD pipelines. You could provide a docker image too.

mscottford:
I thought I would share my use case. I typically don't install node directly on my host operating system, instead, I create development-focused containers that have node installed. Same with Python. When I need to work with/on a python project, I create a container to house that tool.

I still wanted to install this cli, though, so I used `homebrew` to get things up and running:
```
brew install node
brew install python
npm install -g @devcontainers/cli
```

aaronpowell:
Another justification for why this would be good is that one of the values of using a devcontainer is that you minimize the locally installed software, instead, relying on the devcontainer to have the stuff you need for actual development.

Having a hard dependency in which you need to install things like Python means that you now have to have more "stuff" installed on the host machine.

avelino:
why not leave ready-made formulas for each OS?
- homebrew
- `.deb`
- `.rpm`
- ...

MisterTimn:
An image with devcontainers/cli installed along with its dependencies would also go a long way.

eitsupi:
Note that it can be installed simply by running `npm install -g @devcontainers/cli` on `mcr.microsoft.com/vscode/devcontainers/typescript-node`.
This is actually used in https://github.com/devcontainers/features/blob/82960fd16b04fb4003a8765b2381bd23553737b5/.devcontainer.json, for example.

framctr:
I agree with @avelino for having `deb` and `rpm` packages.

chrmarti:
There is now an install script for macOS and Linux that will download and install Node.js and the Dev Containers CLI:
```
curl -fsSL https://raw.githubusercontent.com/devcontainers/cli/main/scripts/install.sh | sh
```

TomasEkeli:
[...]

PR #1142, opened 2026-02-06
Add install script

chrmarti:
https://github.com/devcontainers/cli/issues/63


scripts/install.sh
```diff
@@ -0,0 +1,672 @@
+#!/bin/sh
+# install.sh - Install @devcontainers/cli with bundled Node.js
+#
+# Usage:
+#   curl -fsSL https://raw.githubusercontent.com/devcontainers/cli/main/scripts/install.sh | sh
+#   wget -qO- https://raw.githubusercontent.com/devcontainers/cli/main/scripts/install.sh | sh
+#
+# Options:
+#   --prefix DIR        Installation directory (default: ~/.devcontainers)
+#   --version VER       CLI version to install (default: latest)
```
abdurriq: I'm aware that in the context "CLI" is obvious, but it would be better to be explict about it being the "DevContainers CLI" throughout this script and elsewhere.

PR #1177, opened 2026-03-18
refactor: replace substr with slice and remove dead code in singleContainer.ts

ishwar170695:
What:
Replace `.substr()` with `.slice()` in `findUserArg` (singleContainer.ts L293)
Remove commented-out dead code block for `collapsedFeaturesConfig`

Reason: `.substr() is a legacy method flagged by MDN. .slice() is the modern equivalent with identical semantics here. The dead code block is unreachable and no longer referenced anywhere in the codebase.`


chrmarti (approved):
Thanks!

PR #1199, opened 2026-04-09
fix: always write devcontainer.metadata label as JSON array

### LINKED ISSUE #1054: Devcontainer.Metadata Label can be a Dictionary or List
When a devcontainer is built off an image without a devcontainer.metadata label the resulting image will have a devcontainer.metadata label that is a dictionary instead of a list(assuming there aren't any features that also append to the label). For example when I build my base image I get a config with the following labels:
```
  "Labels": {
        "devcontainer.metadata": "{\"remoteUser\":\"fedora\"}",
        "io.buildah.version": "1.40.1",
        "license": "MIT",
        "name": "fedora",
        "org.opencontainers.image.license": "MIT",
        "org.opencontainers.image.name": "fedora",
        "org.opencontainers.image.url": "https://fedoraproject.org/",
        "org.opencontainers.image.vendor": "Fedora Project",
        "org.opencontainers.image.version": "42",
        "vendor": "Fedora Project",
        "version": "42"
    },
```

After creating another image using this image as a base I get the following labels:
```
    "Labels": {
        "devcontainer.metadata": "[ {\"remoteUser\":\"fedora\"}, {\"remoteUser\":\"fedora\"} ]",
        "io.buildah.version": "1.40.1",
        "license": "MIT",
        "name": "fedora",
        "org.opencontainers.image.license": "MIT",
        "org.opencontainers.image.name": "fedora",
        "org.opencontainers.image.url": "https://fedoraproject.org/",
        "org.opencontainers.image.vendor": "Fedora Project",
        "org.opencontainers.image.version": "42",
        "vendor": "Fedora Project",
        "version": "42"
    },
```

Adding features to an image will have the same effect and create the metadata as a list. On the [devcontainer website](https://containers.dev/implementors/json_reference/) it says the devcontainer.metadata label "can contain an array of json snippets". For the devcontainer cli tool this doesn't seem to be an issue but I am using envbuilder to run my containers and if the metadata isn't a list it will cause envbuilder to crash. Based off the website I think it's fair to assume that it should always be a list but it isn't defined very well.

fechu:
The `getDevcontainerMetadataLabel` function wrote a bare JSON object when there was only one metadata entry (e.g. docker-compose devcontainer with no features), but an array when there were multiple entries
- The [spec](https://containers.dev/implementors/json_reference/) states the label "can contain an **array** of json snippets"
- Tools like [Zed](https://github.com/zed-industries/zed) ([parsing code](https://github.com/zed-industries/zed/blob/b6562e8afadaeb73828dfd7dfa94fe8cf8876e18/crates/dev_container/src/docker.rs#L488)) or [envbuilder](https://github.com/coder/envbuilder) that follow the spec and expect an array fail 

The reading side (`internalGetImageMetadata0`) already handles both formats, but the writing side was inconsistent. In this PR we propose to always write an array of objects for simplicity. For backwards compatibility we leave the parsing, which supports both. Changing this makes the code follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) by being strict what is returned (always an array). 

## Test instructions

Create the following project: 
```
/tmp/zed-devcontainer-repro
└── .devcontainer
    ├── devcontainer.json
    ├── docker-compose.yml
    └── Dockerfile
```

devcontainer.json:
```
{
  "name": "repro",
  "dockerComposeFile": "docker-compose.yml",
  "service": "app",
  "remoteUser": "root"
}
```
docker-compose.yml:
```
services:
  app:
    build:
      context: .
      dockerfile: Dockerfile
    command: sleep infinity
    volumes:
      - ..:/workspace
```
Dockerfile
```
FROM ubuntu:24.04
```

With the latest release the metadata is: 
```
> devcontainer up --workspace-folder /tmp/zed-devcontainer-repro
[...]
> docker inspect --format='{{index .Config.Labels "devcontainer.metadata"}}' \
    $(docker ps -q --filter "label=devcontainer.local_folder=/tmp/zed-devcontainer-repro")

{"remoteUser":"root"}
```

After this change: 

```
[{"remoteUser":"root"}]
```

Fixes #1054

### DIFF
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -498,8 +498,8 @@ export function getDevcontainerMetadataLabel(devContainerMetadata: SubstitutedCo
 		return '';
 	}
 	const imageMetadataLabelValue = `[${metadata
-		.map(feature => ` \\\n${toLabelString(feature)}`)
-		.join(',')} \\\n]`;
+		.map(feature => `\\\n${toLabelString(feature)}`)
+		.join(',')}\\\n]`;
 	return `LABEL ${imageMetadataLabel}="${imageMetadataLabelValue}"`;
 }
 
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -497,11 +497,9 @@ export function getDevcontainerMetadataLabel(devContainerMetadata: SubstitutedCo
 	if (!metadata.length) {
 		return '';
 	}
-	const imageMetadataLabelValue = metadata.length !== 1
-		? `[${metadata
-			.map(feature => ` \\\n${toLabelString(feature)}`)
-			.join(',')} \\\n]`
-		: toLabelString(metadata[0]);
+	const imageMetadataLabelValue = `[${metadata
+		.map(feature => ` \\\n${toLabelString(feature)}`)
+		.join(',')} \\\n]`;
 	return `LABEL ${imageMetadataLabel}="${imageMetadataLabelValue}"`;
 }
 
--- a/src/spec-node/imageMetadata.ts
+++ b/src/spec-node/imageMetadata.ts
@@ -498,8 +498,8 @@ export function getDevcontainerMetadataLabel(devContainerMetadata: SubstitutedCo
 		return '';
 	}
 	const imageMetadataLabelValue = `[${metadata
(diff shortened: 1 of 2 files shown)


fechu:

abdurriq (approved):

abdurriq:
Thank you for fixing this!

Issue #883, opened 2024-08-29
Variable substitution does not allow default values containing a colon


grthr:
The current implementation does not support default values in localEnv or containerEnv substitution that include a colon.

Example:
```
 "WEBSITE_URL": "${localEnv:WEBSITE_URL:https://example.com"
```

Current behavior: If `$WEBSITE_URL` is not present, the variable is just set to `https`

Expected behavior:  If `$WEBSITE_URL` is not present, the variable should be set to `https://example.com`

grthr:
PR #882

PR #1213, opened 2026-04-27
Allow colons in default values for environment variables

### LINKED ISSUE #719: How to specify the default value with colon for localEnv.
Hi DevContainers Team,

The case for me is to set the index url for pip, so what I do is something like

```json
{
    ...
    "remoteEnv": {
        "PIP_INDEX_URL": "${localEnv:PIP_INDEX_URL:https://pypi.org/simple}"
    }
}
```

I perfer a default value to handle the cases that `PIP_INDEX_URL` does not exist. But it turns out that when the env does not exist, the corresponding env in the vscode context will be set to `https` rather than `https://pypi.org/simple`. Seems it is related to the code here: https://github.com/devcontainers/cli/blob/main/src/spec-common/variableSubstitution.ts#L85. Any clues to make it happen?

cbguder:
Anything after a third colon is silently ignored in variable substitution, which prevents using default values with colons in them (e.g. URLs or image:tag).

This limits the split at 3 parts by joining any additional parts back together.

A similar fix was previously proposed in #882 a couple years ago. Open to exploring different solutions (e.g. quoting the entire default value or escaping colons) if there is still a desire to retain the ability to introduce more than 2 arguments in variable substitution.

Fixes #719, #883.

### DIFF
--- a/src/spec-common/variableSubstitution.ts
+++ b/src/spec-common/variableSubstitution.ts
@@ -85,7 +85,8 @@ function evaluateSingleVariable(replace: Replace, match: string, variable: strin
 	const parts = variable.split(':');
 	if (parts.length > 1) {
 		variable = parts[0];
-		args = parts.slice(1);
+		// Preserve colons in the default value (e.g. `${localEnv:VAR:image:tag}`).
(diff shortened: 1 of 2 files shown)


cbguder:

PR #1241, opened 2026-06-07
Add --platform argument to `devcontainer up`

mrexodia:
I noticed that when running `devcontainer up` the `docker pull` does not pass the `--platform` argument in `runArgs` correctly:

```json
{
	"name": "x86-boot-sector-c",
	"image": "ghcr.io/mrexodia/x86-real-mode-bootloader:latest",
	"runArgs": ["--platform=linux/amd64"] // Support Rosetta
}
```

Added a `--platform` CLI override to `devcontainer up` and extract the default from `runArgs`. The CLI argument wins.


mrexodia:
@abdurriq @chrmarti would it be possible to approve the workflows?

PR #1242, opened 2026-06-08
fix: set default base image in Dockerfiles and improve syntax handling

sireeshajonnalagadda:
What issue is about

The warning InvalidDefaultArgInFrom is not from the user’s project Dockerfile.
It comes from internal CLI Dockerfile flows where FROM uses an ARG that can look empty at parse time.
The reported hang after Starting container is likely a separate startup/event path, not this warning itself.
What has been done for you

Root cause was confirmed by tracing internal paths and issue context.
Warning fix was applied in updateUID.Dockerfile:3
ARG BASE_IMAGE changed to ARG BASE_IMAGE=placeholder.
Related hardening changes were also applied in generated feature Dockerfile templates:
containerFeaturesConfiguration.ts
containerFeatures.ts

Validation done

Local CLI build and run path was verified:
npm run compile
node devcontainer.js up --workspace-folder ./src/test/configs/example/ --build-no-cache --remove-existing-container

This succeeded with local source and removed the original BASE_IMAGE warning path in the tested flow.

Current changed files relevant to this issue

updateUID.Dockerfile
containerFeaturesConfiguration.ts
containerFeatures.ts


sireeshajonnalagadda (commented):
Defaulting BASE_IMAGE to placeholder will cause builds to fail with an invalid image reference if the build arg is omitted. If the intent is to provide a safe default, use a valid default value (e.g., scratch or a known minimal base)

vreencrawford9-eng (approved):

PR #1244, opened 2026-06-11
fix: relative symlink resolution in install script

### LINKED ISSUE #1232: install.sh wrapper relative link not resolved correctly
I'm trying to create a generic install method for a dotfile/software install library I maintain.  I want to install a link to the binary in `~/.local/bin/` and keep the devcontainers install under `~/.local/lib/devcontainers/`. The install takes care of this via the `--prefix ~/.local` argument. 

The problem is that when I create a relative link called devcontainer  from `~/.local/bin` to `~/.local/lib/devcontainers/bin/devcontainer` and then invoke the `devcontainer` executable, the generated bin file resolves the link as a relative link, rather than absolute. 

Example install method:
```shell
INSTALL_DIR=$HOME/.local
curl -sSLO https://raw.githubusercontent.com/devcontainers/cli/main/scripts/install.sh
chmod +x install.sh
./install.sh --prefix "${INSTALL_DIR}"/lib/devcontainers
cd "${INSTALL_DIR}"/bin
ln -s ../lib/devcontainers/bin/devcontainer devcontainer
```

Example running `devcontainer` which is already in the PATH of `~/.local/bin/` 

```shell
$devcontainer
/home/vasdee/.local/bin/devcontainer: line 18: cd: ../lib/devcontainers/bin: No such file or directory
```

In `~/.local/bin/devcontainer`  i believe if the follow link logic preferred the full path option of `readlink`, over the relative (which is the opposite of how it is now) then the issue would be solved....at least for me

```shell
# Resolve the installation directory
# Handle both direct execution and symlinked scenarios
if [ -L "$0" ]; then
    # Follow symlink current logic 
    SCRIPT_PATH="$(readlink "$0" 2>/dev/null || readlink -f "$0" 2>/dev/null || echo "$0")"
    # Follow symlink potential fix by preferring the abs path over relative
    # SCRIPT_PATH="$(readlink -f "$0" 2>/dev/null || readlink "$0" 2>/dev/null || echo "$0")"
else
    SCRIPT_PATH="$0"
fi
```

I am happy to contribute a PR, but am not sure of the history behind this choice and what impact the change might have.  I'm looking for guidance in that respect.

V-Subhankar-infy:
## Summary

Fixes relative symlink resolution in the install script for split local layouts (e.g. a launcher in `~/.local/bin` pointing to a payload in `~/.local/lib/devcontainers`). The wrapper now resolves its own location with **portable POSIX shell built-ins** instead of GNU-specific `readlink -f`, so `devcontainer` works regardless of the caller's working directory and shell environment (Linux, macOS, BSD).

Fixes [#1232](https://github.com/devcontainers/cli/issues/1232)

## Root Cause

A single `readlink` call plus `cd "$(dirname ...)" && pwd` resolved **relative** symlink targets against the caller's CWD rather than the link's own directory, and the `readlink -f` fallback isn't portable to macOS/BSD.

## Reproduction of error

~~~sh
./install.sh --prefix "$HOME/.local/lib/devcontainers"
ln -s ../lib/devcontainers/bin/devcontainer "$HOME/.local/bin/devcontainer"
export PATH="$HOME/.local/bin:$PATH"
cd "$HOME" && devcontainer --version   # cd: ../lib/devcontainers/bin: No such file or directory
~~~

## Fix

The script-location resolver now walks the symlink chain one hop at a time, `cd`-ing into each link's directory before reading the next target, then canonicalizes the final directory. This replaces the previous single `readlink` resolution.

## Why It Works

- **Correct relative resolution:** enters each link's directory before reading the next target, so relative symlinks resolve against the link, not the CWD.
- **CWD-independent:** finalized with `cd -P … && pwd` for an absolute, canonical path.
- **Portable:** POSIX-only (`readlink`, `${var%/*}`, `cd -P`, `pwd`) — no `readlink -f`/`realpath`.
- **Safe:** runs in a subshell with `CDPATH=` cleared; no leaked `cd` or stray output.

## Validation

Regression test confirms a relative symlink launched from another directory exits `0` and reports the installed CLI version.

## Scope & Risk

One resolution block; no install/download changes. Direct execution and absolute symlinks are unchanged. Low risk, plus improved cross-platform portability.

## Credits

Thanks to [@klmr](https://github.com/klmr) for the portable POSIX symlink-resolution approach.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -424,7 +424,7 @@ set -e
 # Handle both direct execution and symlinked scenarios
 if [ -L "$0" ]; then
     # Follow symlink
-    SCRIPT_PATH="$(readlink "$0" 2>/dev/null || readlink -f "$0" 2>/dev/null || echo "$0")"
+    SCRIPT_PATH="$(readlink -f "$0" 2>/dev/null || readlink "$0" 2>/dev/null || echo "$0")"
 else
     SCRIPT_PATH="$0"
 fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -422,16 +422,26 @@ set -e
 
 # Resolve the installation directory
 # Handle both direct execution and symlinked scenarios
-if [ -L "$0" ]; then
-    # Follow symlink
-    SCRIPT_PATH="$(readlink -f "$0" 2>/dev/null || readlink "$0" 2>/dev/null || echo "$0")"
-else
-    SCRIPT_PATH="$0"
-fi
+# Uses portable shell built-ins instead of GNU-specific readlink -f and realpath
+# Ignores inherited CDPATH, then resolves the script directory and its parent path
+SCRIPT_DIR="$(
+    CDPATH=
+    self=$0
 
-# Get absolute path to script directory
-SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_PATH")" && pwd)"
-INSTALL_DIR="$(dirname "$SCRIPT_DIR")"
+    # follow the symlink chain one hop at a time so relative targets resolve correctly
+    while [ -L "$self" ]; do
+        cd -- "${self%/*}" >/dev/null || :
+        self=$(readlink "${self##*/}")
+    done
+
+    case "$self" in
+        (*/*) ;;
+        (*) self=./$self ;;
+    esac
+
+    cd -P -- "${self%/*}" >/dev/null && pwd
+)"
+INSTALL_DIR="$(dirname -- "$SCRIPT_DIR")"
 
 # Paths to bundled components
 NODE_BIN="$INSTALL_DIR/node/current/bin/node"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -422,7 +422,7 @@ set -e
 
 # Resolve the installation directory
 # Handle both direct execution and symlinked scenarios
-# Uses portable shell built-ins instead of GNU-specific readlink -f and realpath
+# Uses POSIX shell syntax and avoids GNU-specific readlink -f/realpath (requires readlink(1))
 # Ignores inherited CDPATH, then resolves the script directory and its parent path
 SCRIPT_DIR="$(
     CDPATH=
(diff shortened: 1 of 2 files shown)


klmr:
Beware that `readlink -f` support has only been added to macOS [fairly recently](https://apple.stackexchange.com/a/464138/2392). A portable alternative is much more convoluted, alas:

```diff
--- a/devcontainer      2026-06-19 15:12:58
+++ b/devcontainer      2026-06-19 15:13:11
@@ -6,15 +6,22 @@

 # Resolve the installation directory
 # Handle both direct execution and symlinked scenarios
-if [ -L "$0" ]; then
-    # Follow symlink
-    SCRIPT_PATH="$(readlink "$0" 2>/dev/null || readlink -f "$0" 2>/dev/null || echo "$0")"
-else
-    SCRIPT_PATH="$0"
-fi
-
-# Get absolute path to script directory
-SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_PATH")" && pwd)"
+SCRIPT_DIR="$(
+    CDPATH=
+    self=$0
+
+    while [ -L "$self" ]; do
+        cd -- "${self%/*}" >/dev/null || :
+        self=$(readlink "${self##*/}")
+    done
+
+    case "$self" in
+        (*/*) ;;
+        (*) self=./$self ;;
+    esac
+
+    cd -P -- "${self%/*}" >/dev/null && pwd
+)"
 INSTALL_DIR="$(dirname "$SCRIPT_DIR")"

 # Paths to bundled components
```

(Reference/explanation: https://stackoverflow.com/a/246128/1968)

scripts/install.sh
```diff
@@ -422,16 +422,27 @@ set -e
...
+
+    case "$self" in
+        (*/*) ;;
+        (*) self=./$self ;;
+    esac
 
-# Get absolute path to script directory
-SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_PATH")" && pwd)"
-INSTALL_DIR="$(dirname "$SCRIPT_DIR")"
+    cd -P -- "${self%/*}" >/dev/null && pwd
```
abdurriq, line 443: I don't think `cd --` would work on POSIX/macOS/BSD environments.

abdurriq (changes requested):

scripts/install.sh
V-Subhankar-infy: The cd -- (option terminator) has been part of systems since 1995 and I have just checked on zsh shell in my macos and it supports this.

To be precise, this is completely safe to use cd -- on any system from 2005 onward:

- macOS 10.0+ (2001+)
- FreeBSD 6.0+ (2005+)
- OpenBSD 5.0+ (2008+)
-  NetBSD 1.5+ (1999+)

I am not aware of how much backward compatibility do we need to have, but it seems like a valid decision, considering readlink even the basic version is relatively more recent.

abdurriq (approved):

PR #1249, opened 2026-06-23
Add WSLC support

chrmarti:


src/spec-shutdown/dockerUtils.ts
```diff
@@ -260,13 +265,16 @@ export async function dockerBuildKitVersion(params: DockerCLIParameters | Partia
 	}
 }
 
-export async function dockerEngineVersion(params: DockerCLIParameters | PartialExecParameters | DockerResolverParameters): Promise<{ versionString: string; versionMatch?: string } | undefined> {
+export async function dockerEngineVersion(params: DockerCLIParameters | PartialExecParameters | DockerResolverParameters, options?: { useSimpleVersion?: boolean }): Promise<{ versionString: string; versionMatch?: string } | undefined> {
```
abdurriq, line 268: Q: Curious why we have these helper methods in the shutdown module? I see that the original methods were here too, but I wonder if there's a better place for them in the codebase (not necessary for this PR).

src/spec-node/singleContainer.ts
```diff
@@ -347,8 +347,8 @@ export async function spawnDevContainer(params: DockerResolverParameters, config
 	const exposedPorts = typeof appPort === 'number' || typeof appPort === 'string' ? [appPort] : appPort || [];
 	const exposed = (<string[]>[]).concat(...exposedPorts.map(port => ['-p', typeof port === 'number' ? `127.0.0.1:${port}:${port}` : port]));
 
-	const cwdMount = workspaceMount ? ['--mount', workspaceMount] : [];
-	const additionalMount = additionalMountString ? ['--mount', additionalMountString] : [];
+	const cwdMount = workspaceMount ? (params.cliVariant === CLIVariant.Wslc ? convertMountToVolume(workspaceMount) : ['--mount', workspaceMount]) : [];
```
abdurriq, line 350: Could we have tests for all the differences Wslc expects, such as the mount commands, as well as the dropped flags (`--init`, `--sig-proxy=false`, etc.) so that we can catch future regressions?

src/spec-node/singleContainer.ts
```diff
@@ -446,6 +450,32 @@ async function getPodmanArgs(params: DockerResolverParameters, config: DevContai
 	return [];
 }
 
+// Convert a --mount string (e.g., "type=bind,source=/a,target=/b,consistency=cached") to -v syntax for wslc.
+function convertMountToVolume(mountStr: string): string[] {
```
abdurriq, line 454: It would also be nice to have unit-tests for these.

abdurriq (approved):
Approved, with tests to be added as a fast-follow.

src/spec-shutdown/dockerUtils.ts
chrmarti: The shutdown module was just convenient for reuse. I think the next lower module is free of Docker specifics, so we'd have to create a new module just for the Docker code we want to reuse. Open for suggestions, thanks!

PR #1127, opened 2026-01-08
Add option to mount a worktree's common folder

chrmarti:
https://github.com/microsoft/vscode-remote-release/issues/2215
https://github.com/devcontainers/cli/issues/796

PR #1261, opened 2026-07-02
Mount worktree common dir when a custom workspaceMount is set

### LINKED ISSUE #1243: --mount-git-worktree-common-dir silently ignored when devcontainer.json sets a custom workspaceMount
### Summary

`--mount-git-worktree-common-dir` (added in #1127) has no effect when the `devcontainer.json` defines a custom `workspaceMount`. The worktree common-dir bind mount is never added, so in-container git fails with `fatal: not a git repository` for any git worktree whose devcontainer uses a custom `workspaceMount`.

### Root cause

In `src/spec-node/utils.ts` (`getWorkspaceConfiguration`, ~L386) the entire worktree detection + common-dir mount block is guarded by:

```js
if (workspace && (!workspaceFolder || !('workspaceMount' in config))) { ... }
```

When `workspaceMount` is present, the block is skipped — the `.git` file is never read, the common dir is never resolved, and no mount is added. (Flagged as a possible oversight by @kapouer in #796, Mar 2026.)

### Reproduction (CLI 0.82.0, git 2.52)

1. `git init main && (cd main && git commit --allow-empty -m init && git worktree add --relative-paths ../wt)`
2. Add `wt/.devcontainer/devcontainer.json` with a custom `workspaceMount`, e.g.:
   ```json
   { "image": "mcr.microsoft.com/devcontainers/base:debian",
     "workspaceMount": "source=${localWorkspaceFolder},target=/workspace,type=bind",
     "workspaceFolder": "/workspace" }
   ```
3. `devcontainer up --workspace-folder wt --mount-git-worktree-common-dir true`
4. `devcontainer exec --workspace-folder wt -- git -C /workspace status` → `fatal: not a git repository: (null)`
5. `docker inspect` on the container shows only the workspace bind mount; the worktree common dir was not mounted.

Removing `workspaceMount` makes it work — confirming the guard is the cause.

### Expected

The common-dir mount should be computed and added even when a custom `workspaceMount` is set, by resolving the mount target relative to the configured `workspaceMount` target / `workspaceFolder`. This is the common case for monorepos that mount a parent directory and set `workspaceFolder` to a subpath. Relates to #796.

### Workaround

Adding the common dir manually via a `mounts` entry (binding the main repo's `.git` to the path the worktree's `.git` file resolves to inside the container) restores in-container git — matching the maintainer's earlier suggestion in #796.

AvishayHirsh:
Fixes #1243

## Problem

`--mount-git-worktree-common-dir` (added in #1127) is silently ignored when a `devcontainer.json` sets a custom `workspaceMount`. Git worktrees then break inside the container:

```
$ devcontainer exec --workspace-folder wt -- git -C /workspace status
fatal: not a git repository: (null)
```

This is common in monorepo setups where a custom `workspaceMount`/`workspaceFolder` is used.

## Root cause

In `getWorkspaceConfiguration` (`src/spec-node/utils.ts`), the worktree common-dir computation lives inside the guard that only auto-computes `workspaceMount`/`workspaceFolder` when they aren't already set in the config:

```ts
if (workspace && (!workspaceFolder || !('workspaceMount' in config))) { ... }
```

When the config sets **both** `workspaceFolder` and `workspaceMount`, the whole block is skipped, so `additionalMountString` (the common-dir bind mount) is never built and the main repo's `.git` is never mounted.

## Fix

The common-dir mount is an *additive, separate* mount — it can't conflict with a user-defined `workspaceMount`, so it shouldn't be gated by the "don't overwrite user config" guard:

- Widen the guard so the worktree block also runs when the feature is enabled. `--mount-git-worktree-common-dir` defaults to `false`, so **default behavior is unchanged** unless the flag is opted into.
- Resolve the common-dir target relative to the custom `workspaceMount` target. Because `additionalMountString` is not part of the later variable-substitution pass, the target is substituted here (handles the canonical `target=/workspaces/${localWorkspaceFolderBasename}`). Without a custom mount, the previously computed target is used, so existing behavior is unchanged.

## Tests

Added unit tests in `src/test/workspaceConfiguration.test.ts` (across linux/darwin/win32) covering: the reported scenario, `${localWorkspaceFolderBasename}` substitution in the target, feature-off no-op, non-worktree no-op, and a worktree nested inside the main repo. `yarn lint`, `yarn type-check`, and the test suite all pass locally.

## Known limitations (pre-existing, unchanged by this PR)

- Only a relative `gitdir` is handled (worktrees created with `git worktree add --relative-paths`); absolute gitdirs were already unsupported.
- Assumes the custom mount's `source` is the worktree root (the directory holding the `.git` file).

### DIFF
--- a/src/spec-node/utils.ts
+++ b/src/spec-node/utils.ts
@@ -19,11 +19,12 @@ import { ShellServer } from '../spec-common/shellServer';
 import { inspectContainer, inspectContainers, inspectImage, getEvents, listContainers, ContainerDetails, DockerCLIParameters, dockerExecFunction, dockerPtyCLI, dockerPtyExecFunction, toDockerImageName, DockerComposeCLI, ImageDetails, dockerCLI, removeContainer, CLIVariant } from '../spec-shutdown/dockerUtils';
 import { getRemoteWorkspaceFolder } from './dockerCompose';
 import { findGitRootFolder } from '../spec-common/git';
+import { substitute } from '../spec-common/variableSubstitution';
 import { parentURI, uriToFsPath } from '../spec-configuration/configurationCommonUtils';
 import { DevContainerConfig, DevContainerFromDockerfileConfig, getConfigFilePath, getDockerfilePath } from '../spec-configuration/configuration';
 import { StringDecoder } from 'string_decoder';
 import { Event } from '../spec-utils/event';
-import { Mount } from '../spec-configuration/containerFeaturesConfiguration';
+import { Mount, parseMount } from '../spec-configuration/containerFeaturesConfiguration';
 import { PackageConfiguration } from '../spec-utils/product';
 import { ImageMetadataEntry, MergedDevContainerConfig } from './imageMetadata';
 import { getImageIndexEntryForPlatform, getManifest, getRef } from '../spec-configuration/containerCollectionsOCI';
@@ -417,7 +418,7 @@ export async function getWorkspaceConfiguration(cliHost: CLIHost, workspace: Wor
 	}
 	let { workspaceFolder, workspaceMount } = config;
 	let additionalMountString: string | undefined;
-	if (workspace && (!workspaceFolder || !('workspaceMount' in config))) {
+	if (workspace && (!workspaceFolder || !('workspaceMount' in config) || (mountWorkspaceGitRoot && mountGitWorktreeCommonDir))) {
 		const hostMountFolder = await getHostMountFolder(cliHost, workspace.rootFolderPath, mountWorkspaceGitRoot, output);
 
 		// Check if .git is a file (worktree) with a relative gitdir path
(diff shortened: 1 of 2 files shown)


AvishayHirsh:

AvishayHirsh:
Thanks for the worktree common-dir feature in #1127! This is a small follow-up for #1243: the common-dir mount is skipped when `devcontainer.json` sets a custom `workspaceMount` (common in monorepos), so worktree git operations break in the container (`fatal: not a git repository`).

The fix is gated behind the existing `--mount-git-worktree-common-dir` flag (default off → no behavior change otherwise), and resolves the common-dir target relative to the custom (substituted) `workspaceMount` target. Added unit tests across linux/darwin/win32. CLA is signed; `yarn lint`, `yarn type-check`, and the test suite pass locally.

@Mathiyarasy @chrmarti would appreciate a workflow-run approval and a review whenever you have a moment — thanks!

AvishayHirsh:
Hello world... is there anybody out there ?

kapouer:
With that PR accepted, what will be needed for vscode "reopen in container" to support git worktrees out of the box ?

PR #1267, opened 2026-07-19
Sort Features with more user-defined options first, per spec

### LINKED ISSUE #755: CLI sorts Feature options fewest-first, against spec
> **Note on repo choice:** Per the [`devcontainers/cli` CONTRIBUTING guide](https://github.com/devcontainers/cli/blob/main/CONTRIBUTING.md), issues concerning the CLI should be filed in this spec repository, so I've opened it here even though it's about `devcontainers/cli` behavior. Please let me know if you'd prefer it moved to the `devcontainers/cli` repo instead.

### Summary

When sorting Features within an installation round, the spec says Features are ordered by the **greatest** number of user-defined options first, but `containerFeaturesOrder.ts` orders by the **fewest** options first.

### Specification

[Features spec — Definition: Round Stable Sort](https://containers.dev/implementors/features/#definition-feature-equality):

> Greatest number of user-defined options (note omitting an option will default that value to the Feature's default value and is not considered a user-defined option)

### Implementation

`optionsCompareTo` returns `aKeys.length - bKeys.length`, which is negative when `a` has *fewer* options, so the Feature with fewer options sorts *earlier*:

https://github.com/devcontainers/cli/blob/f683c29f64a20109b4453e5149807e390ff65133/src/spec-configuration/containerFeaturesOrder.ts#L123-L125

This is the option tiebreaker for every source type (OCI, file-path, direct-tarball, legacy).

### Test evidence

The existing test `valid dependsOn with round sorting based on options` (fixture `configs/feature-dependencies/dependsOn/local-with-options`) pins the fewest-first behavior — the `./b` variant with `{}` (zero options) is expected first, ahead of the two-option variants:

https://github.com/devcontainers/cli/blob/f683c29f64a20109b4453e5149807e390ff65133/src/test/container-features/containerFeaturesOrder.test.ts#L135-L209

So the implementation and its tests are internally consistent; only the behavior disagrees with the spec.

### Proposed fix

Sort by the greatest option count to match the spec (`bKeys.length - aKeys.length`), with the fixture/expected order updated accordingly. This changes the install order of the same Feature requested with differing option sets within a round.

nozaq:
The Features spec's "Installation Order" rules state that Features in a round are sorted by "Greatest number of user-defined options", but optionsCompareTo compared option counts ascending, installing the Feature with the fewest options first.

Flip the length comparison so the Feature with the greatest number of options sorts first, and update the round-sorting test expectations accordingly. Equal-length option sets still tie-break by key/value comparison as before.

fixes https://github.com/devcontainers/spec/issues/755

### DIFF
--- a/src/spec-configuration/containerFeaturesOrder.ts
+++ b/src/spec-configuration/containerFeaturesOrder.ts
@@ -121,7 +121,7 @@ function optionsCompareTo(a: string | boolean | Record<string, string | boolean 
 		const aKeys = Object.keys(a);
 		const bKeys = Object.keys(b);
 		if (aKeys.length !== bKeys.length) {
-			return aKeys.length - bKeys.length;
+			return bKeys.length - aKeys.length;
 		}
 
 		aKeys.sort();
--- a/src/test/container-features/containerFeaturesOrder.test.ts
+++ b/src/test/container-features/containerFeaturesOrder.test.ts
@@ -156,35 +156,35 @@ describe('Feature Dependencies', function () {
                 [
                     {
                         userFeatureId: './b',
+                        options: {
+                            optA: 'a',
+                            optB: 'a'
+                        }
+                    },
+                    {
+                        userFeatureId: './b',
+                        options: {
+                            optA: 'a',
+                            optB: 'b'
+                        }
+                    },
+                    {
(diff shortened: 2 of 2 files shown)


nozaq:

Issue #1229, opened 2026-05-15
Devcontainer Build without cache hangs after mostly successful build


ontoneio:
So I am running into a weird issue with my devcontainer environment. I am building without the cache inside a Vscode IDE and the image build seems to get to the very last step and then throws this warning and does not ever resolve the build into the devcontainer environment.

```bash
 1 warning found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG $BASE_IMAGE results in empty or invalid base image name (line 2)
[303440 ms] Start: Run: docker events --format {{json .}} --filter event=start
[303442 ms] Start: Starting container
```

I personally do not have a `BASE_IMAGE` Argument present in my docker image, so I am not understanding why my build would be failing to propagate due to a variable I don't use?

For debugging sake I even tried adding these lines and still get the error from above...

```dockerfile
ARG BASE_IMAGE=mcr.microsoft.com/vscode/devcontainers/base:ubuntu-24.04
FROM ${BASE_IMAGE} AS base
```

Why would I be getting this error?

ontoneio:
@samruddhikhandale OR @chrmarti have either of you seen this before?

I have even put this docker check directive in my Dockerfile and it still pauses the build without resolution.

```dockerfile
# syntax=docker/dockerfile:1
# check=skip=InvalidDefaultArgInFrom
```

sireeshajonnalagadda:
Hi @ontoneio,

Thank you for reporting this issue!

This warning is likely coming from devcontainers/cli’s internal build logic, not from your project’s Dockerfile.

The repo contains an internal Dockerfile fragment like:

Dockerfile
ARG BASE_IMAGE
FROM $BASE_IMAGE
This can trigger BuildKit’s InvalidDefaultArgInFrom warning because the FROM instruction depends on an ARG with no inline default at parse time.

So, seeing a BASE_IMAGE warning does not necessarily mean your project’s Dockerfile defines BASE_IMAGE.

Also, this warning may not be the root cause of the failure — the log suggests the hang happens later, during container startup (Start: Starting container).

If possible, could you please share your Dockerfile so I can try to reproduce the issue? I’ll investigate further once I have that information.

Thank you,
Sireesha

ontoneio:
Hello @sireeshajonnalagadda 

Thanks for the response. I will share my Dockerfile with some info redacted.

```dockerfile
# syntax=docker/dockerfile:1
# check=skip=InvalidDefaultArgInFrom

[...]

PR #1271, opened 2026-07-28
fix: set default base image in Dockerfiles and improve syntax handling

Kaniska244:
## What this PR does

Related #1229. Fixes the `InvalidDefaultArgInFrom` build warning emitted by Docker/BuildKit when building dev containers. The warning is raised whenever a `FROM` instruction references a build `ARG` that has no in-scope default value at parse time.

```
 1 warning found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG $BASE_IMAGE results in empty or invalid base image name (line 2)
[303440 ms] Start: Run: docker events --format {{json .}} --filter event=start
[303442 ms] Start: Starting container
```

This warning does **not** come from the user's project Dockerfile — it originates from the internal Dockerfiles the CLI generates (and the static `updateUID.Dockerfile`), where `FROM` referenced `_DEV_CONTAINERS_BASE_IMAGE` / `BASE_IMAGE` without a declared default.

## Root cause

Two patterns were triggering the linter:

1. **No default on the ARG** — `scripts/updateUID.Dockerfile` declared `ARG BASE_IMAGE` (no default) and then used `FROM $BASE_IMAGE`.
2. **Shell-style fallback that the linter doesn't credit** — the generated feature Dockerfiles used `FROM ${_DEV_CONTAINERS_BASE_IMAGE:-scratch}`. The `${VAR:-scratch}` form is a runtime shell expansion; it is **not** recognized as an `ARG` default by the static linter, so the warning persisted even when a global `ARG ... = placeholder` was declared.

## The fix

Declare the base-image `ARG` with a real default (`scratch` for the generated files, `placeholder` for the UID script) immediately before it's used, and reference it plainly as `$VAR` — the form the linter recognizes as having a valid default.

- **`src/spec-configuration/containerFeaturesConfiguration.ts`** — add `ARG _DEV_CONTAINERS_BASE_IMAGE=scratch` before the `FROM` stages in the generated feature base Dockerfile (the `FROM` lines already reference the ARG directly).
- **`src/spec-node/containerFeatures.ts`** — change the prefix `ARG` default from `placeholder` to `scratch` in both `getImageBuildOptions` (no-features path) and `getFeaturesBuildOptions` (features path).
- **`scripts/updateUID.Dockerfile`** — change `ARG BASE_IMAGE` → `ARG BASE_IMAGE=placeholder`.

### Behavior is unchanged at build time

The CLI always passes the real image via `--build-arg` (`_DEV_CONTAINERS_BASE_IMAGE=<image>`, and `BASE_IMAGE=<image>` for the UID Dockerfile), so the `scratch` / `placeholder` defaults only ever exist to satisfy the static linter and never affect the actual build.

## Tests

Added a daemon-free unit test suite in `src/test/container-features/generateFeaturesConfig.test.ts` (`validate generated Dockerfiles avoid InvalidDefaultArgInFrom`) that:

- Introduces a `findFromArgsWithoutDefault()` helper mimicking the `InvalidDefaultArgInFrom` rule (flags any `ARG` used in a `FROM` without a default declared before it).
- Asserts the generated feature base Dockerfile declares `_DEV_CONTAINERS_BASE_IMAGE` with a default, references it directly, and does **not** use the `${VAR:-default}` shell fallback.
- Asserts `scripts/updateUID.Dockerfile` declares `BASE_IMAGE` with a default before `FROM`.

These run without Docker, so they execute in CI as part of `npm test`.

## Files changed

| File | Change |
|------|--------|
| `src/spec-configuration/containerFeaturesConfiguration.ts` | Declare `ARG _DEV_CONTAINERS_BASE_IMAGE=scratch` before `FROM` |
| `src/spec-node/containerFeatures.ts` | ARG default `placeholder` → `scratch` in both build-option paths |
| `scripts/updateUID.Dockerfile` | `ARG BASE_IMAGE` → `ARG BASE_IMAGE=placeholder` |
| `src/test/container-features/generateFeaturesConfig.test.ts` | Add `InvalidDefaultArgInFrom` regression tests |

## Notes / follow-ups

- The `getImageBuildOptions` no-features path is covered indirectly (its ARG default was corrected), but its Dockerfile string is built inline and isn't directly unit-tested. A small follow-up could extract that string into an exported helper to unit-test it the same way.
- `.devcontainer/devcontainer-lock.json` shows a trailing-newline-only change; consider reverting that if it's unintentional.


src/test/container-features/generateFeaturesConfig.test.ts
```diff
@@ -135,4 +135,52 @@ RUN chmod -R 0755 /tmp/dev-container-features/hello_1 \\
         const javaSettings = java?.features[0]?.customizations?.vscode?.settings;
         assert.isObject(javaSettings);
     });
+});
+
+// returns the names of any ARGs used in a `FROM` that lack a default declared before them.
+function findFromArgsWithoutDefault(dockerfile: string): string[] {
```
abdurriq: Do we not also want to use this for other features?

gvatsal60:
Hi @Kaniska244,
Would it be possible to merge this PR when you get a chance?

The current behaviour is breaking existing workflows.

src/test/container-features/generateFeaturesConfig.test.ts
Kaniska244: Hi @abdurriq 

Thank you for pointing this out. It's done now. Kindly let me know in case of any concern.

abdurriq (approved):

src/spec-configuration/containerFeaturesConfiguration.ts
```diff
@@ -203,6 +203,8 @@ export function getContainerFeaturesBaseDockerFile(contentSourceRootPath: string
 
 #{nonBuildKitFeatureContentFallback}
 
+ARG _DEV_CONTAINERS_BASE_IMAGE=scratch
+
 FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_feature_content_normalize
```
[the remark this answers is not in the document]
Kaniska244: Hi @abdurriq 

Kindly note that its premise ignores the global ARG inserted later [here](https://github.com/Kaniska244/cli/blob/invalid-arg-fix/src/spec-node/containerFeatures.ts#L273-L279) and evaluates an intermediate fragment that is not independently consumed. Moving the declaration is perhaps harmless, but it is unnecessary and would encode the wrong abstraction boundary in the test.

chrmarti (approved):

PR #1278, opened 2026-08-11
Harden OCI registry authentication

chrmarti:
## Summary

Harden OCI Feature and Template authentication against registry-controlled Bearer realms that could trigger SSRF or forward registry credentials to an attacker-selected endpoint.

This builds on the work in #1276 while narrowing the change to the authentication boundary and adding a registry-scoped compatibility escape hatch for legitimate external token services.

<details>
<summary>Session Context</summary>

Key decisions from the development session:

- **Keep registry parsing unchanged**: Localhost, private-IP, single-label, IPv6, and custom-port registries are legitimate. The vulnerability is addressed at the secondary token-request boundary instead of adding a broad registry URL policy.
- **Opt-in hardening**: Legacy OCI authentication behavior remains the default. `--oci-auth-hardening` enables realm restrictions, the requested-registry credential-origin gate, and token redirect refusal.
- **Bind compatibility exceptions to both hosts**: `--allow-cross-origin-auth-host <registry-host>=<auth-host>` is repeatable. A bare auth-host allowlist would let any workspace-selected registry invoke that host as a confused deputy.
- **Keep built-ins minimal**: Docker Hub and GitLab.com are the verified public cross-origin flows. GHCR, Quay, GCR, public ECR, and normal ACR flows are same-origin and use the default policy.
- **Trust the full mapped exchange**: Exact trusted mappings may receive anonymous token requests, Basic credentials, and refresh/identity tokens. The exact HTTPS authority pair is the trust boundary.
- **Fail closed on redirects when enabled**: With `--oci-auth-hardening`, token requests do not follow redirects. Normal registry requests retain existing redirect behavior, and the final response URL is used to validate the challenge origin.
- **Preserve localhost compatibility**: Exact-authority `http://localhost` token realms remain supported for local Feature development; remote HTTP realms remain rejected.

</details>

## Changes

- Add `--oci-auth-hardening` as the explicit opt-in for behavior-changing authentication protections.
- When enabled, validate bearer realms before credential lookup or token endpoint I/O.
- Permit exact-authority HTTPS realms and exact-authority localhost HTTP realms.
- Add built-in Docker Hub and GitLab.com registry-to-auth-host mappings.
- Add repeatable `--allow-cross-origin-auth-host <registry-host>=<auth-host>` mappings for enterprise registries.
- Forward registry credentials, including refresh/identity tokens, across explicitly trusted origin mappings.
- When enabled, disable redirects for token acquisition and handle redirect refusal as an authentication failure.
- Encode `service` and `scope` with `URL.searchParams`.
- Rebuild anonymous fallback requests without credential headers or refresh-token bodies.
- Add focused authentication policy and HTTP behavior tests.

## Validation

- 41 focused OCI authentication tests pass.
- `yarn type-check`
- `yarn lint`
- `yarn package`
- Packaged CLI help exposes the new global option.

Fixes github/codespaces#24328.


chrmarti:
Companion VS Code Dev Containers integration: microsoft/vscode-remote-containers#1235

chrmarti:
Follow-up 7b8e496 removes the internal environment-variable bridge. The option now propagates explicitly as typed command/resolver/request state across OCI paths. It also sets yargs `nargs: 1` so repeated global mappings cannot consume the command positional, with packaged CLI and request-path regression coverage.

chrmarti:
Updated trusted host-pair semantics in 87adb63: an exact HTTPS registry-to-auth-host mapping now authorizes both Basic and refresh/identity-token exchange. Added a real cross-authority Docker `identitytoken` POST regression test; all 41 focused tests, type-check, and lint pass.

src/test/httpOCIRegistry.test.ts
```diff
@@ -0,0 +1,361 @@
...
+			assert.notProperty(cachedAuthHeader, registry);
+		} finally {
+			await Promise.all([close(registryServer), close(tokenServer)]);
+		}
+	});
+
+	it('forwards a refresh token to an explicitly configured auth host', async () => {
+		const token = 'registry-token';
+		const refreshToken = 'registry-refresh-token';
+		const bearerScheme = ['Bear', 'er'].join('');
```
abdurriq: Q: Is this for preventing a credential scanner from triggering on this?

abdurriq (dismissed):

src/test/httpOCIRegistry.test.ts
chrmarti: Copilot had trouble applying this string and used this as a workaround (it explained to me when I asked). Removing it, thanks.

chrmarti:
Updated in 7c650d5: behavior-changing OCI authentication protections are now opt-in through `--oci-auth-hardening`. Legacy cross-origin realm and redirect behavior remains the default; focused tests cover both modes.

chrmarti:
Updated in 3ba63a5: add shadow diagnostics for auth lookups that hardening would block, registry redirects that would prevent forwarding requested-registry credentials, and auth-server redirects. Diagnostics are logged and surfaced in CLI results.

chrmarti:
Prepared version 0.89.0 and added the August 2026 changelog entry in 422a92e.

chrmarti:
Refined the registry redirect diagnostic in 987bbc7 so cross-origin content/blob redirects are ignored; it now reports only redirected 401/403 challenges that would change credential forwarding.

chrmarti:
Added 98e182e to keep the Podman matrix rootless while disabling the hosted runner bundle's  storage option before Feature builds.

chrmarti:
Moved the unrelated Podman runner fix to independent PR #1280; the OCI branch now contains reverts of the prior Podman workaround commits.

src/spec-node/devContainersSpecCLI.ts
```diff
@@ -66,6 +68,28 @@ const mountRegex = /^type=(bind|volume),source=([^,]+),target=([^,]+)(?:,externa
 		.scriptName('devcontainer')
 		.version(version)
 		.demandCommand()
+		.option('oci-auth-hardening', {
+			type: 'boolean',
+			default: false,
+			global: true,
```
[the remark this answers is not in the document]
chrmarti: Addressed in 5c5e0c48. Both global OCI auth options now flow through FeaturesTestCommandInput and generateDockerParams into createDockerParams, and launchProject copies them from common into ProvisionOptions for the actual launch pass.

abdurriq (approved):

PR #1285, opened 2026-08-23
Fix podman keep-id mapping for numeric and named users - fixes #1284

### LINKED ISSUE #1284: Podman fix: derive `--userns=keep-id` mapping from the remote user's actual UID/GID
## The problem

When running a dev container with rootless podman, the Dev Containers CLI injects a plain `--userns=keep-id` flag, which maps the host user id to the *same UID number* inside the container.

This breaks for high/out-of-range host UIDs (e.g. AD/SSSD users which have values like`1400601154`, which are not valid inside the container (they are out of range as the containers are typically given a 0-65535 range via subuid/subgid mapping). As a result the mapping fails silently and the container's `vscode` user remains at UID 1000, not matching the host user, and causing `Permission denied` on the workspace bind mount).

## Current workaround

The current workaround is a fragile manual `runArgs` entry (`--userns=keep-id:uid=1000,gid=1000`). Since the host UID is unmappable to itself, we keep the 1000 id inside the container and map the host user to that. This allows one to use podman on Linux boxes with AD and thus high uid/gid combinations, as it removes the requirement for the exact host UID to be inside the container's range.

## The request for improvement

Instead of injecting a plain `--userns=keep-id`, the CLI should resolve the `remoteUser`'s actual UID/GID in the image and emit `--userns=keep-id:uid=<remote-uid>,gid=<remote-gid>`. This fixes the high-UID case with no regressions: when the host UID already equals the container user's UID, `keep-id:uid=X,gid=X` produces the identical mapping to plain `keep-id`, so existing setups are unaffected.

karypid:
Prefer parsing a numeric user spec directly from config and only fall back to resolving named users via a throwaway container, producing an explicit --userns=keep-id:uid=...,gid=... mapping. Add tests and config fixtures for numeric and named keep-id cases.

Addresses #1284

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Notable changes.
 
 ### [0.88.0]
 - Add WSLc support (https://github.com/devcontainers/cli/pull/1249)
+- Derive the `--userns=keep-id` mapping from the remote user's actual UID/GID when using Podman, so the container user is mapped to the host user even when their UIDs differ (e.g. high UIDs from AD/SSSD). (https://github.com/devcontainers/cli/issues/1284)
 
 ## May 2026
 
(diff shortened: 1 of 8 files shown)


karypid:

PR #1287, opened 2026-08-26
Chown fail fix

karypid:
When podman maps a high ID to a  low ID there is no need to `chown` the workspace. Inside podman the low ID applies both to user (vscode by default, 1000), as well as the mounted workspace.

This complimentary logic makes sure to behave as `updateRemoteUserUID: false`

Once again, if any explicit user configuration is present it will override this "internal default" change so everything is backwards compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Notable changes.
 
 ### [0.88.0]
 - Add WSLc support (https://github.com/devcontainers/cli/pull/1249)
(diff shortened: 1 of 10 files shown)


karypid:
This is a complement to fix #1284 

I have separated the PRs as they are orthogonal.

Both are needed for correct operation when the host UID/GID is abnormally high (as is the case when using SSSD or FreeIPA).

Issue #772, opened 2024-03-12
Secrets for containerEnv


g-kartik:
Currently the --secrets-file sets the secret environment variables as `remoteEnv`. A support for `containerEnv` shall ensure that all the running processes inside the devcontainer can access them.

samruddhikhandale:
I wonder if it's very similar to the ask in https://github.com/orgs/devcontainers/discussions/119. 

From https://github.com/orgs/devcontainers/discussions/119#discussioncomment-8761462

> Not sure if a static env file would cover your case, but there is a feature request for supporting .env files: https://github.com/microsoft/vscode-remote-release/issues/3007.

g-kartik:
Yes it is similar to https://github.com/microsoft/vscode-remote-release/issues/3007

Issue #22, opened 2022-05-19
Add a reference implementation of port forwarding


bamurtaugh:
The specification includes port attributes, but the dev container CLI doesn't include port forwarding support.

Chuxel:
👍On this one. While you can use `appPort`, this is currently only wired into the single container setup and is port publishing rather than full forwarding (e.g., if something only listens to localhost it won't work).  While implementors can communicate with a server via text stream for `devcontainer exec`, it's pretty common to need to communicate on a port instead.

chrmarti:
I was thinking of dynamic port forwarding where we could start something along the lines of `devcontainer port-forward --workspace-folder my_project` and that would forward the ports configured in the devcontainer.json (plus any that we might list in command line arguments).

E.g., we have implemented that in the Dev Containers extension for VS Code by locally listening on ports and forward any connections using `docker exec` to run a little JavaScript snippet inside the container to connect to the target port there. The problem (and the reason we haven't included it in the dev container CLI) with this implementation is that it requires NodeJS inside the container. You could use `socat` inside the container to achieve the same, but that would then require `socat` to be available.

MunsMan:
@chrmarti your example was a helpful confirmation, I had to reverse engineer VS Code to get that inside. 

I added 'static' forwarding via devontainer.json in this branch (https://github.com/MunsMan/devcontainercli/tree/port_forwarding), there it implemented via the Docker CLI and the publish option. Currently, `socat` is not part of the default image to my knowledge, right? VS Code mounts it own node runtimes and server extensions, right?

For automatic forwarding, should it be a feature then? When we demand code in the container, it should be a feature. I was thinking if forwarding is possible by mounting sockets via volumes into the container, but I will build a prototype for myself first.

MunsMan:
I built a prototype for automatic forwarding ports via a UNIX Socket (I had the assumption that WSL should support that), but even macOS didn't support sharing UNIX Sockets via a volume. The macOS support will not come in the near future (https://github.com/docker/for-mac/issues/483).
I build the Prototype in Rust and build images for the corresponding container, which then can directly be mounted. For Linux to Container or Container to Container, I had a Response Time which was 50 Times faster than the one of VS Code. 

[...]

PR #1306, opened 2026-09-18
Add support for the `extends` keyword

SamuelFrost:
(note for reviewers: this is vibe coded based off the existing PR and issues. I have not reviewed it yet beyond minimal local testing-- hence the draft status)

## Summary

Implements configuration inheritance from [spec#22](https://github.com/devcontainers/spec/issues/22): a `devcontainer.json` can inherit another JSON or JSONC file in the same repository via `extends`.

This reapplies [PR #311](https://github.com/devcontainers/cli/pull/311) on current `main`. That branch could not be rebased as a delta. The merge path follows the review on #311: use the image metadata merge logic, not `deepmerge`.

Companion spec PR: https://github.com/devcontainers/spec/pull/772

Does not bump the package version. That belongs in a later release PR.

## Changes

- Add optional `extends?: string` on the `devcontainer.json` config types.
- Resolve `extends` when reading configuration: relative paths only, recursive, with cycle detection. Absolute paths and URLs are rejected.
- Merge the referenced file with the current file using the existing image metadata rules (`forwardPorts` / `capAdd` union, `hostRequirements` max, last-wins for scalars, `init` / `privileged` true if any value is true). Strip `extends` from the result.
- Add tests for a single reference, nested references, cycles, a missing file, and overlapping merge properties.
- Note the change under Unreleased in `CHANGELOG.md` without cutting a version.

## Technical choices

#311 used `deepmerge`. Christof Marti, Josh Spicer, and Chuck Lantz asked to reuse image metadata merge so extending a file matches combining a prebuilt image’s metadata with a project `devcontainer.json`. `mergeDevContainerConfigs` in `imageMetadata.ts` is that shared path.

Relative-path-only matches the spec issue (same repository). Recursive `extends` is included with cycle detection; #311 left nested references unresolved.

## Validation

- `src/test/configContainer.test.ts`: 6 passing (read a basic file, resolve `extends`, nested `extends`, cyclic reference, missing file, merge overlap).
- `yarn type-check`
- `yarn lint`

`devcontainer up` was not run as part of this change. `read-configuration` against a file that extends another produces the merged config and omits `extends`.

## Follow-up

After maintainers are on board, `CONTRIBUTING.md` asks for a docs PR to the [devcontainer.json reference in vscode-docs](https://github.com/microsoft/vscode-docs/blob/main/docs/remote/devcontainerjson-reference.md).

Implements spec#22. Supersedes #311.


SamuelFrost:

src/spec-node/configContainer.ts
```diff
@@ -79,16 +80,48 @@ async function resolveWithLocalFolder(params: DockerResolverParameters, parsedAu
...
 	const documents = createDocuments(cliHost);
-	const content = await documents.readDocument(overrideConfigFile ?? configFile);
+	const content = await documents.readDocument(configUri);
 	if (!content) {
 		return undefined;
 	}
 	const raw = jsonc.parse(content) as DevContainerConfig | undefined;
 	const updated = raw && updateFromOldProperties(raw);
 	if (!updated || typeof updated !== 'object' || Array.isArray(updated)) {
-		throw new ContainerError({ description: `Dev container config (${uriToFsPath(configFile, cliHost.platform)}) must contain a JSON object literal.` });
```
SamuelFrost: Context for this logic:

src/spec-node/imageMetadata.ts
```diff
@@ -199,6 +199,81 @@ export function mergeConfiguration(config: DevContainerConfig, imageMetadata: Im
...
+	const remoteEnv = Object.assign({}, base.remoteEnv, overlay.remoteEnv);
+	assignOrDelete(merged, 'remoteEnv', Object.keys(remoteEnv).length ? remoteEnv : undefined);
+	const containerEnv = Object.assign({}, base.containerEnv, overlay.containerEnv);
+	assignOrDelete(merged, 'containerEnv', Object.keys(containerEnv).length ? containerEnv : undefined);
+	const portsAttributes = Object.assign({}, base.portsAttributes, overlay.portsAttributes);
+	assignOrDelete(merged, 'portsAttributes', Object.keys(portsAttributes).length ? portsAttributes : undefined);
+	const features = Object.assign({}, base.features, overlay.features);
+	assignOrDelete(merged, 'features', Object.keys(features).length ? features : undefined);
+	const customizations = Object.assign({}, base.customizations, overlay.customizations);
+	assignOrDelete(merged, 'customizations', Object.keys(customizations).length ? customizations : undefined);
```
SamuelFrost: Context for this logic

src/spec-node/configContainer.ts
SamuelFrost: This refactor splits **extends resolution** from the rest of config loading.

`readDevContainerConfigObject` loads one JSON/JSONC file and, if it declares `extends`, recursively loads the parent (relative path only), then merges parent + current file before returning. That has to happen **before** workspace setup, variable substitution, and `configFilePath` assignment, which stay in `readDevContainerConfigFile`.

The `seen` set tracks visited config URIs so we can reject cycles (covered in tests). We pass `new Set(seen)` into each recursive call so the same chain is tracked along a single inheritance path.

Review feedback on #311 also asked to resolve nested `extends` (not just one hop) and to drop `extends` from the merged result; this helper is where that lives.

src/spec-node/imageMetadata.ts
SamuelFrost: A plain object spread (`{ ...base, ...overlay }`) would not match the [image metadata merge logic](https://containers.dev/implementors/spec/#merge-logic). That was the main review outcome on #311: `extends` should behave like combining a base image’s metadata with a project `devcontainer.json`, not like a generic deep merge.

So this function reuses the same helpers as `mergeConfiguration`:

- **Union:** `forwardPorts`, `capAdd`, `securityOpt`, `runArgs`, `runServices` (via `unionOrUndefined` / `mergeForwardPorts`)
- **Max / combined:** `hostRequirements` (`mergeHostRequirements`)
- **Per-key merge, overlay wins:** `remoteEnv`, `containerEnv`, `features`, `customizations`, `portsAttributes` (`Object.assign` with overlay last)
- **Any true:** `init` and `privileged` (same idea as `imageMetadata.some(entry => entry.init)` in `mergeConfiguration`)
- **Mounts:** `mergeMounts` on `[base, overlay]` treated as metadata entries

`assignOrDelete` clears properties when the merged value is empty/undefined so we do not leave stale keys from the initial spread. Scalars and scenario-specific fields (`image`, lifecycle commands, compose paths, etc.) still come from the spread with **overlay last**, which matches “current file wins” in the spec.

Net: one merge story for image metadata and for `extends`, instead of two different behaviors.

.github/workflows/dev-containers.yml
```diff
@@ -61,11 +61,12 @@ jobs:
...
+          "--exclude src/test/container-features/containerFeaturesOrder.test.ts --exclude src/test/container-features/registryCompatibilityOCI.test.ts --exclude src/test/container-features/containerFeaturesOCIPush.test.ts --exclude src/test/container-features/e2e.test.ts --exclude src/test/container-features/featuresCLICommands.test.ts --exclude src/test/cli.build.test.ts --exclude src/test/cli.exec.buildKit.1.test.ts --exclude src/test/cli.exec.buildKit.2.test.ts --exclude src/test/cli.exec.nonBuildKit.1.test.ts --exclude src/test/cli.exec.nonBuildKit.2.test.ts --exclude src/test/cli.extends.up.test.ts --exclude src/test/cli.podman.test.ts --exclude src/test/cli.test.ts --exclude src/test/cli.up.test.ts --exclude src/test/httpOCIRegistry.test.ts --exclude src/test/imageMetadata.test.ts 'src/test/**/*.test.ts'",
```
SamuelFrost: Context for this change:

src/spec-node/configContainer.ts
```diff
@@ -79,16 +80,53 @@ async function resolveWithLocalFolder(params: DockerResolverParameters, parsedAu
...
+	}
+
+	const parentUri = resolveConfigFilePath(cliHost, configUri, extendsPath);
+	const parent = await readDevContainerConfigObject(cliHost, parentUri, new Set(seen));
+	if (!parent) {
+		throw new ContainerError({ description: `Dev container config extended from (${uriToFsPath(configUri, cliHost.platform)}) was not found: ${uriToFsPath(parentUri, cliHost.platform)}.` });
+	}
+	return mergeDevContainerConfigs(parent, updated, extendsMergeMode);
+}
+
```
SamuelFrost: Context for this change:

src/spec-node/imageMetadata.ts
```diff
@@ -199,6 +199,114 @@ export function mergeConfiguration(config: DevContainerConfig, imageMetadata: Im
...
+}
+
+function assignOrDelete<K extends keyof DevContainerConfig>(target: DevContainerConfig, key: K, value: DevContainerConfig[K] | undefined) {
+	if (value !== undefined) {
+		target[key] = value;
+	} else {
+		delete target[key];
+	}
+}
+
```
SamuelFrost: Context for this change:

src/spec-node/imageMetadata.ts
```diff
@@ -199,6 +199,114 @@ export function mergeConfiguration(config: DevContainerConfig, imageMetadata: Im
...
+	]);
+	if ('runServices' in merged || runServices) {
+		(merged as DevContainerFromDockerComposeConfig).runServices = runServices;
+		if (!runServices) {
+			delete (merged as DevContainerFromDockerComposeConfig).runServices;
+		}
+	}
+
+	return merged;
+}
```
SamuelFrost: Context for this change:

src/spec-node/imageMetadata.ts
```diff
@@ -199,6 +199,114 @@ export function mergeConfiguration(config: DevContainerConfig, imageMetadata: Im
...
+	assignOrDelete(merged, 'portsAttributes', Object.keys(portsAttributes).length ? portsAttributes : undefined);
+	const features = Object.assign({}, base.features, overlay.features);
+	assignOrDelete(merged, 'features', Object.keys(features).length ? features : undefined);
+	const customizations = Object.assign({}, base.customizations, overlay.customizations);
+	assignOrDelete(merged, 'customizations', Object.keys(customizations).length ? customizations : undefined);
+	const hostRequirements = Object.assign({}, base.hostRequirements, overlay.hostRequirements);
+	assignOrDelete(merged, 'hostRequirements', Object.keys(hostRequirements).length ? hostRequirements : undefined);
+
+	return merged;
+}
```
SamuelFrost: Context for this change:

.github/workflows/dev-containers.yml
SamuelFrost: **What this does (simple terms)**

The CLI already had integration tests that run a real `devcontainer up` (pull/build, start a container, assert JSON + env, tear down). Those live in files like `cli.up.test.ts` and run in their **own CI matrix job** so they get Docker and enough time, and they are **excluded** from the big catch-all `src/test/**/*.test.ts` job to avoid running them twice.

This change adds `cli.extends.up.test.ts` to that same pattern: it installs the **packaged** CLI from this PR (`yarn package` → `devcontainers-cli-*.tgz`) and verifies that `extends` (and `extendsMergeMode`) work end-to-end—not only in unit tests.

**Spec / product context**

[spec#22](https://github.com/devcontainers/spec/issues/22) defines `extends` for inheriting another `devcontainer.json` in-repo. The companion spec PR ([#772](https://github.com/devcontainers/spec/pull/772)) documents merge behavior. These tests guard that the CLI behavior matches what users will configure after the spec lands.

**What the test covers**

- **`extends-up/`** — default `extendsMergeMode` (`combine`): merged `remoteEnv` from base + overlay, container actually starts.
- **`extends-up-override/`** — `extendsMergeMode: "override"`: e.g. `forwardPorts` from the child only, not unioned with the base.

Fixtures use `ubuntu:latest` only (no Features) to keep CI fast, same spirit as `configs/image/`.

src/spec-node/configContainer.ts
SamuelFrost: **What this does (simple terms)**

When you open a `devcontainer.json`, the CLI used to read that one file and continue. With `extends`, your file can say “start from this other JSON file in the repo, then apply my changes on top.”

`readDevContainerConfigObject` is the piece that:

1. Reads one config file (JSON/JSONC).
2. If it has `"extends": "./some-other.json"`, loads that parent (which may itself extend another file).
3. Merges parent + current file into **one** config (using `extendsMergeMode` on the **current** file).
4. Stops if it sees a loop (cycle detection via `seen`).
5. Strips `extends` and `extendsMergeMode` from the result so downstream code sees a normal flat config.

Only after that does `readDevContainerConfigFile` do the usual work: workspace folder, `${localWorkspaceFolder}` substitution, `configFilePath` pointing at the **entry** file you opened.

**Spec context**

- [spec#22](https://github.com/devcontainers/spec/issues/22): `extends` is a relative path in the same repository; no URLs or absolute paths.
- Documented in [spec PR #772](https://github.com/devcontainers/spec/pull/772): optional `extendsMergeMode` — `combine` (default) or `override`.
- Merge semantics for `combine` match [image metadata merge logic](https://containers.dev/implementors/spec/#merge-logic), per review on [CLI #311](https://github.com/devcontainers/cli/pull/311) (Christof Marti, Josh Spicer, Chuck Lantz)—so inheriting a file behaves like layering a project config on top of image metadata.

**Why a separate helper**

Recursive `extends`, validation, and merge must happen **before** substitution and workspace setup. Keeping that in `readDevContainerConfigObject` avoids changing the public shape of `readDevContainerConfigFile` while supporting nested chains (see tests under `src/test/configs/extends/`).

src/spec-node/imageMetadata.ts
SamuelFrost: **What this does (simple terms)**

`mergeDevContainerConfigs` answers: “I have a **base** config (from the extended file) and an **overlay** (the file that declared `extends`). What is the single config we should use?”

When `extendsMergeMode` is **`combine`** (default), we do **not** use a generic JSON deep merge. We use the same rules the CLI already uses when merging **image metadata** with a project `devcontainer.json`—implemented via shared helpers (`mergeForwardPorts`, `mergeHostRequirements`, `mergeMounts`, etc.).

**Intuition for `combine`**

| Kind | Behavior |
|------|----------|
| Lists like `forwardPorts`, `capAdd` | **Union** (both sides contribute) |
| `hostRequirements` | **Maximum** per field (e.g. higher memory wins) |
| Maps like `features`, `remoteEnv` | **Merge keys**; overlay wins on conflict |
| `init` / `privileged` | **True if either** side is true |
| Scalars (`name`, `image`, commands, …) | **Overlay wins** (last file in the chain) |

That matches the spec’s “configuration inheritance” section for `combine` in [spec PR #772](https://github.com/devcontainers/spec/pull/772) and [merge logic](https://containers.dev/implementors/spec/#merge-logic).

**Why not only `{ ...base, ...overlay }`?**

A spread would wrong-foot ports (union vs replace), host requirements (max vs overwrite), and booleans (any-true vs child-only). #311 explicitly asked to reuse image-metadata merge so users have **one** mental model for “layering” configs.

**Specification note**

Other tools (VS Code, Codespaces, etc.) are expected to implement the same merge rules for `extends` over time; the CLI is the reference implementation in this PR.

src/spec-node/imageMetadata.ts
SamuelFrost: **What this does (simple terms)**

`mergeDevContainerConfigsOverride` is the second merge mode, selected when the file that declares `extends` sets `"extendsMergeMode": "override"`.

It is for users who think: “my `devcontainer.json` should **replace** pieces of the base, not **combine** with them.” Example: base forwards ports `[80, 5432]` but you only want `[2222]` in your overlay—you use `override` so you are not stuck with the union.

**Behavior for `override`**

- **Arrays and scalars** set on the overlay **replace** the base (via spread: overlay wins).
- **Object maps** (`remoteEnv`, `containerEnv`, `features`, `customizations`, `portsAttributes`) and **`hostRequirements`** are **shallow-merged** per key so a partial overlay (e.g. only `memory`) does not wipe unrelated base keys (e.g. `cpus`, `storage` from the parent).

**Spec context**

Documented in [devcontainers/spec#772](https://github.com/devcontainers/spec/pull/772) as distinct from `combine`. Default remains **`combine`** so existing spec wording and #311’s “same as image metadata” story stay the default path.

**When to use which**

- **`combine`** — shared team defaults + personal/ repo-specific additions (ports accumulate, stricter host requirements win, features from both).
- **`override`** — “I inherit the base image/compose setup but I want my ports/env/requirements to **mean** what I wrote, not merged with the base lists/max rules.”

Neither `extends` nor `extendsMergeMode` appears in the merged config returned to `up` / `read-configuration`.

src/spec-node/imageMetadata.ts
SamuelFrost: **Context for this block (entry point + two modes)**

`mergeDevContainerConfigs` is the single merge API used when resolving `extends`. It takes:

- **base** — fully resolved parent config (already merged if the parent had its own `extends` chain)
- **overlay** — the current file’s content (without `extends` / `extendsMergeMode`)
- **extendsMergeMode** — read from the **overlay** file only (`combine` if omitted)

It dispatches to:

1. **`combine`** — `mergeDevContainerConfigs` body above this comment: image-metadata rules (see reply on the `combine` hunk).
2. **`override`** — `mergeDevContainerConfigsOverride` below: overlay-style replace/shallow-merge (see reply on the override hunk).

**End-to-end flow**

`devcontainer up` / `read-configuration` → `readDevContainerConfigObject` walks the `extends` chain innermost-to-outermost, calling `mergeDevContainerConfigs` at each step with that file’s `extendsMergeMode` → one flat config → then variable substitution and container create.

**Spec**

Properties and both modes are defined in [spec PR #772](https://github.com/devcontainers/spec/pull/772); implements [spec#22](https://github.com/devcontainers/spec/issues/22). This PR does not bump the npm version; release notes are in `CHANGELOG.md` under Unreleased.

src/spec-node/imageMetadata.ts
```diff
@@ -199,6 +199,100 @@ export function mergeConfiguration(config: DevContainerConfig, imageMetadata: Im
...
+ */
+function mergeDevContainerConfigsOverride(base: DevContainerConfig, overlay: DevContainerConfig): DevContainerConfig {
+	const merged = {
+		...base,
+		...overlay,
+	} as DevContainerConfig;
+	delete merged.extends;
+	delete merged.extendsMergeMode;
+	return merged;
+}
```
SamuelFrost: Context for this change:

SamuelFrost: **What this does (simple terms)**

When the file that declares `extends` sets `extendsMergeMode` to **`override`**, we merge with plain object spread: `{ ...base, ...overlay }`, then remove `extends` / `extendsMergeMode`. No extra merge helpers run on this path.

**Intuition for `override`**

| Situation | Behavior |
|-----------|----------|
| Overlay **omits** a top-level property | Keep the value from **base** (the already-resolved parent chain). |
| Overlay **sets** a top-level property | The overlay’s value **fully replaces** the inherited value for that key. |

Examples:

- `forwardPorts: [443]` on the overlay → `[443]` only (not unioned with the base list).
- `remoteEnv: { FOO: "1" }` on the overlay → only that map; base env keys are **not** retained.
- `hostRequirements: { memory: "4gb" }` → only that object; base `cpus` / `storage` are **not** shallow-merged in.

**Why keep it this simple**

`combine` is the mode for additive layering (union ports, merge map keys, max `hostRequirements`, and so on). `override` is for a **thin** layer on top: replace whole properties you care about, inherit everything you leave out—without re-stating team defaults.

Partial edits to object maps belong on a **`combine`** hop, not on `override`. Shallow-merging maps in override would make the two modes hard to distinguish on the same file.

**Layering**

Each `extends` hop uses **that file’s** `extendsMergeMode`. A common pattern is base → shared `team.json` (`combine`) → local `devcontainer.json` (`override` with only a few keys).

**Spec / commit**

Documented in [spec PR #772](https://github.com/devcontainers/spec/pull/772). Behavior finalized in [`c4e9697`](https://github.com/devcontainers/cli/commit/c4e9697) (integration/unit tests updated in the same commit).

src/test/configs/extends/.devcontainer.base.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.cycle-a.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.cycle-b.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.invalid-merge.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.missing.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.nested.json
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.override.json
SamuelFrost: context for changes:

src/test/configContainer.test.ts
SamuelFrost: context for changes:

src/test/imageMetadata.test.ts
SamuelFrost: context for changes:

src/test/configs/extends/.devcontainer.base.json
SamuelFrost: **Purpose**

Shared **parent** config for the `extends` fixture set. It stands in for a team or repo-wide baseline (image, ports, caps, host requirements, env) that child files inherit from.

**What it exercises (simple terms)**

This file is never opened directly by tests. Child configs point at it with `"extends": "./.devcontainer.base.json"`. When the CLI resolves a child, this JSON is the **base** side of the merge for that hop.

**Spec**

Matches [spec PR #772](https://github.com/devcontainers/spec/pull/772): `extends` is a **relative path** to another devcontainer JSON in the same repo. The merged result should look like one normal config (no `extends` on output).

**Why it looks minimal**

Uses `ubuntu:latest` only—no features or heavy images—so `readDevContainerConfigFile` tests stay fast and focus on **inheritance + merge**, not feature install or Docker `up`. Combine semantics for ports, caps, `hostRequirements`, and `remoteEnv` are asserted in `configContainer.test.ts` against the resolved `raw` config.

src/test/configs/extends/.devcontainer.json
SamuelFrost: **Purpose**

Primary **entry** fixture for default **`extendsMergeMode: combine`** (omitted on the file ⇒ combine).

**What it exercises (simple terms)**

1. CLI reads this file as the user’s devcontainer entry.
2. Loads `.devcontainer.base.json` via `extends`.
3. Merges base + this file with **image-metadata-style combine** rules.
4. Returns a flat config with `extends` / `extendsMergeMode` stripped.

The test checks concrete outcomes: port **union** `[80, 443]`, cap **union**, `hostRequirements.cpus` kept from base with **max** memory after normalizing bytes, `remoteEnv` **key merge** (child wins `OVERRIDE_ME`), scalar `name` from child.

**Spec**

Documents the default inheritance mode in [spec PR #772](https://github.com/devcontainers/spec/pull/772)—same merge model as [merge logic](https://containers.dev/implementors/spec/#merge-logic) for layering project config on image metadata.

**Related tests**

Paired with `.devcontainer.base.json`; nested chain builds on this file via `.devcontainer.nested.json`.

src/test/configs/extends/.devcontainer.nested.json
SamuelFrost: **Purpose**

Proves **multi-hop** `extends` chains work: child → parent → grandparent, each step merged before the next.

**What it exercises (simple terms)**

This file extends `./.devcontainer.json`, which itself extends `./.devcontainer.base.json`. The CLI walks the chain recursively (with cycle detection), applying **combine** at each hop unless a file sets `extendsMergeMode`.

Expected result: settings from base and middle file accumulate, then this file adds `name: "Nested"` and port `2222` → `[80, 443, 2222]`.

**Spec**

[spec#22](https://github.com/devcontainers/spec/issues/22) / #772: inheritance is not limited to depth 1; relative paths resolve from each file’s directory.

**Why a separate fixture**

Single-hop tests don’t catch bugs in “merge then extend again” ordering or in stripping `extends` only on the **final** resolved object.

src/test/configs/extends/.devcontainer.override.json
SamuelFrost: **Purpose**

Entry fixture for **`extendsMergeMode: "override"`** on the **current** file (the file that declares both `extends` and the mode).

**What it exercises (simple terms)**

After loading `.devcontainer.base.json`, merge uses `{ ...base, ...overlay }` semantics—not port union or map deep-merge. The test expects:

- `forwardPorts: [443]` only (not `[80, 443]`)
- `remoteEnv` only `{ OVERRIDE_ME: "child" }` (base `FROM_BASE` dropped)
- `hostRequirements: { memory: "4gb" }` only (no inherited `cpus`)
- `init: false` from the overlay

**Spec**

Override mode is optional and documented in [spec PR #772](https://github.com/devcontainers/spec/pull/772); behavior aligned with [`c4e9697`](https://github.com/devcontainers/cli/commit/c4e9697) on the implementation side.

**Contrast**

Use `.devcontainer.json` (same base parent) to compare **combine** vs **override** side by side in review.

src/test/configs/extends/.devcontainer.cycle-a.json
SamuelFrost: **Purpose**

Half of a **cycle** pair (`cycle-a` ↔ `cycle-b`) used to verify the CLI rejects infinite `extends` chains.

**What it exercises (simple terms)**

Opening `cycle-a` tries to load `cycle-b`, which tries to load `cycle-a` again. The resolver tracks seen config paths and fails with a **cyclic "extends" reference** error before merging.

**Spec**

Inheritance must be acyclic; otherwise relative `extends` graphs could loop forever. This is required behavior for any compliant implementation, not CLI-specific.

**Note**

Image on this file is irrelevant to the test; only the `extends` graph matters. `configContainer.test.ts` starts from `cycle-a` and matches the error message.

src/test/configs/extends/.devcontainer.cycle-b.json
SamuelFrost: **Purpose**

Second half of the **cycle** fixture (`cycle-b` extends `cycle-a` while `cycle-a` extends `cycle-b`).

**What it exercises (simple terms)**

Same as `cycle-a`: ensures cycle detection works regardless of which file in the loop participates. The test enters via `cycle-a`; both files are required to form the loop.

**Spec**

Complements #772 / #22—implementations must detect cycles when resolving relative `extends` paths, not only when a file directly extends itself.

src/test/configs/extends/.devcontainer.invalid-merge.json
SamuelFrost: **Purpose**

Negative test fixture: **`extendsMergeMode` must be `combine` or `override` only**.

**What it exercises (simple terms)**

`"extendsMergeMode": "deepmerge"` is invalid. During read/resolve, the CLI should fail fast with a clear error (test matches message mentioning allowed values)—not silently default or use npm-style deep merge.

**Spec**

#772 defines exactly two modes. Rejecting unknown values keeps behavior predictable across tools (VS Code, Codespaces, CLI).

**Historical note**

Early #311 discussion mentioned deep merge; spec + this PR intentionally use **image metadata merge** for `combine` instead.

src/test/configs/extends/.devcontainer.missing.json
SamuelFrost: **Purpose**

Negative test: **`extends` target must exist** as a relative file from the declaring config.

**What it exercises (simple terms)**

Points at `./does-not-exist.json`. Resolution should fail with a **not found** style error, not produce a partial config.

**Spec**

[spec#22](https://github.com/devcontainers/spec/issues/22): `extends` is a relative path within the repository/workspace context—missing targets are an error at read time (same class of failure as a broken path in normal config loading).

src/test/configContainer.test.ts
SamuelFrost: **Purpose / goal**

End-to-end coverage of **`extends` at config read time** without Docker `up`. These tests extend the existing `readDevContainerConfigFile` suite and assert the **resolved `raw` config** after inheritance—what commands like `read-configuration` and `up` consume later.

**What the process does (simple terms)**

Each test calls `readDevContainerConfigFile` on a fixture under `src/test/configs/extends/`. That runs the full path: parse JSONC → resolve `extends` chain → merge per hop → strip inheritance keys → return substituted config wrapper. We assert selected fields on `config.raw` plus absence of `extends` / `extendsMergeMode`.

**Coverage map**

| Test | Fixture | Intent |
|------|---------|--------|
| resolve extends | `.devcontainer.json` | Default **combine** |
| nested | `.devcontainer.nested.json` | Multi-hop chain |
| cyclic | `.devcontainer.cycle-a.json` | Error |
| override mode | `.devcontainer.override.json` | **override** merge |
| invalid mode | `.devcontainer.invalid-merge.json` | Validation |
| missing file | `.devcontainer.missing.json` | Error |

**Spec**

Validates CLI behavior against [spec PR #772](https://github.com/devcontainers/spec/pull/772) and [merge logic](https://containers.dev/implementors/spec/#merge-logic) for combine; override per spec section on `extendsMergeMode`.

**Test strategy (why not e2e `up`)**

Commit `6302f11` removed `cli.extends.up.test.ts` in favor of this layer: faster CI, same merge assertions without pulling images/features. **Unit** checks for merge primitives live in `imageMetadata.test.ts` (`mergeDevContainerConfigs`); **this file** checks wiring + real JSON files + error paths.

**Conventions**

Follows existing file patterns: chai `assert`, `can` / `rejects` test titles, small `expectReadConfigError` helper for CLI error shape (`err.description || err.message`).

src/test/imageMetadata.test.ts
SamuelFrost: **Purpose / goal**

**Unit tests** for `mergeDevContainerConfigs`—the function `readDevContainerConfigObject` calls for each `extends` hop. They document expected **combine** vs **override** output without reading JSON from disk or starting Docker.

**What the process does (simple terms)**

Two inline base/overlay objects are merged:

1. **Default (`combine`)** — same rules as image metadata merge: union ports, max `hostRequirements`, overlay wins on scalars/commands, boolean `init`/`privileged` OR semantics, etc.
2. **`override`** — spread merge; overlay replaces whole top-level properties it sets; omitted overlay keys retain base.

**Spec**

These cases mirror normative behavior in [spec PR #772](https://github.com/devcontainers/spec/pull/772) and [merge logic](https://containers.dev/implementors/spec/#merge-logic). Fixture tests in `configContainer.test.ts` prove the **full read + extends chain**; this file proves the **merge primitive** in isolation (easier to debug when a single rule regresses).

**Why here, top-level `describe`**

Lives in `imageMetadata.test.ts` next to `mergeConfiguration` tests, using Node **`assert`** and **`should …`** titles like the rest of that file. The suite is a **sibling** of `describe('Image Metadata')` so it does **not** run the Docker/npm `before('Install')` hook—fast in CI and when filtered with `-g mergeDevContainerConfigs`.

**Review tip**

Compare combine expectations here with `.devcontainer.json` + `.devcontainer.base.json` integration test—they should tell the same story at different layers.

SamuelFrost:
Requesting review from @chrmarti @joshspicer @samruddhikhandale and @devcontainers/maintainers when you have bandwidth.

This implements [spec#22](https://github.com/devcontainers/spec/issues/22) / supersedes [CLI #311](https://github.com/devcontainers/cli/pull/311): `extends` (relative JSON/JSONC only, recursive, cycle detection) and optional `extendsMergeMode` (`combine` default, `override`).

**Behavior (short):**
- **`combine`**: same merge rules as image metadata (`mergeDevContainerConfigs` in `imageMetadata.ts`), per earlier [CLI #311](https://github.com/devcontainers/cli/pull/311) discussion.
- **`override`**: `{ ...base, ...overlay }` for each hop—set top-level keys replace inherited values; omitted keys keep the resolved parent chain.

**Companion spec PR:** https://github.com/devcontainers/spec/pull/772

**Where to look:**
- Resolution: `readDevContainerConfigObject` in `configContainer.ts` (inline review threads have more context).
- Merge: `mergeDevContainerConfigs` in `imageMetadata.ts` (combine vs override).
- Tests: `configContainer.test.ts` + fixtures under `src/test/configs/extends/`; unit merge tests in `imageMetadata.test.ts` (`mergeDevContainerConfigs`).

Happy to adjust anything that should match maintainer expectations before [spec PR #772](https://github.com/devcontainers/spec/pull/772) merges. Thanks!

